PR #89, opened 2014-05-26
Added the ability to add program/VM arguments to runtime configs

nekosune:
This PR lets modders set in their buildscript client and server program/VM arguments to the provided runtime configs in addition to the defaults provided.
This allows modders to deal with needed arguments such as -Dfml.coreMods.load or the like with ease, rather then having to remember to readd it every time do a gradle idea or eclipse.


AbrarSyed:
59b82347033a316194c810b1cb7b797b8c120754

PR #168, opened 2014-12-05
Tweak mod plugin

robotbrain:
Add the ability for users to create basic tweaker mods without forge or liteloader


killjoy1221:
You should add the TweakClass manifest entry so the built jar can be put in the mods folder. Other entries like TweakName, TweakAuthor, TweakVersion, and TweakOrder are useful.

AbrarSyed:
Now Exists in FG2.

PR #204, opened 2015-03-18
fixed can`t compile when using jdk7 or jdk8

yuxuanchiadm:
Must set bootclasspath when doing cross version compile. Otherwise you will get a warning. then -Werror will let warning be error.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,6 +81,10 @@ dependencies {
 }
 
 compileJava {
+  if(System.env.JDK6_HOME != null) {
+    options.fork = true
+    options.bootClasspath = "$System.env.JDK6_HOME/jre/lib/rt.jar"
+  }
 	options.deprecation = true
 	options.compilerArgs += ["-Werror"]
 	//options.compilerArgs += ["-Werror", "-Xlint:unchecked"]


build.gradle
```diff
@@ -81,6 +81,10 @@ dependencies {
 }
 
 compileJava {
+  if(System.env.JDK6_HOME != null) {
+    options.fork = true
+    options.bootClasspath = "$System.env.JDK6_HOME/jre/lib/rt.jar"
```
cazzar: This implies everyone has set up a JDK6_HOME variable

AbrarSyed:
no.

either you comment out that section if you wanan built it yourself, or you set up your environment properly with this: https://gist.github.com/AbrarSyed/9440192

LexManos:
You could scan for JDK installs by finding the default and popping a few directories.

matthewprenger:
I'd just remove [this](https://github.com/MinecraftForge/ForgeGradle/blob/FG_1.2/build.gradle#L85) line. Failing the build because of a warning is mildly annoying.

AbrarSyed:
I like that line, it keeps my code free of warnings. I would do the unchecked thing too, but apparently
it doesnt respect the annotations to override it..

but anyways, my earlier point still stands. Either comment it out, or setup your bootstrap classpaths.

PR #207, opened 2015-03-20
Fix common tasks documentation

Unh0lyTigg:
dev and source jar artifacts archives were not set properly.


AbrarSyed:
I have it already fixed locally.
As it is now, a full FG build is kicked off every commit.. so im tying to minimze the doc-only commits until I can fix the jenkins to work the way I want

PR #229, opened 2015-07-02
Update the MCP website URL

kashike:


jk-5:
Just `http://modcoderpack.com` would suffice

kashike:
@jk-5 I used the URL it redirects to, but I can see the reasoning.

jk-5:
Yeah. It just redirects you to that url so better keep it short :+1:

PR #234, opened 2015-08-05
Fix error in SRG format

RX14:


RX14:
This change doesn't seem to have shown up on the docs: https://forgegradle.readthedocs.org/en/FG_1.2/user-guide/shading/ , can this be fixed?

diesieben07:
It is fixed in the latest version: http://forgegradle.readthedocs.org/en/latest/user-guide/shading/
But I do not know why just the root URL defaults to the "FG_1.2" version.

RX14:
Shouldn't FG_1.2 refer to the branch, and the docs commit is on the branch.

Issue #244, opened 2015-08-24
setupCIWorkspace creates classes with inexistent inner classes


Minecrell:
Using `setupCIWorkspace` on the `tweaker-server` plugin creates classes where inner classes are referenced that actually don't even exist in the created JAR:

Here is an example, the class header from `Block` in the created JAR:

![](https://i.imgur.com/caGJYMr.png)

... and `Block$EnumOffsetType` doesn't exist inside the created JAR:

![](https://i.imgur.com/2WXWP3m.png)

This is probably not a problem for `setupDecompWorkspace` because recompilation removes the inner class, however it is causing the compilation to fail on CI servers for some projects because itis trying to find the class: https://travis-ci.org/SpongePowered/SpongeCommon/jobs/77010585#L446-L449

```
error: cannot access EnumOffsetType
  class file for net.minecraft.block.Block$EnumOffsetType not found
  Consult the following stack trace for details.
  com.sun.tools.javac.code.Symbol$CompletionFailure: class file for net.minecraft.block.Block$EnumOffsetType not found
1 error
:compileJava FAILED
```

AbrarSyed:
After some debugging I have found that the that specific enum class is not in the server jar at all. It could be that the exceptor is adding the reference to the class even though the class itself doesnt exist. 

Unfortunately this isnt really a problem i can fix... The exceptor (McInjector) is something I dont know anything about, nor have any desire to touch. If possible, I ask that you guys fix it from your end.

LexManos:
The enum exists as we should be inerged space here, nothing to do with the server or client seperation anymore

Minecrell:
The problem this is causing is basically not only affecting the Mixin annotation processor, but basically every annotation processor trying to collect information about the classes with the bad inner classes. For example, if an annotation processor is trying to check which methods/fields exist in one of these classes it will print this error because it is trying to load the inner class too:

```java
TypeElement element = processingEnv.getElementUtils().getTypeElement("net.minecraft.block.Block");
System.out.println(element);
for (Element e : element.getEnclosedElements()) {
        System.out.println(e);
}
```
Results in:
```
error: cannot access EnumOffsetType
  class file for net.minecraft.block.Block$EnumOffsetType not found
  Consult the following stack trace for details.
  com.sun.tools.javac.code.Symbol$CompletionFailure: class file for net.minecraft.block.Block$EnumOffsetType not found
1 error
```

PR #251, opened 2015-08-29
Scan the exceptor configuration for unknown classes before applying

Minecrell:
This fixes the problem reported in #244 by adding a new deobfuscation step that will scan the generated classes and remove any invalid inner class references it finds.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/tasks/DeobfuscateJar.java
+++ b/src/main/java/net/minecraftforge/gradle/tasks/DeobfuscateJar.java
@@ -13,8 +13,10 @@ import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.zip.ZipFile;
 
 import net.md_5.specialsource.AccessMap;
 import net.md_5.specialsource.Jar;
@@ -291,6 +293,10 @@ public class DeobfuscateJar extends CachedTask
                     }
                 });
             }
+
+            // Remove unknown classes from configuration
+            removeUnknownClasses(inJar, struct);
+
             File jsonTmp = new File(this.getTemporaryDir(), "transformed.json");
             json = jsonTmp.getCanonicalPath();
             Files.write(JsonFactory.GSON.toJson(struct).getBytes(), jsonTmp);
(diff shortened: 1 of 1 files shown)


cpw:
I don't think this is the right way to fix the problem.

Minecrell:
@cpw I agree, it should be probably fixed in Exceptor/MCInjector instead (or whereever this problem is coming from), but after talking with @AbrarSyed we figured this might be the best temporary solution for now.

cpw:
Well, I think we need to figure that out. Exceptor and mcinjector are both projects we can fix..

Minecrell:
@cpw I have changed the fix to scan the provided exceptor configuration for unknown classes instead, and remove them if there are any. While this is something that could be also fixed in Exceptor itself in my opinion it should be probably up to the user of Exceptor to make sure the configuration is valid.

So basically it will now go through the exceptor configuration, check if the classes and inner classes exist and remove them if not. Currently also the class entries itself are removed if the classes don't exist (so for example, client classes when decompiling only the server), however this could be also changed to simply skip them if needed. Shouldn't make a big difference.

LexManos:
THe exceptor config is valid you're just applying a merged config onto a unmerged jar.
This is a USER ERROR nothing to do with exceptor or FG or anything. This is on you.
However, if you want to filter them out, then do it properly in the https://github.com/ModCoderPack/MCInjector damn mcinjector project.

cpw:
Well, that explains the bug then Lex..

AbrarSyed:
So I mistakenly assumed that the EXC and SRG files could be shared between all the jars, including the server, client and merged jars. Apperantly I can only use the EXC on the merged and client jars, and this problem appears if I use it on the server jar.

The options to fix his problem are as follows, orderred by correctness.
1. ship multiple EXCs in the MCP zip and apply the correct EXC file as part of FG.
2. fix the issue in McInjector, so that it doesnt inject claasses into the jar that dont exist.
3. fix it in ForgeGradle like in this here PR.

Now, due to my time constraints, I opted to do number 3. Minecrell was nice enough to send a PR for it before I actually got a chance to do it myself. This is meant to be a stop-gap measure to make stuff work until a more complete fix can be made.
Number 1 required me to add special casing for the server jar, which would be hard to remove later, and generally be a pain. I would also have to talk to bspkrs and have the correct files added to the MCP zip FG downloads. Its much easier to add this as a single removable function as was done in this PR.
Number 2 would definitely be preferred, but I dont have the time to delve into the MC Injector code at this time. If someone could do that for us, thatd be great, but for the time being this current fix is good enough, and isnt all together ugly or anything either.

bloodmc:
@AbrarSyed any update on this? We would like to update Sponge to FG2. Thanks!

LexManos:
In before scala users have issues because scala is dumb and requires all ACCESSED classes to have a inner class attribute as well, so this won't work.

AbrarSyed:
well its still snapshots, if theydont like they can go back to the older stable.. We arnt focring everyone to magically update anymore :)

PR #267, opened 2015-09-14
Make folders to scan for access transformers configurable

Minecrell:
This extends the user extension to make the source sets that will be scanned for access transformers configurable. Currently only the Forge plugin scans for access transformers, for the Vanilla plugins you need to do it manually. While this could be theoretically also added using a simple toggle in the extension (like "scan for ATs") this way is more customizable because any other custom source sets could be simply added to be scanned for access transformers too.

The naming is possibly a bit weird, I'm open for suggestions to improve that.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/UserBaseExtension.java
+++ b/src/main/java/net/minecraftforge/gradle/user/UserBaseExtension.java
@@ -20,6 +20,7 @@
 package net.minecraftforge.gradle.user;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


AbrarSyed:
Instead of taking SourceSet objects, make it take Objects that later resolve to SourceSets. Delaying is always good when possible. Specifically I think it would be best to handle the case where the sourceSet name is only known.

And if your going through improving this anywyas, it would be nice to allow people to specify maybe certain folders as well. Tat way I can only grab ATs in the META_INF like FML does from the main SourceSet, and then everything from API, or some such. The more controls the better :P. Ofcourse thats just a QOL enahncement.

Minecrell:
Well I would basically need to add a quite huge chain of `instanceOf` checks to make that work. And this would make it a bit confusing, for example: would `minecraft.at 'main'` look for an access transformer in the file `main` or would it look for a source set `main`..?

AbrarSyed:
I was thinking it would still be a seperate atSOurce thing, but it would take Files, FileCollections, Strings, SourceSets.. etc

Minecrell:
Ah I see. Will look into that tomorrow.

Minecrell:
@AbrarSyed Sorry for the delay, I think this should do the job. Basically the syntax now is:

```gradle
minecraft {
    atSource sourceSets.main
    atSource 'main'
    atSource files('myfolder')
    atSource delayedFile('...')
    // etc...
}
```

PR #268, opened 2015-09-21
Add an easy way to reobfuscate multiple jars.

killjoy1221:
Example usage: https://gist.github.com/killjoy1221/c9487befa9cafd65054f

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/TaskSingleReobf.java
+++ b/src/main/java/net/minecraftforge/gradle/user/TaskSingleReobf.java
@@ -19,8 +19,6 @@
  */
 package net.minecraftforge.gradle.user;
 
-import groovy.lang.Closure;
-
 import java.io.BufferedOutputStream;
 import java.io.BufferedWriter;
 import java.io.File;
@@ -36,16 +34,6 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import java.util.zip.ZipOutputStream;
 
-import net.md_5.specialsource.Jar;
-import net.md_5.specialsource.JarMapping;
-import net.md_5.specialsource.JarRemapper;
-import net.md_5.specialsource.provider.ClassLoaderProvider;
-import net.md_5.specialsource.provider.JarProvider;
-import net.md_5.specialsource.provider.JointProvider;
-import net.minecraftforge.gradle.common.Constants;
-import net.minecraftforge.gradle.util.GradleConfigurationException;
-import net.minecraftforge.gradle.util.mcp.ReobfExceptor;
-
 import org.gradle.api.DefaultTask;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.tasks.ParallelizableTask;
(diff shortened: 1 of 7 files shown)


killjoy1221:
I've added srgExtra now.  TODO: Config defaults

AbrarSyed:
please squash

killjoy1221:
All squashed.

src/main/java/net/minecraftforge/gradle/user/ReobfConfig.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        this.classpath = classpath;
+    }
+
+    public FileCollection getClasspath()
+    {
+        return classpath;
+    }
+
+    public void setSrgMcp()
```
matthewprenger: This should be called `setSrgNotch` not `setSrgMcp`.

src/main/java/net/minecraftforge/gradle/user/UserBasePlugin.java
```diff
@@ -204,6 +220,34 @@ public void execute(TaskSourceCopy t)
...
+
+            task.setClasspath(reobf.getClasspath());
+            task.setJar(((Jar) jar).getArchivePath());
+            task.setPrimarySrg(delayedFile(reobf.getSrg()));
+
+            task.dependsOn(TASK_GENERATE_SRGS, jar);
+            task.mustRunAfter("test");
+
+            project.getTasks().getByName("build").dependsOn(task);
+            project.getTasks().getByName("assemble").dependsOn(task);
```
matthewprenger: `build` already depends on `assemble` so having both of these statements is redundant.

AbrarSyed:
I think it would be better to explicitely force the specification of the SRG as opposed to using the plugin default (srg names for forge).
Should also allow for secondary SRG files, and extra SRG lines. It might actually be a good idea to have a common interface for the reobf task and this reobf config object, or something that can just be copy-pasted between them...

killjoy1221:
Oops. I must've accidentally removed the srg extra when I squashed. I'll have to fix that. 

I chose to make mcp the default because only Forge used the srg names, which means I only had to override the setupReobf method in the Forge plugin.

AbrarSyed:
Also, please add javadocs to your new reobf extension thing. When I eventually get arround to writing the docs, I will be frequently linking to the javadocs for more information about specific features. It would also be nice to add some javadocs to the setupReobf method just to remind me what it does.. 

I really need to javadoc more...

src/main/java/net/minecraftforge/gradle/user/TaskSingleReobf.java
```diff
@@ -268,6 +268,61 @@ public File getPrimarySrg()
...
+        list.addAll(extraSrgLines);
+        return list;
+    }
+
+    @Override
+    public void extra(Collection<Object> o)
+    {
+        for (Object obj : o)
+        {
+            if (obj instanceof String)
```
matthewprenger: Might need a `GString` check here too.

AbrarSyed:
I wanna release 2.0.2 tonight, please finish this ASAP.

killjoy1221:
All done. ready for pull.

killjoy1221:
Got everything working for real this time. Just need to squash.

PR #271, opened 2015-09-23
Fix tweaker plugins using wrong run classes in start dependency

Minecrell:
Currently the tweaker plugins will always use the original main class in the generated start dependency, instead of calling Launchwrapper to call the tweakers. This causes all tweakers to be skipped when running from the IDE.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/tweakers/ClientTweaker.java
+++ b/src/main/java/net/minecraftforge/gradle/user/tweakers/ClientTweaker.java
@@ -29,4 +29,11 @@ public class ClientTweaker extends TweakerPlugin
     {
         return true;
     }
+
+    @Override
+    protected String getClientRunClass(TweakerExtension ext)
+    {
+        return "net.minecraft.launchwrapper.Launch";
+    }
+
 }
--- a/src/main/java/net/minecraftforge/gradle/user/tweakers/ServerTweaker.java
+++ b/src/main/java/net/minecraftforge/gradle/user/tweakers/ServerTweaker.java
@@ -121,4 +121,11 @@ public class ServerTweaker extends TweakerPlugin
     {
         return false;
     }
+
+    @Override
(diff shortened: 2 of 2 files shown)


bloodmc:
@AbrarSyed can you check this for us? Thanks!

killjoy1221:
Wouldn't it be better to put it in `TweakerPlugin` instead of its subclasses?

Minecrell:
@killjoy1221 Yes and no, it would be probably clearer, but with the client tweaker plugin, running Launchwrapper on the client is not necessary, and the same thing with the server.

killjoy1221:
Now I'm thinking it might be a good idea to allow this to be customized via the minecraft block so things can have its own main class like the Forge server does.
```groovy
minecraft {
    version = '1.8'
    tweakClass = 'some.Tweaker'
    main = 'some.Main' // default to net.minecraft.launchwrapper.Launch
}
```

Minecrell:
Not sure if this would really have any use, because you normally start Launchwrapper first and can do anything from this point on in your Tweaker, e.g. by providing a custom main class for Launchwrapper to load or handling additional things when injecting the classloader.

killjoy1221:
LiteLoader provides its own main class during dev that does things
necessary during dev (setting system properties, logging into Yggdrasil).
This could be done in a tweaker, but I think @Mumfrey wanted running
LiteLoader in dev to be as easy as using Start as main.

Minecrell:
@killjoy1221 This is a good point, however you should create a separate issue about this so it can be implemented in the future. This is mostly just a bugfix so the existing functionality can work.

Minecrell:
@AbrarSyed Can we get this fixed? I would like to remove the hack from SpongeVanilla as soon as possible, and currently it is basically conflicting with all other projects using the tweaker plugin.

Minecrell:
Fixed in https://github.com/MinecraftForge/ForgeGradle/commit/27664502154316b32df64c174f0009502a341196

PR #274, opened 2015-09-29
Fix incorrect coremod launch argument

simon816:
There should be an equals sign not a space.

Note: I have not tested this change, I only modified it from the GitHub web interface.


Minecrell:
You need to change it above for the client too.

simon816:
Oh yeah. Somehow didn't notice that

AbrarSyed:
Just something you noticed? Or is this an actualy broken for the the corenod projects right now?

simon816:
Noticed this when re-setting my Sponge workspace when it updated to FG 2.
Also this: https://forums.spongepowered.org/t/running-sponge-in-eclipse/9432

PR #276, opened 2015-10-14
Fix sources for dependencies in the deobfCompile and deobfProvided configurations not getting downloaded and remapped

Texasjake95:
While I was playing with the new dependency remapping in FG2, I noticed that the sources were not downloading, getting remapped, or getting attached to the jars in an eclipse project. This fixes it.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/UserBasePlugin.java
+++ b/src/main/java/net/minecraftforge/gradle/user/UserBasePlugin.java
@@ -611,12 +611,19 @@ public abstract class UserBasePlugin<T extends UserBaseExtension> extends BasePl
 
                 int taskId = 0;
 
+                // FOR SOURCES!
+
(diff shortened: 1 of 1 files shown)


AbrarSyed:
wait what?? please explain your changes and why they work.

Texasjake95:
I was printing what ids that were being added to the map prior to the change and it did not add any of the dependencies in either the deobfCompile or deobfProvided configurations. Since the incoming dependencies did not have the dependencies I had specified, I dug around in the Gradle source code to find how the eclipse plugin resolved the sources.

Eventually I found that this method https://github.com/gradle/gradle/blob/master/subprojects/ide/src/main/groovy/org/gradle/plugins/ide/internal/IdeDependenciesExtractor.java#L67 is called to resolve the artifacts as well as their sources. 

When I compared this chunk of code to the one FG uses it appeared that the only difference is that the Gradle code uses resolves the artifacts it also collects the ComponentIdentifier's then uses those to collect. When I found that out I applied the same idea to the FG code and it worked after I tested it.

PR #277, opened 2015-10-15
Allow to remove access modifiers in AccessTransformers  without breaking source patches, add final modifier as special case

Barteks2x:
Currently some access transformers may break source patches (one example is removing final modifier from SpawnerAnimals class).  This fixed it.


AbrarSyed:
squash please

PR #295, opened 2015-11-22
Fix sourceJar for Gradle 2.9

matthewprenger:


AbrarSyed:
Wat... Really? That fixes it? But how?

PR #296, opened 2015-11-22
Prevent the vecmath jar from OSX from breaking stuff

kenzierocks:
So that Forge can do stuff without worrying about stupid platform things.


kenzierocks:
Note: Currently untested.

kenzierocks:
Now tested, works perfectly.

PR #299, opened 2015-11-24
Tell gradle about the inputs/outputs of downloadAssets

matthewprenger:
This fixes the task running every build, even when it doesn't need to.


matthewprenger:
Nevermind, after further investigation this won't work.

Issue #218, opened 2015-04-14
Gradle fails with 403 when attempting to GET from https://libraries.minecraft.net/


Shukaro:
> Could not resolve cofh:CoFHCore:[1.7.10]3.0.2-262-dev.
  Required by:
      com.shukaro.artifice:Artifice:1.7.10-1.1.5R-DEV

   > Could not GET 'https://libraries.minecraft.net/cofh/CoFHCore/%5B1.7.10%5D3.
0.2-262-dev/CoFHCore-%5B1.7.10%5D3.0.2-262-dev.pom'. Received status code 403 fr
om server: Forbidden

```
buildscript {
    repositories {
        mavenCentral()
        maven {
            name = "forge"
            url = "http://files.minecraftforge.net/maven"
        }
        maven {
            name = "sonatype"
            url = "https://oss.sonatype.org/content/repositories/snapshots/"
        }
        ivy {
            name "CoFHCore"
            artifactPattern 'http://addons-origin.cursecdn.com/files/2229/536/[module]-[revision].[ext]'
        }
    }
    dependencies {
        classpath 'net.minecraftforge.gradle:ForgeGradle:1.2-SNAPSHOT'
    }
}
```

```
dependencies {
    compile group: 'cofh', name: 'CoFHCore', version: '[1.7.10]3.0.2-262-dev', ext: 'jar'
}
```

AbrarSyed:
NEVER TOUCH THE BUILDSCRIPT BLOCK  unless you know what you are doing. which you dont apprantly.

Dependency repositories go outside the buildscript block.

Shukaro:
```
buildscript {
    repositories {
        mavenCentral()
        maven {
            name = "forge"
            url = "http://files.minecraftforge.net/maven"
        }
        maven {
            name = "sonatype"
            url = "https://oss.sonatype.org/content/repositories/snapshots/"
        }
    }
    dependencies {
        classpath 'net.minecraftforge.gradle:ForgeGradle:1.2-SNAPSHOT'
    }
}
repositories {
    ivy {
        name "CoFHCore"
        artifactPattern 'http://addons-origin.cursecdn.com/files/2229/536/[module]-[revision].[ext]'
    }
}
```

Still occurs after editing build.gradle and running gradlew --refresh-dependencies

AbrarSyed:
then gradle cant find the artifact in the repositories you have specified. A Great way to verify is to try hitting the url in a browser with the tokens replaced.

PR #301, opened 2015-11-27
Store DIR_MCP_MAPPINGS path as a system property when mappings are set

kashike:
This allows the use of the MCP mappings path without having to call `delayedFile`.

https://github.com/MinecraftForge/ForgeGradle/pull/302


AbrarSyed:
But... Why?

kashike:
For example: https://github.com/SpongePowered/SpongeCommon/blob/master/gradle/minecraft.gradle#L11-L60

AbrarSyed:
See #218
Wouldn't it be a better idea to present all the data in a readOnly map or something?

kashike:
So long as it's accessible everywhere needed, yes.

PR #305, opened 2015-12-04
Add mapping type

killjoy1221:
Should allow for quick determination between using notch and searge names.

As requested by @Mumfrey.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/ReobfTaskFactory.java
+++ b/src/main/java/net/minecraftforge/gradle/user/ReobfTaskFactory.java
@@ -29,6 +29,7 @@ import org.gradle.api.Task;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.tasks.bundling.Jar;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 
 import groovy.lang.Closure;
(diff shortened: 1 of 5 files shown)


Mumfrey:
Having thought about this a bit more, I think I can actually do the refmap a better way. If I make refmaps contain both sets of obfuscation mappings then knowing which one to contribute no longer becomes an issue, the mixin processor can just select the appropriate one at runtime. Sorry to waste your time.

killjoy1221:
I'll leave this up to @AbrarSyed101 to decide on this. I don't need it, so
it doesn't matter to me.

Mumfrey:
You didn't read the edit (because github doesn't send update emails). I do actually need it, just not for the reason I originally thought. I need to feed the right SRGs into the obfuscator, the refmap I've dealt with in the way I mentioned above, by just bundling both obf sets in the file.

killjoy1221:
Alright, got the deprecation in there as well as printing a warning when used. I also changed the setupReobf method to use the wrapper so we can possibly track things done in that.

src/main/java/net/minecraftforge/gradle/user/ReobfTaskFactory.java
```diff
@@ -196,15 +220,26 @@ public void extraFiles(Object... o)
...
+        @Deprecated
+        @Override
         public void useNotchSrg()
         {
-            reobf.setPrimarySrg(plugin.delayedFile(Constants.SRG_MCP_TO_NOTCH));
+            setMappingType(ReobfMappingType.NOTCH);
+            warnDeprecation("useNotchSrg()", "mappingType");
+        }
+
+        private void warnDeprecation(String old, String new_)
```
kashike: Perhaps this method should be moved to a more common area (`BasePlugin`?) so that it can be reused? If so, just be more clear in the passed parameters.

PR #316, opened 2015-12-31
Rewrite Download.java to use buffered IO

chaoskagami:
I'm unaware if this is a problem for any other people, but on my linux system for some reason the way Download.java does byte-by-byte writes causes me to get DL speeds not exceeding 100 bytes/s. This causes me to be unable to build any mod from source code without sitting for ~8hrs; I'm sure it isn't a common problem.

Regardless, using byte-by-byte writes is highly inefficient even without any outstanding issues like I have. I rewrote it to use a buffer instead. Quickly checked to make sure result was as expected, md5 and shasums match with anything downloaded, and for my rare edge case it functions normally as well now.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/tasks/Download.java
+++ b/src/main/java/net/minecraftforge/gradle/tasks/Download.java
@@ -27,6 +27,10 @@ import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
 
+import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ReadableByteChannel;
+
 import net.minecraftforge.gradle.common.Constants;
 import net.minecraftforge.gradle.util.caching.Cached;
 import net.minecraftforge.gradle.util.caching.CachedTask;
@@ -59,21 +63,15 @@ public class Download extends CachedTask
         connect.setRequestProperty("User-Agent", Constants.USER_AGENT);
         connect.setInstanceFollowRedirects(true);
 
-        InputStream inStream = connect.getInputStream();
-        OutputStream outStream = new FileOutputStream(outputFile);
+        ReadableByteChannel  inChannel  = Channels.newChannel(connect.getInputStream());
+        FileChannel          outChannel = new FileOutputStream(outputFile).getChannel();
 
-        int data = inStream.read();
-        while (data != -1)
-        {
-            outStream.write(data);
+        // If length is longer than what is available, it copies what is available according to java docs.
+        // Therefore, I use Long.MAX_VALUE which is a theoretical maximum.
+        outChannel.transferFrom(inChannel, 0, Long.MAX_VALUE);
 
-            // read next
-            data = inStream.read();
-        }
-
-        inStream.close();
-        outStream.flush();
-        outStream.close();
+        outChannel.close();
+        inChannel.close();
 
         getLogger().info("Download complete");
     }


cpw:
You are aware you're just moving the loop, right? Look at the implementation of read..

chaoskagami:
I've found this to be triggerable on some versions of OpenJDK, including the one I built from source code on my system.

I'm aware that this is literally an edge case and in most circumstance makes zero changes to what actually happens, yes.

AbrarSyed:
tried using channels? Its supposed to be faster...
https://thomaswabner.wordpress.com/2007/10/09/fast-stream-copy-using-javanio-channels/

chaoskagami:
I was trying to be as non-invasive as possible with my changes in this PR.

That would require a whole new import added to the top and proper checks since it's nonblocking. :P

AbrarSyed:
this isnt forge, I take massive overhaul PRs all the time. Pick the way that makes it the fastest

chaoskagami:
Alright, then. Just being careful, I try not to do too much when I'm not sure if it's needed.

Should I close this one, rewrite it to Channels, then submit another PR? I'm up for it, if we're going for efficiency here.

AbrarSyed:
doesnt matter. I generally prefer 1 PR for a given task unless there is some technical problem why it cant be (github requires different PRs to different branches)

chaoskagami:
I try to keep commit histories for things I work on clean, and rewriting again would mean this rewrite is moot and would be a useless commit in the history.

I'll just continue from current state if that's how you prefer. This shouldn't take long.

AbrarSyed:
1. please rebase so its a single commit
2. is the indentation weird?

chaoskagami:
1. Sorry. Can do.
2. I might have goofed. I'll fix it. I'm still trying to break my tab habit.

AbrarSyed:
last thing, you should actually test to see if its any faster. This can be done by symlinking this repo to the buildSrc folder of the test project. Dont forget to comment-out the buildscript[} or plugins{} block so that it doesnt try to get FG from anywhere other than the buildSrc folder.

to be clear.. you want something like this.
```
 - project/
   - buildSrc/ (FG lives here)
   - build.gradle
```

chaoskagami:
I don't think that it will be easy to generate a fair comparison when involving all the other factors in a build. For example, a crappy dual core is obviously going to underperform compared to a six-core. Using a prebuilt FG as a comparison is always going to seem slower since it needs to build FG with and without for a fair comparison.

If we're just talking the algorithms involved in the download, I've whipped up this class to test both implementations when downloading the 1.8.9 jar; https://gist.github.com/chaoskagami/55baf4e056e8dad9ab4e

time java ForgeFetch b
real    0m6.260s
user    0m1.096s
sys     0m0.152s

So, approximately seven seconds to download an ~8.1M jar. That's the maximum I have for download bandwidth, anyways. The other implementation doesn't work on my linux system, but with java on a windows machine (started from a cygwin bash shell), both implementations:

time java ForgeFetch a
real   0m33.323s
user   0m0.000s
sys   0m0.000s

time java ForgeFetch b
real  0m7.410s
user   0m0.000s
sys   0m0.000s

So it's roughly ~3-4x as fast per jar. Not a huge savings, but certainly notable.

diesieben07:
This is *not* guaranteed to actually copy the whole data. Use a helper method like `ByteStreams.copy(from, to)` to perform a buffered copy from one channel to another. It depends on the channel implementation how many bytes `transferFrom` actually copies, usually it is just what is available in the input buffer and *not* everything.

chaoskagami:
I'll make changes for safety, but is that actually implementation specific? I can't find a source for that. Do you have a documentation link?

I was under the impression that the number of bytes available with a ReadableByteStream over an http connection is the transfer size of whatever is being fetched, and it will copy the number of bytes the download consists of, blocking until it is done.

diesieben07:
From the `transferFrom` JavaDocs:

>Fewer than the requested number of bytes will be transferred if the source channel has fewer than `count` bytes remaining, or if the source channel is non-blocking and has fewer than `count` bytes immediately available in its input buffer.

chaoskagami:
Right, I read that when I started implementing this. The source should use blocking i/o here, though, so the second condition shouldn't ever occur. The first condition will be true, since I deliberately use the maximum possible size - it should stop at the 'end' of the http connection. Unless I'm misinterpreting the docs here. Am I?

Either way, I'm rewriting to use a buffer loop for sanity's sake.

AbrarSyed:
@diesieben07  `ByteStreams.copy(from, to)` uses a straight byte-loop copy the same as was the original FG implementation.

@chaoskagami you can use gradle ... --profile to get per-task timings and metrics to use for the profiling. However, you mentioned that you got speeds as slow as 100b/s and the point of this PR was to speed up the downlaod to actually get a speed that s closer to the pipe limit. I would consider this PR valid if it solves that issue. Otherwise its a premature optimization, and certainly not the bottleneck of the overall setup, and maybe not worth pursuing at all.

chaoskagami:
I changed it (again) to use a buffer with channels to avoid implementation-specific nonsense. It still achieves roughly the same thouroughput as before with use of transferFrom. It is also still at least 3x faster on my windows box compared to the previous algorithm.

It also does solve the 100b/s oddity I encountered and allow me to actually build mods again.

I agree it's a bit of a overkill optimization, though. It can probably save a minute or two for CI builds where it already works fine.

diesieben07:
@AbrarSyed No, it doesn't. It copies via a 4 kilobyte buffer. You copied byte by byte.
@chaoskagami Why did you do this manually instead of using the existing method?

chaoskagami:
Because guava is not part of the standard library, and importing that seems like overkill for ten lines of code? And it isn't possible to change the sub-optimal buffer size of 4k in that class.

diesieben07:
It's not, but FG already includes it: https://github.com/MinecraftForge/ForgeGradle/blob/master/build.gradle#L67

chaoskagami:
It was more the second reason here. Time for one call with the 4k buffer size resulted in roughly three-second slower averages.

I mean, three seconds isn't a lot so if majority of people here would prefer the code simplified to use guava, I'll just change it again.

AbrarSyed:
ok I dont really crae what happens anymore.. as long as its..
1) faster
2) doesnt add any new jar dependencies, you are free to use guava and the stuff already there.

google says channels are the fatest way, I use them for file copying: https://github.com/MinecraftForge/ForgeGradle/blob/master/src/main/java/net/minecraftforge/gradle/common/Constants.java#L311-L352
I dont see a  problem with continuing to use the channels, and I doubt it will break because of different implementations, as both network and fileIO should be blocking streams/channels.

As for code complexity.. Im not worried about that. Its a single class for 1 thing, if there is any complexity in the class at all, it should be in the primary task of the class.

diesieben07:
Abrar, as far as I understand it the class that you linked there is also technically incorrect.

AbrarSyed:
but it works reliably. I use it in quite a few places in FG.

chaoskagami:
Actually, as far as I undertand, that class is fine since FileChannels have a deterministic size, and again, it blocks.

In fact, what I was using before should have been technically correct because it blocked.

AbrarSyed:
I think using the channels without a buffer is fine. So switch to that, and il pull.

chaoskagami:
Can do. Gimme a sec.

PR #320, opened 2016-01-04
[WIP] Add a pack configuration where all artifacts will be packed into the …

matthewprenger:
…jar and the appropriate manifest entry will be created.

Forge/FML Feature: https://github.com/MinecraftForge/MinecraftForge/commit/dc4dc55dcdcb86ec800f2a9170e468673046685b


AbrarSyed:
this is fine for other deps, but requires more if it is also allowed for mod dependencies or mod API dependencies. @cpw is that planned?

killjoy1221:
How will this handle obfuscation? Project deps?

Lunatrius:
The current implementation has a potential problem. When two mods ship different versions of the same dependency FML won't load up (duplicate mod screen etc).

matthewprenger:
@Lunatrius Yeah thats why I stuck the WIP tag on it for now and put it on the back burner. @cpw mentioned he wanted to add an optional semver flag to Mods/Apis so FML could choose the more up to date dep to load.

PR #335, opened 2016-01-23
Fix #334 invalid regex for methods/fields

### LINKED ISSUE #334: [FG_2.1] Not all javadocs from the MCP csv's get added
I've noticed that a few javadocs end up missing in the final processed forgeSrc jar when they exist in the MCP snapshot data.

Example: The lines in the `methods.csv` pulled from the gradle cache:
```
func_71518_a,getCommandUsage,2,Gets the usage string for the command.
func_71514_a,getCommandAliases,2,Get a list of aliases for this command. <b>Never return null!</b>
```
The corresponding methods in the forgeSrc jar `ICommand.java`:

You can see that `getCommandUsage` has its javadocs attached, while `getCommandAliases` does not.

tterrag1098:
Explanation:

Added this regex to the pattern for "words" in the declaration: (?:<.+>)?

This is a noncapturing group which matches anything inside angle brackets (<>), or nothing (the ? character at the end assures it only captures 0-1 instances).

Also added this regex to pattern (method only) to avoid matching return statments use srg names: (?!return)

This is a negative lookahead to check that there is no return statement just after the whitespace.

See http://regexr.com/3cl58 for evidence of the fix.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
+++ b/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
@@ -60,8 +60,8 @@ public class RemapSources extends AbstractEditJarTask
     private final Map<String, String> params       = Maps.newHashMap();
 
     private static final Pattern      SRG_FINDER   = Pattern.compile("func_[0-9]+_[a-zA-Z_]+|field_[0-9]+_[a-zA-Z_]+|p_[\\w]+_\\d+_\\b");
-    private static final Pattern      METHOD       = Pattern.compile("^((?: {4})+|\\t+)(?:[\\w$.\\[\\]]+ )+(func_[0-9]+_[a-zA-Z_]+)\\(");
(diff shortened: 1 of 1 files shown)


matthewprenger:
:+1: Tested and works fine.

kashike:
@AbrarSyed

matthewprenger:
Bump on this.

kashike:
Bump... again.

kenzierocks:
It looks like this has been "merged" for a while, it should be closed.

matthewprenger:
This version of the regex actually has some issues still. I worked with tterrag a few weeks ago and we came up with https://regex101.com/r/NdqI2Y/11. I've had it in my fork for a while and it seems to works fine.

LexManos:
You need test cases and examples of when the current system is wrong.
And why yours is better.

tterrag1098:
All the test cases are at that link. The current impl fails on method
generics (i.e. `public <T> List <T> foo()`)

LexManos:
Still unexplained, and demonstrated that it isn't the proper fix. Come back and actually explain yourself correctly.

Issue #340, opened 2016-02-13
FG2.1 crash on gen mcp patches from non-clean project


Prototik:
As title says FG2.1's patcher crashed when tries to generate mcp patches for project (which based not on `clean`, in this example it patches forge) even without any modifications of project.

build.gradle:
```
    projects {
        forge {
            rootDir 'third-party/MinecraftForge'
            patchDir 'third-party/MinecraftForge/patches/minecraft'
            patchAfter 'clean'
            genPatchesFrom 'clean'
            genMcpPatches = false
            applyMcpPatches = false
            with common
        }
        testForge {
            rootDir '.'
            patchDir 'patches/TestForge'
            patchAfter 'forge'
            genPatchesFrom 'forge'
            genMcpPatches = true
            applyMcpPatches = true
            with common
        }
    }
```
[Setup log](http://pastebin.com/0yv84JCx)  
[Gen patches log](http://pastebin.com/br4iXJS3)

This crash caused by [this](https://github.com/MinecraftForge/ForgeGradle/blob/14461db7fbe930b368f369e5f5137a35bdb055a8/src/main/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java#L830) line which seems are broken for me. There are should be something else, for example output of `remapTestForgeJar` tasks (not sure).

You can download project setup from [git](https://gitlab.prok.pw/Prototik/FGCrash) or as [archive](https://prok.pw/files/TestForge.zip) (in that case don't forge initialize `third-party/MinecraftForge` subproject)

Best regard, Prototik.

Prototik:
@AbrarSyed @LexManos I fixed this issue, which brunch I should create PR? FG_2.1, FG_2.0, maybe master?

liach:
Of course FG_2.1, since this exists in FG 2.1.

AbrarSyed:
looks like your test forge project has no sources.. judging by the log..

AbrarSyed:
uh... so.. whats up?

AbrarSyed:
36e0b600cc64611339b8c010d25a102598fee7b0

PR #341, opened 2016-02-14
Fix gen and apply patches issues on chained projects

Prototik:
See #340


AbrarSyed:
36e0b600cc64611339b8c010d25a102598fee7b0

PR #342, opened 2016-02-19
Use Gradle's built-in zipTree rather than a custom impl

matthewprenger:


LexManos:
The entire point we switched to this is because the default implementation extracts the zip file which is stupid.

matthewprenger:
@LexManos I tested and that apparently isn't the case anymore with newer versions of gradle. Abrar asked me to open this PR.

LexManos:
I'd like to see the changelog from gradle for this.
Considering looking at the https://github.com/gradle/gradle/blob/460b7944980e72f4c4f13b0c04c71e97e4feafb7/subprojects/core/src/main/groovy/org/gradle/api/internal/file/archive/ZipFileTree.java I don't see anything about it....
ANd forcing the fiules to be extracted is thousands of unnessasary tiny files all over the place killing harddrives.

matthewprenger:
Don't know when it happened, but after my testing with this change I couldn't find any expandedArchives anywhere on my filesystem.

matthewprenger:
@LexManos OK managed to find [THIS](https://issues.gradle.org/browse/GRADLE-2959?focusedCommentId=20797&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20797) on gradle's issue tracker.
> zipTree/tarTree as input no more creates temp directory by default. A temp directory still gets created if the task calls the getFiles method on the zipTree/tarTree instance. However this doesn't happen in most cases.

Minecrell:
I could imagine one related changelog is the one for [Gradle 2.9](https://docs.gradle.org/2.9/release-notes#changes-to-up-to-date-checking):

> For input of type zipTree or tarTree, Gradle will no longer extract the archive, but will simply compare the backing archive file for changes.

matthewprenger:
Just as a note: The main reason this PR exists is because the internal `MinimalFileTree` interface that the custom ZipTree impl implemented got a new method recently causing a problem. I thought I'd try testing with Gradle's built-in ZipTree and realized that the temporary files were not being created, so here we are. One less dependency on internal Gradle classes is a good thing IMO.

LexManos:
We need to be SURE that getFiles is never called in any of the codebase then. Im fine with it the only requirement is that it doesnt do the stupid extraction anymore.

matthewprenger:
Tested with both userdev and forgedev. No extracted files to be found.

PR #346, opened 2016-03-01
Liteloader Plugin

Mumfrey:
This PR makes the liteloader plugin fully functional. In particular it:

* allows LiteLoader projects to be set up, fetches information on available versions from the json manifests on my main repository
* supports retromapping the liteloader dependency with the currently selected mappings
* adds a task to generate the liteloader mod descriptor file `litemod.json` via directives in the gradle build file
* brings over idiomatic behaviour from my previous recommended build scripts, such as automatically prepending `mod-` to jar and affixing `-mc<version>` (inhibited if the user sets a value for the artefact name)
* automatically adds the `ModType` manifest entry if building to `jar` rather than `litemod`
* supplies free cookies and beer to all

Changes made to other parts of FG for this to be supported:

* moves `UserBasePlugin::remapDeps` into `UserBasePlugin` itself so it can be called from the derived class
* addresses an issue with source remapping which causes duplicate entries to brick the generator

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/tasks/AbstractEditJarTask.java
+++ b/src/main/java/net/minecraftforge/gradle/tasks/AbstractEditJarTask.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipException;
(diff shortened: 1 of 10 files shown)


killjoy1221:
Can you make this for 2.1, too?

AbrarSyed:
itl get there when I merge 2.0 into 2.1, which will hopefully be soon

PR #351, opened 2016-03-21
Append source set name for IntelliJ runs classpath module

Minecrell:
With these changes ForgeGradle will automatically append the `main` source set name by default to the module name for the classpath. This is required for the run configurations to properly work in IntelliJ IDEA 2016.1 without additional manual changes.

I've also made the special source set configurable, because some project may use a different source set at runtime than the main one. As an example, SpongeForge has an additional `java6` source set which is used to print a proper error to the user to upgrade Java, therefore at runtime the `java6` source set is required (which then depends on the `main` source set). In that case, with this pull request it can be simply configured with:

``` gradle
minecraft {
    runSourceSet = 'java6'
}
```

**I'm still trying to come up with a way to support IntelliJ IDEA 15 at the same time but I haven't found anything simple yet to detect the version the currently created project is using.** I'm open for suggestions.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/UserBaseExtension.java
+++ b/src/main/java/net/minecraftforge/gradle/user/UserBaseExtension.java
@@ -48,6 +48,7 @@ public class UserBaseExtension extends BaseExtension
     private List<Object>            clientRunArgs    = Lists.newArrayList();
     private List<Object>            serverJvmArgs    = Lists.newArrayList();
     private List<Object>            serverRunArgs    = Lists.newArrayList();
+    private Object                  runSourceSet     = "main";
 
     public UserBaseExtension(UserBasePlugin<? extends UserBaseExtension> plugin)
     {
@@ -373,6 +374,14 @@ public class UserBaseExtension extends BaseExtension
         return resolve(getServerRunArgs());
     }
 
+    public SourceSet getRunSourceSet() {
+        return resolveSourceSet(this.runSourceSet);
+    }
+
+    public void setRunSourceSet(Object runSourceSet) {
+        this.runSourceSet = runSourceSet;
+    }
+
     /**
      * Set the run arguments for the server run config
      *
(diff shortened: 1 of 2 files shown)


Minecrell:
@AbrarSyed Any chance to merge this? I've searched the whole file intensively, but there is no real easy way to detect the IntelliJ version without reading additional files (to check if a `_main` module exists), and even then it depends on the project layout you select (e.g. if there is no `.idea` folder).

There will always be one version you will need to make manual changes with, but in my opinion it should be always the latest release you can use without making any additional changes.

cpw:
Good to know, I wasn't aware of this. Likely I'll talk to Lex and merge this in.

kashike:
Should be noted that this PR targets the `master` branch, while is not the active branch currently, @Minecrell.

LexManos:
As it sits seince we have to keep each version seperate for old versions, I'm liking how each version just gets its own branch and 'master' is dead.

AbrarSyed:
I agree with that, but there should be some indication which is 'current', and so far that has been the one on master

On July 6, 2016 2:57:06 PM CDT, LexManos <notifications@github.com> wrote:
>As it sits seince we have to keep each version seperate for old
>versions, I'm liking how each version just gets its own branch and
>'master' is dead.
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/MinecraftForge/ForgeGradle/pull/351#issuecomment-230887990

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Mumfrey:
Settings -> Branches -> Default Branch

LexManos:
Oh they added that? Done. 2.2

AbrarSyed:
Lol, I remember about asking for power to do that ages ago...

On July 6, 2016 3:25:20 PM CDT, LexManos <notifications@github.com> wrote:
>Oh they added that? Done. 2.2
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/MinecraftForge/ForgeGradle/pull/351#issuecomment-230895314

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Minecrell:
@kashike Changed target branch to FG_2.2 so this could be merged directly now.

thatsIch:
Is there a reason this is still not merged in?

Minecrell:
Bump. Can we get this merged? This is still a problem every time you generate IntelliJ IDEA run configurations.

MarkL4YG:
@thatsIch It appears that it might just happen not to get merged anytime.

AbrarSyed:
I guess its been long enough that no one will be using a version of intellij where this wont work anymore.. 
Somebody test and rebase and all that please, and il merge.

Minecrell:
@AbrarSyed As far I can see it still applies cleanly and I don't see any reason why it would be broken now. Should be fine to merge.

AbrarSyed:
@LexManos I approve, take a quick lookover plz

Issue #347, opened 2016-03-10
LiteLoader plugin unusable with subprojects


killjoy1221:
I have a couple projects with the liteloader plugin applied. When the root project has it, it crashes. I'm suspecting a configuration is somehow null.

Log: http://pastebin.com/QthGEk4t
Repo: https://github.com/killjoy1221/TabbyChat-2/tree/1.8.9

AbrarSyed:
@mumfrey !!!

Mumfrey:
Fix should be [something like this](https://github.com/Mumfrey/ForgeGradle/commit/aac624101a9b20982d4333d787c22da7d3304d6c)

Mumfrey:
I'm on an event from tomorrow for the following week so if that doesn't fix it then I'll look into it when I get back.

killjoy1221:
On this note, what's the reason you decided to add liteloader to every project instead of just the one it was applied on? I think that removing the `allprojects` entirely would resolve it.

Mumfrey:
@killjoy1221 oh I see you're right. I was in a hurry yesterday so didn't actually look at it that closely. I think what happened is that the `allprojects` closure is supposed to add the repo the same way it's done [here](https://github.com/MinecraftForge/ForgeGradle/blob/master/src/main/java/net/minecraftforge/gradle/user/UserVanillaBasePlugin.java#L67-L74), [here](https://github.com/MinecraftForge/ForgeGradle/blob/master/src/main/java/net/minecraftforge/gradle/user/patcherUser/PatcherUserBasePlugin.java#L191-L197) and [here](https://github.com/MinecraftForge/ForgeGradle/blob/master/src/main/java/net/minecraftforge/gradle/common/BasePlugin.java#L160-L167) but the second part of the logic should *not* be in the `allprojects` closure.

This makes the fix look more like [this](https://github.com/Mumfrey/ForgeGradle/commit/82496476556e2319d7a35cdd521110957e3d004a) where the logic for that method is moved out of the `allprojects` closure.

Mumfrey:
Hey @killjoy1221 I'm back now, if you can confirm that the changes in the commit I mentioned above fix the issue then I'll open a PR.

killjoy1221:
:+1:

AbrarSyed:
PR please?

PR #353, opened 2016-03-31
Improve the liteloader plugin

killjoy1221:
This includes Mumfrey's fix for #347 

It also fixes the build number being incremented every time gradle is run. Now it's only when the task gets ran.

Additionally, the litemod.json now runs by default and is added to the jar if there is no litemod.json in the resources main sourceset.


killjoy1221:
I just added some description support. Tell me what you think.

@Mumfrey

killjoy1221:
I think Mumfrey's got this actually.

PR #354, opened 2016-04-01
Fix #334 invalid regex for methods/fields

### LINKED ISSUE #334: [FG_2.1] Not all javadocs from the MCP csv's get added

matthewprenger:
Cherry picked from 9ebc7b0821ee81a2bc2e18e569ba8744669a0ba7


matthewprenger:
Just realized @tterrag1098 left his old PR open (#335). Best to just use that one.

PR #376, opened 2016-07-09
Gradle 2.14 + 3.0 Fixes

kenzierocks:
Fixes compiling on 2.14 as well as importing to Eclipse (and maybe IntelliJ).
Makes all `Closure` implementations have owners, strips `thisObject` since it may lead to serialization loops.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,6 +91,8 @@ dependencies {
     //Stuff used in the GradleStart classes
     compileOnly 'com.mojang:authlib:1.5.16'
     compileOnly "net.minecraft:launchwrapper:1.11"
+
+    testCompile 'junit:junit:4.12'
 }
 
 sourceSets {
--- a/src/main/java/edu/sc/seis/launch4j/Launch4jPlugin.java
+++ b/src/main/java/edu/sc/seis/launch4j/Launch4jPlugin.java
@@ -137,7 +137,7 @@ public class Launch4jPlugin implements Plugin<Project>
         task.setGroup(LAUNCH4J_GROUP);
         // more stuff with the java plugin
         //task.with(configureDistSpec(project));
-        task.into(new Closure<File>(null)
+        task.into(new Closure<File>(Launch4jPlugin.class)
         {
             @Override
             public File call(Object... obj)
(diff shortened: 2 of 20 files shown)


kenzierocks:
I've just realized I'd like to bundle the other 2.14 fixes with this.

kenzierocks:
@LexManos or @AbrarSyed: Can this be merged soon? Gradle 3.0 is officially released.

CarrotEngineer:
These changes are required to be able to use Gradle 3's Kotlin build script support, which lets you ditch Groovy, and ForgeGradle otherwise seems to support it just fine (for example: https://github.com/CarrotCodes/Thump/blob/develop/build.gradle.kts). Thump on ForgeCraft is using it and seems fine so far.

LexManos:
I am not seeing how this fixes anything you need to be more specific with wide sweeping changes and make sure that they are backwards compatiblle, it doesn't seem yours are consideirng they call functions they may/may not exist in the old versions {all those new constructors}.
Why does changing all those params to class need to happen?
@CarrotEngineer Don't particularly care to support your shiney new toy. "just because" We need more details then that.
The suggested setup is to use gradle wrapper so there should be no real downside to not doing this as the wrapper lets you use specific gradle versions.

CarrotEngineer:
Build scripts in Kotlin are a core feature of Gradle 3 - not just a toy. They are also still considered a beta feature. I added a comment because it has larger impact than just upgrading for the sake of it, and I consider dropping Groovy to be a huge improvement. There are several announcement videos and blog posts detailing the thinking behind offering Groovy and Kotlin in Gradle 3 if you want to know more, and I don't think me detailing it on a GitHub comment is helpful.

I can set my wrapper to use 3.0 with ForgeGradle without any changes, and it'll crash once due to LogLevel changes, but then function properly after that - so probably not urgent.

LexManos:
I really hare language changes for the sake of language changes -.-
However, going though and actually testing the closure changes seem necessary because gradle pukes on nulls. Can't seem to find documentation as to WHY it pukes or what the proper owner should be. I don't think it's the class, I think it should be the project, but whatever. As I said can't seem to find any documentation as closures are groovy magic and just 'work'
I REALLY don't like these sweeping changes that do a bunch of different things without documentation or explanation.
The deprecation warning one would of been a 'sure its deprecated, we should switch' but there is also concerns about backwards compatibility, however FG2.2 seems to be only for 2.7+ Which has the necessary classes so compatibility *shouldn't* be an issue.
The compile issues part of this pr is actually a no brainer as well and would be a simple 'Sure, not sure why Abrar used these internal classes in the first place' 

But again, this wasn't a 1 change PR it was a Lets put all the things together! And don't explain the biggest one!

LexManos:
Ugh wonderful, your PR has a compile error, awesome! -.-

PR #393, opened 2016-12-04
Fork the decompile task if there's not enough memory

kenzierocks:
I'm not sure if I'm the only one, but I've noticed a few people, including myself, get tripped up on this extra memory requirement. So I wrote this PR to fork the FF decompile if there's not enough free memory in the current process. This has the advantage of not doing the fork for people who currently give Gradle 3GB anyways.

This assumes the decompile will need 2.5GB free, which seems to be
about the right amount.


kenzierocks:
@AbrarSyed is there anything blocking this?

LexManos:
Mainly, just haven't had time to test this.
And i'm reluctant to pull this in to an existing system as it may break current builds.
Any chance of getting some proof it doesn't bork shit?

kenzierocks:
Sure, I'll get you proof of working status for both with and without the required amount of memory for Windows 10, Ubuntu 16.10, and macOS Sierra with Java 8. If anyone else wants to test something else, let me know and I can help.

kenzierocks:
The macOS tests are up under the names `macOS_memory_(1|3)G.txt` here: https://github.com/kenzierocks/ForgeGradleTestProject

View a diff of the two and see the only difference is that the decompile task is not forked in the 3G build. [Here's a pre-made diff](https://gist.github.com/kenzierocks/ca1272c91c702da5c0302827ec1dfa58/revisions).

kenzierocks:
I've made a test script for `*NIX` if anyone else wants to test. I've also uploaded Ubuntu 16.10 runs. Lex, let me know if you want more info included in these runs besides just what gradle writes on the info streams, like Java version (although that can usually be found in the logs).

kenzierocks:
A test script for `Windows` (`test_script.bat`) has now been created. Windows 10 runs with that script are uploaded to the test project. I've also performed Windows Subsystem for Linux tests just to verify that's not broken either.

Everything seems to be working perfectly, is there anything else you want Lex?

PR #397, opened 2017-01-07
Fix licenses

killjoy1221:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sat Aug 20 15:56:50 PDT 2016
+#Fri Jan 06 23:20:47 EST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14-all.zip
```
kashike: This shouldn't be changed.

src/main/java/net/minecraftforge/gradle/common/BasePlugin.java
```diff
@@ -368,7 +368,7 @@ public boolean isSatisfiedBy(Dependency element)
         }), null);
         if (fgDep == null) {
             // This shouldn't happen, unless people are doing really wonky stuff
-            throw new StopExecutionException("Missing FG dep in buildscript classpath");
```
LexManos: Atleast log it, also what are you doing that needs this?

killjoy1221: This fails during the tests. I'm assuming it's because "ForgeGradle" is not in the buildscript dependencies during the tests.

gradle/wrapper/gradle-wrapper.properties
liach: Why not upgrading to gradle 3.3?

src/main/java/net/minecraftforge/gradle/common/BasePlugin.java
kenzierocks: FG needed to be copied to the forked decompile task's classpath so everything worked. This was to cause early failures if it couldn't be located.

PR #398, opened 2017-01-07
Fix tests for fernflower deps

killjoy1221:
Fix tests by adding all the buildscript dependencies instead of just selecting the fg related ones. I trust the old way worked fine when getting fg from maven, but the artifact was always null during tests.

This requires moving the deps declaration to afterEvaluate so they can safely be resolved.

AFAIK, adding all the artifacts doesn't hurt.


killjoy1221:
#399 has this covered

PR #402, opened 2017-01-11
Add support for metadata in the litemod json

killjoy1221:
This builds on the work done for the description, adding a prefix to the map.

Usage:
```groovy
litemod.json {
    description = 'metadata example'
    metadata {
        foo = 'bar'
    }
}
```


killjoy1221:
I've been thinking. I may change this to be more inline with the existing gradle api. Instead of having just a metadata object on the json, I'll have a `merge` method in the extension which takes a map. This would be merged into the json before it is written.

killjoy1221:
Closing in preference to #409

Issue #405, opened 2017-01-17
Memory-Settings Ignorance on decompilation


handicraftsman:
Logs:
```
Executing task ':decompileMc' (up-to-date check took 0.001 secs) due to:
  No history is available.
Note: ForgeGradle is forking a new process to run decompilation.
[...]

PR #406, opened 2017-01-24
Allow users to disable forking decompile at their own risk

kenzierocks:
This adds a flag to revert to the old decompile behavior of ForgeGradle, which is to never fork a new process. This shouldn't be needed except in extreme cases like #405.


handicraftsman:
It works. Yay. It uses even smaller amount of RAM.

PR #415, opened 2017-04-08
Add mkdocs again and update for ForgeGradle 2.2

vinyldarkscratch:


killjoy1221:
You should recommend the [shadow plugin](http://imperceptiblethoughts.com/shadow/) instead for shading.

vinyldarkscratch:
@killjoy1221 I've never heard of that plugin until just now, but it looks interesting, I'll have to dig into it more!  If you were to write the appropriate changes for that plugin, I'd be more than happy to merge it into my own fork so our changes are together in the same PR.  🙂

killjoy1221:
It's OK to keep the current documentation, but suggest using that plugin instead.

LexManos:
Documents should go in the Documentation repo.

vinyldarkscratch:
Ah, didn't know that there was a separate repo for that!  Will forward my contributions over there when I can.  😉

kashike:
I don't see why ForgeGradle documentation would go in the MInecraftForge documentation repository.. they're separate projects, and the documentation is unrelated.

LexManos:
Its the same org. It all goes in one place.

PR #429, opened 2017-06-29
Let user define accessToken etc themselves

MCmodding4K:


AbrarSyed:
You are mistaken.
https://github.com/MinecraftForge/ForgeGradle/blob/FG_2.2/src/main/resources/net/minecraftforge/gradle/GradleStartCommon.java#L74-L78
Notice that the default arguments are put in the Map first, then the CLI args are parse dand entered into the same Map. All of those properties can be overridden by the user. In fact, this PR has absolutely no functional difference from the existing code because all that is entered into an empty Map.

MCmodding4K:
hm, i tried to, but it didn't worked. Now using a random alt i found on the internet :(

Issue #435, opened 2017-08-07
Gradle 4 compatibility


xaeroverse:
To be clear up front, developing a mod with Gradle 4.0 works fine, it's just compiling FG with it that fails.

Log:  https://gist.github.com/xaeroverse/57464ef6454e9c8dc9710e7ccfcfee12

It seems that Gradle 4.0 removed the @ParallelizableTask annotation in their API.  I don't have time to look into this so I'm not sure what the replacement is (if any; it could be redundant now).  Nevertheless, I don't expect this to be high priority anyway since downgrading Gradle compiles fine and fixing it might mean losing backwards compatibility with older Gradle versions.

matthewprenger:
More info about `@ParallelizableTask` being removed: https://discuss.gradle.org/t/parallelizabletask-is-removed-in-gradle-4-0/22970. From the sounds of it, unless you set a gradle config option, the annotations never actually did anything.

Pokechu22:
This was fixed a while back in 3b167e971bea9c250683df4b62586d46d98197df; everything builds fine with Gradle 4.4.1.  It compiles with Gradle 4.6, but the tests fail due to #469; still, this issue can probably be closed.

PR #436, opened 2017-08-07
Update Gradle to 3.0

octylFractal:
For speed and keeping up with Gradle's progress, I think upgrading to 3.0 is a good idea. No mods should be using Gradle 2 anymore. I'd also like to take a look at upgrading to 4.1 (#435), but that's much more of a jump than 2.14 -> 3.0.

I'm not really sure what baseline FG should actually have, so I just went with 3.0.


octylFractal:
This PR is essentially irrelevant now -- FG 3.0 is already on a much later version.

PR #442, opened 2017-08-25
Add a new replace() method with an optional file pattern parameter

### LINKED ISSUE #441: Give replace() optional third parameter specifying file to modify
So then
```gradle
replace "%%%VERSION%%%", project.version
replaceIn "LongFallBoots.java"
```
can be simplified to
```gradle
replace "%%%VERSION%%%", project.version, "LongFallBoots.java"
```
if desired. This would be a lot cleaner than using `replaceIn`.

InsomniaKitten:
Resolves #441


AbrarSyed:
looks good to me

PR #445, opened 2017-09-02
Update SpecialSource and close SS Jar files after use

mezz:
This is a continuation of the work in this PR to SS https://github.com/md-5/SpecialSource/pull/44.

PR #447, opened 2017-09-10
Close Jar files after use

nallar:
* Updates SpecialSource with fixes for jar files not being closed - from #445
* Updates asm to 6.0_BETA to match the version required for new SpecialSource
* Close jar files in ForgeGradle always, using try with resources
* Close jar related input streams always
* Close URLClassLoaders after using for inheritance providers

Fixes md-5/SpecialSource/issues/45

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/patcherUser/TaskApplyBinPatches.java
+++ b/src/main/java/net/minecraftforge/gradle/user/patcherUser/TaskApplyBinPatches.java
@@ -26,6 +26,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.nio.file.*;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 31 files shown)


src/main/java/net/minecraftforge/gradle/user/patcherUser/TaskApplyBinPatches.java
```diff
@@ -106,7 +106,6 @@ public void doTask() throws IOException
                 {
                     ZipEntry n = new ZipEntry(e.getName());
                     n.setTime(e.getTime());
-                    out.putNextEntry(n);
```
nallar: @mezz I think this might be the problem.

Not sure why this line is removed - I don't remember wanting to remove it, but obviously I did... :(

mezz: Ah... that would do it heh

PR #449, opened 2017-09-13
Close Jar files after use

nallar:
Sorry about #447.

This PR includes the previous one, tests for most of the tasks with changes, and fixs for bugs introduced in the binpatch task and the fernflower artifact saver.

I've manually tested the MDK and TickProfiler with dev and decomp workspaces with these changes.

The remaining changed tasks which don't have tests are download tasks. I'm not sure how best to test them.


nallar:
The latest SpecialSource version fails to remap some things. This causes `java.lang.NoSuchFieldError: x at net.minecraft.client.gui.GuiTextField.<init>(SourceFile:50)` with runClient on a non-decomp workspace.

Haven't yet got an mcve of the issue.

Latest test commit bundles a version of SpecialSource with the JarFile changes but without the java 9 support and without ASM 6. This doesn't have the remapping bug.

nallar:
Confirmed the source of the remapping issue as a bug in SpecialSource, see md-5/SpecialSource/issues/47.

nallar:
@mezz Should be ready to go now.

Can you confirm this still solves your original problem with multi-project builds and JVM crashes?

mezz:
Tested with Binnie's and it built successfully.

Barteks2x:
This could explain why when using gradle daemon I had to kill the daemon before running gradle clean when on windows (otherwise it wasn't able to delete the jar files because they are still used by the daemon that tries to delete them).

LexManos:
Lets try this again, mezz says its good to go..

Issue #451, opened 2017-09-25
setupDecompWorkspace fails


dshadowwolf:
Updated the Forge Version and Snapshot version of a project earlier today after a strange error started showing up (unable to find a symbol - never seen that happen before). As I habitually use 'clean' before 'build' I ran a 'cleanCache' to clear the cache and see if the issue was resolved. Afterwards I get an error when trying the 'setupDecompWorkspace' task to get things back on track. (setupDevWorkspace has the code unable to even build thanks to a lot of "missing symbol" errors)

Issue with "setupDecompWorkspace" is:
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index: 24, Size: 9
        at java.util.ArrayList.rangeCheck(ArrayList.java:653)
        at java.util.ArrayList.get(ArrayList.java:429)
        at org.jetbrains.java.decompiler.struct.consts.ConstantPool.<init>(ConstantPool.java:109)
        at org.jetbrains.java.decompiler.struct.StructClass.<init>(StructClass.java:84)
        at org.jetbrains.java.decompiler.struct.StructClass.<init>(StructClass.java:72)
        at org.jetbrains.java.decompiler.struct.StructContext.addArchive(StructContext.java:159)
        at org.jetbrains.java.decompiler.struct.StructContext.addSpace(StructContext.java:100)
        at org.jetbrains.java.decompiler.struct.StructContext.addSpace(StructContext.java:78)
        at org.jetbrains.java.decompiler.main.Fernflower.addAllClasspath(Fernflower.java:119)
        at org.jetbrains.java.decompiler.main.Fernflower.<init>(Fernflower.java:47)
        at org.jetbrains.java.decompiler.main.decompiler.BaseDecompiler.<init>(BaseDecompiler.java:32)
        at net.minecraftforge.gradle.tasks.fernflower.FernFlowerInvoker.runFernFlower(FernFlowerInvoker.java:64)
        at net.minecraftforge.gradle.tasks.fernflower.FernFlowerInvoker.main(FernFlowerInvoker.java:41)
:decompileMc FAILED

dshadowwolf:
Testing with the Forge MDK for 1.12 and it is going through, with several months old mappings file.  Are the mappings from SEP-2017 broken in some manner ?

darkhax:
@dshadowwolf I have had success using the `snapshot_20170919` mappings. So far I have only been able to reproduce this issue when ATs are involved.

[...]

PR #452, opened 2017-09-25
WIP: Don't shade the entire world

LunNova:
* never shade module-info.class - can only be one module-info, must be at root of jar, doesn't make sense for a shaded dependency to provide it
* don't shade transitive dependencies of shaded dependencies
* error if duplicate jar entries would be added

* [x] Tested decomp workspace, dev workspace and builds of mdk, Reliquary, TickProfiler.
* [ ] Tested build of MinecraftForge.
* [x] Tested with gradlew 4.2 and gradle 2.14.

See https://github.com/MinecraftForge/ForgeGradle/issues/451#issuecomment-332007744

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,10 +125,22 @@ processResources {
 }
 
 jar {
+    // Shading could try to add duplicate classes
+    // fail instead of silently adding multiple entries with the same name
+    duplicatesStrategy=DuplicatesStrategy.FAIL
 
-    configurations.shade.each { dep ->
-        from(project.zipTree(dep)){
-            exclude 'META-INF', 'META-INF/**'
+    configurations.shade.getResolvedConfiguration().getResolvedArtifacts().each { dep ->
+        def id = dep.getModuleVersion().id
+        if (configurations.shade.dependencies.find { orig -> orig.group == id.group && orig.name == id.name }) {
+            // shade direct dependencies
+            from(project.zipTree(dep.file)) {
+                exclude 'META-INF', 'META-INF/**', 'module-info.class'
+            }
+        } else {
+            // add transitive dependencies as normal compile dependencies
+            // if a shaded dependency has a transitive dependency which must also be shaded
+            // it must be explicitly added as a shaded depncendy
+            dependencies.compile(id.toString())
         }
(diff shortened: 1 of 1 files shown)


CDAGaming:
Related:
https://travis-ci.org/CDAGaming/MinecraftForge (From #451 )

build.gradle
```diff
@@ -125,10 +125,22 @@ processResources {
 }
 
 jar {
-
-    configurations.shade.each { dep ->
-        from(project.zipTree(dep)){
-            exclude 'META-INF', 'META-INF/**'
+    // Shading could try to add duplicate classes
+    // fail instead of silently adding multiple entries with the same name
+    duplicatesStrategy=DuplicatesStrategy.FAIL
```
AbrarSyed: not `DuplicatesStrategry.EXCLUDE`?

build.gradle
```diff
@@ -125,10 +125,22 @@ processResources {
...
+
+    configurations.shade.getResolvedConfiguration().getResolvedArtifacts().each { dep ->
+        def id = dep.getModuleVersion().id
+        if (configurations.shade.dependencies.find { orig -> orig.group == id.group && orig.name == id.name }) {
```
AbrarSyed: so... only shade stuff thats in the MinecraftForge group?

nallar: That would miss de.oceanlabs stuff.

build.gradle
nallar: With these changes there aren't any duplicate entries, so duplicates should be a failure as if we did see duplicates it would indicate that something had gone wrong when shading.

killjoy1221:
I personally think we should use the shadow plugin. Just need to wait on the module issue there.

https://github.com/johnrengelman/shadow/issues/289

nallar:
@killjoy1221 It's good to avoid duplicating work.

Will using the shadow plugin will include all dependencies in the shaded jar? If it's possible to only shade the relocated packages that is great.

This lets us keep using transitive dependencies (which gradle will download from maven when using the plugin) instead of a fat jar with no dependencies set.

I might not be remembering correctly how it works.

PR #464, opened 2017-12-30
[WIP, experimental] Class javadocs

Pokechu22:
See ModCoderPack/MCPBot-Issues#589.

Experimental support for defining class javadocs, more to show that it can be done than to actually do it.  Requires a `classes.csv` file to be given to the `RemapSources` task; otherwise nothing happens.  Here is [the csv I've been using](https://gist.github.com/Pokechu22/295368414eb621b451315fb62e823f34).

The technique I'm using is to look for class declarations, and then combine the name there and the file it's declared in.  I also re-use this for packages.

This is just a proof of concept at the moment; the actual logic is messy (for instance, I don't know if including filenames in a CSV file is good, packages might be better).

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
+++ b/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
@@ -125,6 +125,7 @@ public class RemapSources extends AbstractEditJarTask
     @Override
     public String asRead(String name, String text)
     {
+        String className = name.replace(".java", "").replaceAll("[/\\\\]", ".");
         ArrayList<String> newLines = new ArrayList<String>();
         for (String line : Constants.lines(text))
         {
@@ -132,7 +133,7 @@ public class RemapSources extends AbstractEditJarTask
             // if we aren't doing javadocs... screw dat.
             if (addsJavadocs)
             {
-                injectJavadoc(newLines, line, name, methodDocs::get, fieldDocs::get, classDocs::get);
+                injectJavadoc(newLines, line, className, methodDocs::get, fieldDocs::get, classDocs::get);
             }
             newLines.add(replaceInLine(line));
         }
@@ -146,11 +147,11 @@ public class RemapSources extends AbstractEditJarTask
      *
      * @param lines The current file content (to be modified by this method)
      * @param line The line that was just read (will not be in the list)
-     * @param name The file's name (including ending .java)
+     * @param name The fully qualified name of the class
      * @param methodFunc A function that takes a method SRG id and returns its javadoc
      * @param fieldFunc A function that takes a field SRG id and returns its javadoc
-     * @param classFunc A function that takes a filename (with .java) followed by the 
-     * class name and returns its javadoc
+     * @param classFunc A function that takes a qualified class name, and then the inner
+     * class name, and returns its javadoc
      */
     public static void injectJavadoc(List<String> lines, String line, String name,
             Function<String, String> methodFunc, Function<String, String> fieldFunc, Function<String, String> classFunc)
@@ -193,7 +194,7 @@ public class RemapSources extends AbstractEditJarTask
             return;
         }
 
-        if (name.endsWith("package-info.java")) {
+        if (name.endsWith("package-info")) {
             matcher = PACKAGE_JAVADOC_PATTERN.matcher(line);
             if (matcher.find()) {
                 String javadoc = classFunc.apply(name + ":" + matcher.group("name"));
--- a/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
+++ b/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
@@ -52,6 +52,9 @@ public class RemapSources extends AbstractEditJarTask
     @InputFile
     private DelayedFile               paramsCsv;
 
+    @InputFile
+    private Object                    classesCsv;
+
     private boolean                   addsJavadocs = true;
 
     private final Map<String, String> methods      = Maps.newHashMap();
@@ -59,36 +62,56 @@ public class RemapSources extends AbstractEditJarTask
     private final Map<String, String> fields       = Maps.newHashMap();
     private final Map<String, String> fieldDocs    = Maps.newHashMap();
     private final Map<String, String> params       = Maps.newHashMap();
+    private final Map<String, String> classDocs    = Maps.newHashMap();
 
 
     private static final Pattern      SRG_FINDER             = Pattern.compile("func_[0-9]+_[a-zA-Z_]+|field_[0-9]+_[a-zA-Z_]+|p_[\\w]+_\\d+_\\b");
     private static final Pattern      METHOD_JAVADOC_PATTERN = Pattern.compile("^(?<indent>(?: {4})+|\\t+)(?!return)(?:\\w+\\s+)*(?<generic><[\\w\\W]*>\\s+)?(?<return>\\w+[\\w$.]*(?:<[\\w\\W]*>)?[\\[\\]]*)\\s+(?<name>func_[0-9]+_[a-zA-Z_]+)\\(");
     private static final Pattern      FIELD_JAVADOC_PATTERN  = Pattern.compile("^(?<indent>(?: {4})+|\\t+)(?!return)(?:\\w+\\s+)*(?:\\w+[\\w$.]*(?:<[\\w\\W]*>)?[\\[\\]]*)\\s+(?<name>field_[0-9]+_[a-zA-Z_]+) *(?:=|;)");
+    private static final Pattern      CLASS_JAVADOC_PATTERN  = Pattern.compile("^(?<indent>(?: {4})*|\\t*)(?:\\w+\\s+)*(?:class|interface|enum) (?<name>\\w+)");
 
     @Override
     public void doStuffBefore() throws Exception
     {
         // read CSV files
-        CSVReader reader = Constants.getReader(getMethodsCsv());
-        for (String[] s : reader.readAll())
+        try (CSVReader reader = Constants.getReader(getMethodsCsv()))
         {
-            methods.put(s[0], s[1]);
-            if (!s[3].isEmpty() && addsJavadocs)
-                methodDocs.put(s[0], s[3]);
+            for (String[] s : reader.readAll())
+            {
+                methods.put(s[0], s[1]);
+                if (!s[3].isEmpty() && addsJavadocs)
+                    methodDocs.put(s[0], s[3]);
+            }
         }
 
-        reader = Constants.getReader(getFieldsCsv());
-        for (String[] s : reader.readAll())
+        try (CSVReader reader = Constants.getReader(getFieldsCsv()))
         {
-            fields.put(s[0], s[1]);
-            if (!s[3].isEmpty() && addsJavadocs)
-                fieldDocs.put(s[0], s[3]);
+            for (String[] s : reader.readAll())
+            {
+                fields.put(s[0], s[1]);
+                if (!s[3].isEmpty() && addsJavadocs)
+                    fieldDocs.put(s[0], s[3]);
+            }
         }
 
-        reader = Constants.getReader(getParamsCsv());
-        for (String[] s : reader.readAll())
+        try (CSVReader reader = Constants.getReader(getParamsCsv()))
         {
-            params.put(s[0], s[1]);
+            for (String[] s : reader.readAll())
+            {
+                params.put(s[0], s[1]);
+            }
+        }
+
+        if (classesCsv != null)
+        {
+            try (CSVReader reader = Constants.getReader(getClassesCsv()))
+            {
+                for (String[] s : reader.readAll()) {
+                    if (!s[3].isEmpty() && addsJavadocs) {
+                        classDocs.put(s[1] + ":" + s[0], s[3]);
+                    }
+                }
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


Pokechu22:
I've changed the code to use fully qualified class names instead of file names because that solves platform-specific issues (which aren't an issue for FG, but are for MCP).

I've also implemented the same thing for MCP; here's a diff:

```diff
diff --git a/runtime/commands.py b/runtime/commands.py
index 66e07c2..c6ef348 100644
--- a/runtime/commands.py
+++ b/runtime/commands.py
@@ -467,6 +467,7 @@ class Commands(object):
         self.has_map_csv = False
         self.has_name_csv = False
         self.has_doc_csv = False
+        self.has_class_doc_csv = False
         self.has_param_csv = False
         self.has_renumber_csv = False
         header_classes = csv_header(self.csvclasses)
@@ -482,6 +483,8 @@ class Commands(object):
             self.has_name_csv = True
         if header_methods >= set(['searge', 'name', 'desc', 'side']) <= header_fields:
             self.has_doc_csv = True
+        if set(['name', 'declaring', 'side', 'desc']) <= header_classes:
+            self.has_class_doc_csv = True
         if set(['param', 'name', 'side']) <= header_params:
             self.has_param_csv = True
         if set(['client', 'server', 'newid']) <= header_newids:
@@ -1987,66 +1990,86 @@ class Commands(object):
             if (int(row['side']) == side or int(row['side']) == 2) and row['desc']:
                 fields[row['searge']] = row['desc'].replace('*/', '* /')
 
+        classes = {}
+        if (self.has_class_doc_csv):
+            classesreader = csv.DictReader(open(self.csvclasses, 'r'))
+            for row in classesreader:
+                #HINT: Only include classes that have a non-empty description
+                if (int(row['side']) == side or int(row['side']) == 2) and row['desc']:
+                    classes[row['declaring'] + ':' + row['name']] = row['desc'].replace('*/', '* /')
+
         regexps = {
-            'field': re.compile(r'^(?P<indent>(?: {4})+|\t+)(?!return)(?:\w+\s+)*(?:\w+[\w$.]*(?:<[\w\W]*>)?[\[\]]*)\s+(?P<name>field_[0-9]+_[a-zA-Z_]+) *(?:=|;)'),
-            'method': re.compile(r'^(?P<indent>(?: {4})+|\t+)(?!return)(?:\w+\s+)*(?P<generic><[\w\W]*>\s+)?(?P<return>\w+[\w$.]*(?:<[\w\W]*>)?[\[\]]*)\s+(?P<name>func_[0-9]+_[a-zA-Z_]+)\('),
+            'field': re.compile(r'^(?P<indent> {4}|\t)(?:[\w$.[\]]+ )*(?P<name>field_[0-9]+_[a-zA-Z_]+) *(?:=|;)'),
+            'method': re.compile(r'^(?P<indent> {4}|\t)(?:[\w$.[\]]+ )*(?P<name>func_[0-9]+_[a-zA-Z_]+)\('),
+            'class': re.compile(r'^(?P<indent>(?: {4})*|\t*)(?:\w+\s+)*(?:class|interface|enum) (?P<name>\w+)'),
+            'package': re.compile(r'^package (?P<name>.+);')
         }
         wrapper = TextWrapper(width=120)
 
+        def insert_javadoc(buf_out, indent, desc, force_multiline):
+            """Insert the given javadoc into the buffer, applying wrapping"""
+            prev_line = buf_out[-1].strip()
+            if not force_multiline and len(desc) < 70 and desc.find(r'\n') == -1:
+                if prev_line != '' and prev_line != '{':
+                    buf_out.append('\n')
+                buf_out.append(indent + '/** ')
+                buf_out.append(desc)
+                buf_out.append(' */\n')
+            else:
+                wrapper.initial_indent = indent + ' * '
+                wrapper.subsequent_indent = indent + ' * '
+                if prev_line != '' and prev_line != '{':
+                    buf_out.append('\n')
+                buf_out.append(indent + '/**\n')
+                desc_lines = desc.split(r'\n')
+                for desc_line in desc_lines:
+                    wrapper.drop_whitespace = desc_line != ' '
+                    buf_out.append(wrapper.fill(desc_line) + '\n')
+                wrapper.drop_whitespace = True
+                buf_out.append(indent + ' */\n')
+
         # HINT: We pathwalk the sources
         for path, _, filelist in os.walk(pathsrclk[side], followlinks=True):
+            package = path[len(pathsrclk[side]) + 1:].replace('\\', '.').replace('/', '.')
             for cur_file in fnmatch.filter(filelist, '*.java'):
                 src_file = os.path.normpath(os.path.join(path, cur_file))
                 tmp_file = src_file + '.tmp'
+                class_name = package + '.' + cur_file[:-len('.java')]
+                is_package = class_name.endswith('package-info')
                 with open(src_file, 'r') as fh:
                     buf_in = fh.readlines()
 
                 buf_out = []
-                #HINT: Look for method/field declarations in this file
+                #HINT: Look for method/field/class declarations in this file
                 for line in buf_in:
                     fielddecl = regexps['field'].match(line)
                     methoddecl = regexps['method'].match(line)
+                    classdecl = regexps['class'].match(line)
+                    packagedecl = regexps['package'].match(line) if is_package else None
                     if fielddecl:
-                        prev_line = buf_out[-1].strip()
                         indent = fielddecl.group('indent')
                         name = fielddecl.group('name')
                         if name in fields:
                             desc = fields[name]
-                            if len(desc) < 70 and desc.find(r'\n') == -1:
-                                if prev_line != '' and prev_line != '{':
-                                    buf_out.append('\n')
-                                buf_out.append(indent + '/** ')
-                                buf_out.append(desc)
-                                buf_out.append(' */\n')
-                            else:
-                                wrapper.initial_indent = indent + ' * '
-                                wrapper.subsequent_indent = indent + ' * '
-                                if prev_line != '' and prev_line != '{':
-                                    buf_out.append('\n')
-                                buf_out.append(indent + '/**\n')
-                                desc_lines = desc.split(r'\n')
-                                for desc_line in desc_lines:
-                                    wrapper.drop_whitespace = desc_line != ' '
-                                    buf_out.append(wrapper.fill(desc_line) + '\n')
-                                wrapper.drop_whitespace = True
-                                buf_out.append(indent + ' */\n')
+                            insert_javadoc(buf_out, indent, desc, False)
                     elif methoddecl:
-                        prev_line = buf_out[-1].strip()
                         indent = methoddecl.group('indent')
                         name = methoddecl.group('name')
                         if name in methods:
                             desc = methods[name]
-                            wrapper.initial_indent = indent + ' * '
-                            wrapper.subsequent_indent = indent + ' * '
-                            if prev_line != '' and prev_line != '{':
-                                buf_out.append('\n')
-                            buf_out.append(indent + '/**\n')
-                            desc_lines = desc.split(r'\n')
-                            for desc_line in desc_lines:
-                                wrapper.drop_whitespace = desc_line != ' '
-                                buf_out.append(wrapper.fill(desc_line) + '\n')
-                            wrapper.drop_whitespace = True
-                            buf_out.append(indent + ' */\n')
+                            insert_javadoc(buf_out, indent, desc, True)
+                    elif classdecl:
+                        indent = classdecl.group('indent')
+                        name = class_name + ':' + classdecl.group('name')
+                        if name in classes:
+                            desc = classes[name]
+                            insert_javadoc(buf_out, indent, desc, True)
+                    elif packagedecl:
+                        indent = ''
+                        name = class_name + ':' + packagedecl.group('name')
+                        if name in classes:
+                            desc = classes[name]
+                            insert_javadoc(buf_out, indent, desc, True)
                     buf_out.append(line)
 
                 with open(tmp_file, 'w') as fh:
```

This does not include the regex changes from ModCoderPack/MCPBot-Issues#590, which should probably also be applied.

LexManos:
First off, I dont trust your regexs. They look oddly similar to the ones we had a while back that we had to make more strict because it was injecting comments where they did not belong.
Secondly, as you mention in your other thread, there are quite a few places that don't get comments on them. Such as enums. With the regex approach this doesn't quite have a simple solution. I've just chalked it up to the fact that we're dealing with decompiled code, so this is the best we get.

Pokechu22:
> First off, I dont trust your regexs. They look oddly similar to the ones we had a while back that we had to make more strict because it was injecting comments where they did not belong.   Ah I see you don't actaully revert them in the FG version just the MCP one. Still not good to do.

For the field/method regexes, it's because they haven't been updated for MCP yet &mdash; MCP is still using the old regexes (ModCoderPack/MCPBot-Issues#590).  They definitely should be updated; I just didn't want to include that in the diff to avoid confusion.

I'm fairly confident in the regexes I'm adding, though they don't actually handle 100% of the possible cases.

For the class regex:
* Per [JLS 8.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.1), `{ClassModifier} class Identifier [TypeParameters] [Superclass] [Superinterfaces] ClassBody` is the format for class declarations.  `ClassModifier` is [defined](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-ClassModifier) as the combination of the keywords for classes and annotations.  (`{}` means 0 or more).
    * My code does not care about anything after identifier, which is fine.
    * My code does _not_ handle annotations (it doesn't accept `@`).  This would be a problem for code like `@Mod public class Foo` but the decompiler always puts annotations on a different line.
    * My code expects a single space after the `class` keyword.  This matches the decompiler output, but technically any other whitespace would be allowed.
        * We need to use caution to avoid `Foo.class` syntax.  However I am reasonably certain that this will never cause issues, as there will always be punctuation before and after it.
* Per [JLS 8.9](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9) enums are essentially the same as classes, just with `enum` instead of class.
* Per [JLS 9.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.1), `{InterfaceModifier} interface Identifier [TypeParameters] [ExtendsInterfaces] InterfaceBody` is the format for interface declarations.  `InterfaceModifier` is [defined](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-InterfaceModifier) much the same way as `ClassModifier`.
    * As with before, my code doesn't handle annotations in the interface modifier.
* My code does not attempt to handle annotation declarations ([JLS 9.6](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6)).  Minecraft does not declare any (the only one is `mcp.MethodsReturnNonNullByDefault`, and that is not eligable for user javadocs)
* All of these assume that the class declaration is on a single line, which is what the decompiler should output.

For the package regex:
* [JLS 7.4](https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html#jls-7.4) details package declarations, which are relatively simple.
    * There are some evil things that are actually legal -- specifically whitespace (including newlines!) in the middle of package declarations.  However since the package-info template is manually created, we can fairly safely assume that the template isn't designed to be evil.
        * We must also assume that annotations are on a different line.
    * JLS says that `package-info.java` is the strongly recommended file, but technically it can be any file.  I don't think we need to worry about annotated packages being in a different file though.

Migrating away from a regex-based approach would be nice (to some kind of AST?), but I don't think it's feasible (especially since the same thing would need to be done in both MCP and FG).

> Secondly, as you mention in your other thread, there are quite a few places that don't get comments on them. Such as enums. With the regex approach this doesn't quite have a simple solution. I've just chalked it up to the fact that we're dealing with decompiled code, so this is the best we get.

Indeed.  One possibility I recently thought of might be to create another annotation that's injected into the class before it's decompiled, that specifies an SRG id for those members.  But that would require a fair bit of work which I'm not sure is worthwhile for now.

> Also, would be nice to get a comprehensive test case for the entirety of the MC codebase. It would be easy enough to mock up a test comment for every class.

Working on doing this now.

Pokechu22:
Using the following change (which excludes package javadocs and forge's side annotations, and also does some manual trimming to avoid word wrapping -- ugly, I know):

```diff
diff --git a/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java b/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
index 0aa7a26..eca2961 100644
--- a/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
+++ b/src/main/java/net/minecraftforge/gradle/tasks/RemapSources.java
@@ -133,7 +133,7 @@ public class RemapSources extends AbstractEditJarTask
             // if we aren't doing javadocs... screw dat.
             if (addsJavadocs)
             {
-                injectJavadoc(newLines, line, className, methodDocs::get, fieldDocs::get, classDocs::get);
+                injectJavadoc(newLines, line, className, methodDocs::get, fieldDocs::get, (s) -> (s.contains("package-info") || s.contains("minecraftforge") ? null : "Doc for " + (s.length() > 80 ? s.substring(0, 40) +  "..." + s.substring(s.length() - 40) : s)));
             }
             newLines.add(replaceInLine(line));
         }
```

`git diff --stat` gives `2066 files changed, 8148 insertions(+)`.  [Here's the full diff](https://gist.github.com/Pokechu22/b6a2857f5b14a15122abcc3ba2af3b53).

Figuring out how whether this is the correct set of changes is a bit hard.  I decided on the regex `^CL: [^ ]+ (?:[^\s$]+|.+\$\D[^\s$]+)$` to count the number of non-anonymous classes (using joined.srg, where I'm detecting an anonymous class as one that has a number right after a `$`), and get 2716.  2716 &times; 3 (number of lines per comment) is 8148, so it seems like I've got all the classes.

williewillus:
what's the progress on this? :D

LexManos:
It isnt looking to bad, we'll need to see about bspkrs and how the export/bot will work. Thats a whole other project that needs modernization.
However, this also should go into FG3, not sure how much we'll care about FG2.3 after FG3's official release. As its for older MC versions. We'll have to see if there is anyone in the community who's trustworthy enough, and wants to step up as the legacy maintainer of the older versions.

DaemonUmbra:
Closing this PR as it does not target the currently supported ForgeGradle branch (``FG_3.0``).

Note that an ongoing effort to update legacy MinecraftForge branches to utilize FG_3.0 is ongoing and accepting informed contributions. (For more information [join the Discord](https://discord.gg/UvedJ9m) and  read through the #mcpconfig channel)

PR #467, opened 2018-01-20
Rewrite McpCleanup.stripComments to fix comment stripping bugs

aucguy:
The method contained some bugs, like how asterisk in multi line comments would
not be stripped and strings ending with asterisks would cause comments to be
recognized within strings subsequent. The old version of the method was patchable,
but it would be more difficult than necessary to think about the correct behavior
as the logic for each state(if the current character was in code, a string or a comment)
was handled in several different places. Instead, the rewrite puts each state
into its own part of the code to make it easy to reason about.

The McpCleanup.stripComments, when applied to src/test/resources/StripComments yields
```
public class StripComments
{

    boolean flag = false;
    *
    public static final String greeting = "hello world'a/*'";
    public static final String nastyStr = "yadadada*";
    //I should get removed
    public static final String victimStr = "something or

    public void someMethod()
    {
        System.out.println("some test");
    }
}
```
Notice the stray asterisk after `boolean flag = false;` and the missing `other";` in `victimStr = "something or`. These are the bugs I'm talking about.

The stray asterisk bug occurs because when the old version came across an asterisk, it did not correctly handle the case where it is in a comment and is not followed by a forward slash. The `victimStr` bug occurs because the code looked ahead a character to handle certain cases and if the case is not met, it blindly added that character. In this case, the asterisk in `nastyStr` caused a look ahead to see if there was a forward slash, which would have caused the end of a comment. Since the next character was not a forward slash, the handling of the quote was skipped so `inComment` was the opposite of what it was supposed to be.

Since there might have been more bugs lurking in the method, I decided to rewrite it.

I've tested that the post decompile task (named fixMcSources under ForgePlugin) yields the same results under the setupDecompWorkspace task. Note that this bug does not come up in minecraft. Over at my fork which decompiles optifine, I found that some optifine classes exemplify this bug.

### DIFF
--- a/src/test/java/net/minecraftforge/gradle/util/mcp/StripCommentsTest.java
+++ b/src/test/java/net/minecraftforge/gradle/util/mcp/StripCommentsTest.java
@@ -1,0 +1,62 @@
+/*
+ * A Gradle plugin for the creation of Minecraft mods and MinecraftForge plugins.
+ * Copyright (C) 2013 Minecraft Forge
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+package net.minecraftforge.gradle.util.mcp;
+
+import org.junit.Test;
+
+import com.google.common.io.ByteStreams;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import org.junit.Assert;
+
+public class StripCommentsTest
+{
+    private static final String[] INPUTS    = new String[] {"StripComments", "UnendedString"};
+    private static final String OUTPUT_POSTFIX = "Out";
+
+    @Test
+    public void testStripComments() throws IOException
+    {
+        for (String i : INPUTS)
+        {
+            String input = readResource(i);
+
+            input = McpCleanup.stripComments(input);
+
+            String[] expected = readResource(i + OUTPUT_POSTFIX).split("\r\n|\r|\n");
+            String[] actual = input.split("\r\n|\r|\n");
+
+            for (int k = 0; k < expected.length; k++)
+            {
+                System.out.println("EXPECTED >>"+expected[k]);
+                System.out.println("ACTUAL   >>"+actual[k]);
+                Assert.assertEquals(expected[k], actual[k]);
+            }
+        }
+    }
+
+    private String readResource(String name) throws IOException
+    {
+        InputStream stream = this.getClass().getClassLoader().getResourceAsStream(name);
+        return new String(ByteStreams.toByteArray(stream));
+    }
+}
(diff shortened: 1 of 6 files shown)


LexManos:
Do you have a diff for the full MC decompile?
Does this actually effect anything in the case where it is invoked? As we're stripping out the FF comments, not any user code.

Pokechu22:
Performed a diff, no changes on the result for regular MC.  That's the same result they had, apparently:

> I've tested that the post decompile task (named fixMcSources under ForgePlugin) yields the same results under the setupDecompWorkspace task. Note that this bug does not come up in minecraft. Over at my fork which decompiles optifine, I found that some optifine classes exemplify this bug.

src/test/java/net/minecraftforge/gradle/util/mcp/StripCommentsTest.java
```diff
@@ -0,0 +1,62 @@
...
+                System.out.println("ACTUAL   >>"+actual[k]);
+                Assert.assertEquals(expected[k], actual[k]);
+            }
+        }
+    }
+
+    private String readResource(String name) throws IOException
+    {
+        InputStream stream = this.getClass().getClassLoader().getResourceAsStream(name);
+        return new String(ByteStreams.toByteArray(stream));
```
Pokechu22: Not super important, but this does leave the stream unclosed.  For my change (#463) I did this instead:

```java

    private String readResource(String name) throws IOException
    {
        try (InputStream stream = this.getClass().getClassLoader().getResourceAsStream(name))
        {
            return new String(ByteStreams.toByteArray(stream));
        }
    }
```

aucguy:
There are no diffs for decompFixed jar (output of the postDecompileTask) for the ForgePlugin. To make sure I made [this](https://gist.github.com/aucguy/e77141fb56ec7a72a11ba4abdd6e2914) bash script so the results can be replicated easily. Run `./genDiffs forgeGradleDirUpstream forgeGradleDirPR mdkDir decompFixed`. ForgeGradleDirUpstream should be the directory of the clone checked out at `FG_2.3`. ForgeGradleDirPR should be a clone checked out at `comment-fix`. MdkDir should be the extracted mdk, but with the line `mavenLocal()` inserted above `jcenter()` in the `build.gradle` file. If there are no differences the script should print `no differences` at the end. Note that the script messes around with the `tmpCommentFix` directory.

The script runs setupDecompWorkspace for both branches, extracts the decompFixed jar and diffs them. To make sure this script actually complains when there are differences, in the comment-fix branch, append a string to the return statement of `McpCleanup.StripComments` (ie `return text + "mess things up";`) and comment out the body of `StripCommentsTest.testStripComments`. It should then output `there are differences`.

I have not tested the patch with PatcherPlugin.

src/test/java/net/minecraftforge/gradle/util/mcp/StripCommentsTest.java
aucguy: I copied and pasted the test code from some another test (probably FmlCleanupTest). That needs to be fixed too. Should I include this fix?

aucguy:
I've tested MinecraftForge with the updated version of the script, and there are still no diffs. However, during the period that this pull request was made and now, there are some changes to ForgeGradle. I will have to test with those.

[The script](https://gist.github.com/aucguy/e77141fb56ec7a72a11ba4abdd6e2914/6d99fd94bea54eca5847e1def0ae4692fb89dba6) was updated to support the `task` argument, which specifies the command to use (ie `setupDecompWorkspace` or `setup`). The usage is now `./genDiffs forgeGradleDirUpstream forgeGradleDirPR mdkDir decompFixed task`. `mdkDir` should be either the mdk or a clone of `MinecraftForge`. `mavenLocal()` needs to be added above `jcenter()` in `build.gradle` for `MinecraftForge` too. `task` should be `setupDecompWorkspace` for the mdk or `setup` for `MinecraftForge`. The usage for the rest of the arguments was described in my previous comment.

LexManos:
Sure we'll try it, it should make literally no difference as this code is only run over MC code... 
This also needs a python conversion when i get a chance as this needs to be synced with MCP to make sure the finalized code output is identical.

PR #468, opened 2018-01-25
Update to SpecialSource 1.8.2

Pokechu22:
Mainly for md-5/SpecialSource#50.


LexManos:
I'll wait till he pushes a version bump, so we dont run into issues again where hes updated and broken things. Move away from the snapshot pulling.

Pokechu22:
Fair enough.  FYI there already has been a version bump for 1.8.1, but that doesn't include these changes.

LexManos:
Ya, I saw, we were using the snapshot because there were breaking changes we needed, or more specifically he fixed a bug that was breaking peoples things. Like I said, i'll wait till the next stable so we can fully test it.

Pokechu22:
SpecialSource 1.8.2 has been released; I've updated this PR to use that rather than the snapshot.

PR #486, opened 2018-03-13
[FG2.3] Fix "MCP unknown"

jriwanek:


LexManos:
Honestly we should just remove that map. It means nothing.

jriwanek:
Well, remove or fix, having it say "Powered my MCP unknown" on every run is just ugly.

PR #516, opened 2018-09-04
Fix class loading.

Hallowizer:
Since the CoremodTweaker is loaded before the FMLTweaker, no class loader exclusion is set for CoreModManager. This results in loading the CoreModManager in the LaunchClassLoader, then once it is excluded, the changes are never seen. Here, I forced the CoreModManager to load CoreModManager using the parent loader.


Hallowizer:
It turns out that the CoremodTweaker is loaded after the FMLTweaker, I was just confused for a bit.

PR #521, opened 2018-10-12
Skip problematic mac native when generating pom

### LINKED ISSUE #520: FG3 not downloading OSX natives
Appears to be a regression of the issue fixed by PR #424 and described here: https://github.com/MinecraftForge/MinecraftForge/issues/3964

Example error:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/Users/grondag/MinecraftForge/build.gradle' line: 69

* What went wrong:
A problem occurred evaluating root project 'Forge'.
> A problem occurred configuring project ':clean'.
   > Could not resolve all files for configuration ':clean:downloadDeps0'.
      > Could not find java-objc-bridge-natives-osx.jar (ca.weblite:java-objc-bridge:1.0.0).
        Searched in the following locations:
            https://repo.maven.apache.org/maven2/ca/weblite/java-objc-bridge/1.0.0/java-objc-bridge-1.0.0-natives-osx.jar
```

JDLogic:
Fixes #520 ~~and removes duplicate entries from the pom file~~.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
@@ -168,6 +168,7 @@ public class MinecraftRepo extends BaseRepo {
                 VersionJson meta = Utils.loadJson(json, VersionJson.class);
                 for (VersionJson.Library lib : meta.libraries) {
                     //TODO: Filter?
+                    builder.dependencies().add(lib.name, "compile");
                     if (lib.downloads.classifiers != null) {
                         if (lib.downloads.classifiers.containsKey("test")) {
                             builder.dependencies().add(lib.name, "test").withClassifier("test");
@@ -175,8 +176,6 @@ public class MinecraftRepo extends BaseRepo {
                         if (lib.natives != null && lib.natives.containsKey(CURRENT_OS) && !lib.getArtifact().getName().contains("java-objc-bridge")) {
                             builder.dependencies().add(lib.name, "runtime").withClassifier(lib.natives.get(CURRENT_OS));
                         }
-                    } else {
-                        builder.dependencies().add(lib.name, "compile");
                     }
                 }
             } else {
--- a/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
@@ -168,14 +168,15 @@ public class MinecraftRepo extends BaseRepo {
                 VersionJson meta = Utils.loadJson(json, VersionJson.class);
                 for (VersionJson.Library lib : meta.libraries) {
                     //TODO: Filter?
-                    builder.dependencies().add(lib.name, "compile");
                     if (lib.downloads.classifiers != null) {
                         if (lib.downloads.classifiers.containsKey("test")) {
                             builder.dependencies().add(lib.name, "test").withClassifier("test");
                         }
-                        if (lib.natives != null && lib.natives.containsKey(CURRENT_OS)) {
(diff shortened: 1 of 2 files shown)


JDLogic:
Looking into this further, do the natives need to be specified in the pom file at all? The extract natives task already downloads all those libraries from Mojang and extracts the contents into common folder. The only thing that is missing is specifying that folder using the `java.library.path` environment variable.

src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
```diff
@@ -168,14 +168,15 @@ protected File findPom(String side, String version, File json) throws IOExceptio
...
                     if (lib.downloads.classifiers != null) {
                         if (lib.downloads.classifiers.containsKey("test")) {
                             builder.dependencies().add(lib.name, "test").withClassifier("test");
                         }
-                        if (lib.natives != null && lib.natives.containsKey(CURRENT_OS)) {
+                        if (lib.natives != null && lib.natives.containsKey(CURRENT_OS) && !lib.getArtifact().getName().contains("java-objc-bridge")) {
                             builder.dependencies().add(lib.name, "runtime").withClassifier(lib.natives.get(CURRENT_OS));
                         }
+                    } else {
+                        builder.dependencies().add(lib.name, "compile");
```
LexManos: This shouldn't be correct, we want the main artifact AND the classified one...

src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
```diff
@@ -168,14 +168,15 @@ protected File findPom(String side, String version, File json) throws IOExceptio
...
                 for (VersionJson.Library lib : meta.libraries) {
                     //TODO: Filter?
-                    builder.dependencies().add(lib.name, "compile");
                     if (lib.downloads.classifiers != null) {
                         if (lib.downloads.classifiers.containsKey("test")) {
                             builder.dependencies().add(lib.name, "test").withClassifier("test");
                         }
-                        if (lib.natives != null && lib.natives.containsKey(CURRENT_OS)) {
+                        if (lib.natives != null && lib.natives.containsKey(CURRENT_OS) && !lib.getArtifact().getName().contains("java-objc-bridge")) {
                             builder.dependencies().add(lib.name, "runtime").withClassifier(lib.natives.get(CURRENT_OS));
```
LexManos, line 177: Is there a better way then hard codding the specific file name? Seems REALLY prone to breaking.
Also, any idea what the official natives method for gradle is? I dont think us manually extracting things is the best approach.

src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
JDLogic: The version json list the main artifact by itself and then again with classified artifacts. Therefore, the main artifact gets added to the dependency list twice. I realize though that there is a better way to deal with this in case this changes in the future, so I will leave this for a separate PR.

src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
JDLogic: I agree that this is not a great solution. With that said, I do not know what other options we have here. Once this dependency is added to the pom, gradle will try to resolve it and fail. ~~Even with Mojang's repo added it will fail because their repo does not use proper classifiers.~~

Also, I did search around quite a bit. It does not seen like gradle has any builtin support for natives, so I am not sure if we can get around manually extracting them.

Edit: Mojang's repo works just fine. I was just being stupid. Also, Gradle does not check their repo for this artifact because it finds the pom file for the main file on maven central and assumes this artifact is there as well.

LexManos:
Still not a fan of hardcoded hacks, but macs are horrible... oh well..

PR #522, opened 2018-10-27
Use 'MC_VERSION', for 'downloadMCMeta' task, instead of 'MCP_VERSION'

OhYea777:
It is currently trying to download the MC meta using the MCP version, rather than the MC version.


OhYea777:
Closing as it has been fixed in commit LexManos@deb157dcc69e86229f648a0f607f830d7626290d

PR #525, opened 2018-11-07
Set up for JUnit 5 testing

OhYea777:
- Uses [JUnit 5](https://junit.org/junit5/), the latest version of the JUnit test framework
- Enables Gradle's native JUnit 5 support (added natively in [Gradle 4.6](https://docs.gradle.org/4.6/release-notes.html))
- Added some simple unit tests for the config loading classes that are in the common source set


AbrarSyed:
Should go all in on Junit 5 and update the existing tests to use whatever new API there is. And hopefully ditch the vitnage engine while you at it.

LexManos:
Do not include MCPConfig in the test data, download it instead.
Also, not sure what these tests are meant to actually.. test... you're just loading json... We know that GSON works...

PR #526, opened 2018-11-07
Update and actually use the licenser plugin

jriwanek:


LexManos:
This needs a PR to enable Travis first, so we can verify it wont break the builds, and new PRs wont break due to header info.
Also, the replacements *should* work, did licensor change how it replaces the date/name in the file?

jriwanek:
Not that I'm aware of, I just did it the same way it was done in forge itself, I can switch it back over to using replacements if that's preferred.

LexManos:
I'd prefer replacements, because I don't like hardcoding things. Esp the year, considering it's about to change. So having it 2018-{year} would be nice. Tho, that would mean we have to do a pass on updating the headers come new years...

jriwanek:
Okay, will do.

liach:
fixed in b8ec4ff2623d1a1f65a9d077c2cbe054c04071b7 already

PR #528, opened 2018-11-13
Ensure output folder is created when downloadEtag

Librazy:
So no FileNotFoundException from FileOutputStream as it requires the directory exists.

When building forge 1.13-pre in a clean Windows environment (should be same for Linux), ForgeGradle will create `.gradle\caches\forge_gradle\minecraft_repo\versions` and attemp to download things to `1.13` folder which does not exist yet, causing a 
[FileNotFoundException](https://github.com/MinecraftForge/ForgeGradle/files/2577079/FileNotFoundException.log) and breaks the build.


LexManos:
You know, i literally thought to myself, "make sure you make the directories" as i was writing the efile line..... just skipped it when i wrote the write.

PR #529, opened 2018-11-15
Avoid crashes in userdev.

gigaherz:
- Don't crash if local repository file does not exist. 
- Fix userdev injections crashing.


LexManos:
https://github.com/MinecraftForge/ForgeGradle/commit/d5889eba3366caa934a6fafdfc220ce8dd0c16ff

PR #534, opened 2018-12-06
Write "libraries.txt" in UTF-8.

mcenderdragon:
This fixes crashes from "setupMCP" like this one: https://hastebin.com/inewirulop.cs. 
It is caused if the path contains UTF characters which then gets saved in a different format then read by the jetbrains decoder.

### DIFF
--- a/src/mcp/java/net/minecraftforge/gradle/mcp/function/ListLibrariesFunction.java
+++ b/src/mcp/java/net/minecraftforge/gradle/mcp/function/ListLibrariesFunction.java
@@ -27,12 +27,15 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 import java.io.File;
+import java.io.FileOutputStream
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.PrintWriter;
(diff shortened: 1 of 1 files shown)


liach:
Nice job. Is this the first pr to fg that's totally done on github?

mcenderdragon:
> 
> 
> Nice job. Is this the first pr to fg that's totally done on github?

Possible :D, but I also solved a merge conflict only in the webbrowser ^^.

Edit : #528  well I think this one could also be done this way. (but it has a correct branch name, so I doubt it)

mcenderdragon:
Lex regarding you concerns if it always is UTF8:
according to your [patch](https://github.com/MinecraftForge/ForgeFlower/blob/master/FernFlower-Patches/0019-Add-cfg-argument-Used-to-specify-a-text-file-with-ad.patch#L52) `Files.lines(path) ` is used which just calls `Files.lines(path, StandardCharsets.UTF_8)` ([according to the javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#lines-java.nio.file.Path-))

PR #535, opened 2018-12-16
Fix crash in forgedev due to working directory not being assigned.

gigaherz:


tterrag1098:
Can confirm this fixes the problem.

PR #536, opened 2018-12-19
Apply MCInjector to userdev jar

### LINKED ISSUE #530: FG_3.0 Extra Parameter Annotation being added to TextFormatting.class
Ive been messing with 1.13 forge, and found an issue when trying to build a mod that uses TextFormatting.class 

Java compiler output when running `gradlew.bat build`:
https://gist.github.com/modmuss50/54677952de67d2e752456506875d28c8 

This seems to be caused by an extra @Nullable paramter being added when it shouldn't be. (seen using idea's bytecode viewer)

idea64_2018-11-19_19-36-24

This can be reproduced by cloning and importing the lastest 1.13 test MDK into idea and adding the following line

`LOGGER.info(TextFormatting.RED);` into the example mod, any use of the text formatting class seems to cause this issue.

Let me know if you need anymore info.

JDLogic:
Fixes #530

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
@@ -22,7 +22,6 @@ package net.minecraftforge.gradle.userdev;
 
 
 import java.io.*;
-import java.nio.file.FileSystem;
 import java.nio.file.*;
 import java.util.*;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


src/userdev/java/net/minecraftforge/gradle/userdev/tasks/ApplyMCInjector.java
```diff
@@ -0,0 +1,109 @@
...
+    private File output;
+    private File mcp;
+
+    private File exceptions;
+    private File access;
+    private File constructors;
+
+    public ApplyMCInjector() {
+        tool = "de.oceanlabs.mcp:mcinjector:3.7.3:fatjar";
+        args = new String[] { "--in", "{input}", "--out", "{output}", "--log", "{log}", "--lvt=LVT", "--exc", "{exceptions}", "--acc", "{access}", "--ctr", "{constructors}" };
```
LexManos: This information should be pulled from the config.

JDLogic:
Updated the new task to be more generic. All data is now read from the MCP config file.

PR #537, opened 2018-12-20
Add 'api' source set to userdev

OhYea777:


LexManos:
Why? This settings to add hard lessons,  also you can add this yourself if you're doing api in your mod.

OhYea777:
I suppose that's fair to delegate it for the user to add if deemed necessary

PR #543, opened 2019-01-09
[1.13] Fixes 'methodMissing' issue. Allows merging parent RunConfigs into multiple RunConfigs

OhYea777:
This is the refactored section I am using in my _**build.gradle**_:

```groovy
runs {
    client {
        ideaModule "${group}.${name}.main"
        workingDirectory project.file('run').canonicalPath

        source sourceSets.api
        source sourceSets.main
    }

    testClient {
        ideaModule "${group}.${name}.test"
        workingDirectory project.file('run').canonicalPath

        source sourceSets.api
        source sourceSets.main
        source sourceSets.test

        merge 'client'
    }

    server {
        ideaModule "${group}.${name}.main"
        workingDirectory project.file('run').canonicalPath

        source sourceSets.api
        source sourceSets.main
    }

    testServer {
        ideaModule "${group}.${name}.test"
        workingDirectory project.file('run').canonicalPath

        source sourceSets.api
        source sourceSets.main
        source sourceSets.test

        merge 'server'
    }
}
```

It removes the need to use `client = {` and also allows the ability of creating custom RunConfigs that merge from its parent's

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
@@ -20,33 +20,79 @@
 
 package net.minecraftforge.gradle.common.util;
 
+import com.google.common.collect.ImmutableList;
+import groovy.util.MapEntry;
+import org.gradle.api.Project;
+import org.gradle.api.Task;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.JavaExec;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.TaskProvider;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.io.File;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-
-import org.gradle.api.tasks.SourceSet;
-
-import com.google.common.base.Joiner;
-
-import groovy.lang.Closure;
-import groovy.lang.MissingPropertyException;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class RunConfig implements Serializable {
+
     private static final String MCP_CLIENT_MAIN = "mcp.client.Start";
     private static final String MC_CLIENT_MAIN = "net.minecraft.client.main.Main";
+
     private static final long serialVersionUID = 1L;
-    private String main;
-    private List<String> args;
-    private List<String> jvmArgs;
-    private Map<String, String> env;
-    private Map<String, String> props;
+
+    private transient final Project project;
+    private final String name;
+
+    private String taskName, main, ideaModule, workDir;
+
     private boolean singleInstance = false;
-    private String ideaModule;
-    private String workDir;
+
+    private List<String> args, jvmArgs;
     private List<SourceSet> sources;
+    private List<RunConfig> parents, children;
+    private Map<String, String> env, props, tokens;
+
+    public RunConfig(@Nonnull final Project project, @Nonnull final String name) {
+        this.project = project;
+        this.name = name;
+    }
+
+    public final String getName() {
+        return name;
+    }
+    public void setTaskName(String taskName) {
+        this.taskName = taskName;
+    }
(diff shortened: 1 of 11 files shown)


src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
```diff
@@ -196,12 +211,12 @@ private String replace(Map<String, String> vars, String value) {
         return resolved == null ? value : resolved;
     }
 
-    public static class Container {
+    public static class Container extends GroovyObjectSupport {
         private Map<String, RunConfig> runs = new HashMap<>();
```
LexManos: I guess this is the random magic that makes it actually call methodMissing? You should change the comment below to reflect that.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherExtension.java
```diff
@@ -144,7 +144,7 @@ public void setExc(List<File> files) {
         return excs;
     }
 
-    public void runs(Closure<? super RunConfig.Container> value) {
+    public void runs(Closure value) {
         value.setResolveStrategy(Closure.DELEGATE_FIRST);
```
LexManos: Is there a real need to change this? As it'd be better to not have the blank generic.

src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
```diff
@@ -165,7 +165,13 @@ public void validate(Configuration cfg, Map<String, RunConfig> runs, File native
...
 
             parent.getConfig().runs.forEach((name, dev) -> {
-                runs.computeIfAbsent(name, k -> new RunConfig()).merge(dev, false, vars);
+                List<RunConfig> runsToMerge = runs.values().stream().filter(run -> name.equals(run.getMerge())).collect(Collectors.toList());
+
+                if (runs.containsKey(name)) {
+                    runsToMerge.add(runs.get(name));
+                }
+
+                runsToMerge.forEach(run -> run.merge(dev, false, vars));
```
LexManos: This doesn't create if the build.gradle doesn't have it. Also it doesn't merge things that are not provided by the MDK.
What you probably want is to introduce a 'parent/child' system into to the run configs. And then walk the parents:

```
  client {
  }
  testClient {
    parent runs.client
  }
```

``` 
  run = runs.computeIfAbsent(name, new run);
   run.merge(dev);
   runs.forEach(e -> if (e.getParent() == null) propogate(run));

void propogate(run) {
  run.getChildren().forEach(child -> {
    child.merge(run);
    propogate(child)
  }
}
````

File: src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java [hunk 4]
```diff
@@ -20,40 +20,34 @@
...
-import org.gradle.api.NamedDomainObjectContainer;
-import org.gradle.api.NamedDomainObjectFactory;
-import org.gradle.api.Plugin;
-import org.gradle.api.Project;
-import org.gradle.api.Task;
+import net.minecraftforge.gradle.mcp.MCPRepo;
+import net.minecraftforge.gradle.userdev.tasks.GenerateSRG;
+import net.minecraftforge.gradle.userdev.tasks.RenameJarInPlace;
+import org.codehaus.groovy.runtime.StringGroovyMethods;
+import org.gradle.api.*;
```
LexManos: We do not use wildcard imports on our project. And refrain for blanket import reorg.

OhYea777:
This pull request depends on MinecraftForge/MinecraftForge#5333 and vice versa

OhYea777:
I refactored the extension logic to make it more abstract so that we aren't writing two sets of duplicate code for both the patcher and userdev extensions. This now _should_ allow for loading from multiple sources. I have made a very overly complicated project to show as an example for what this allows you to be able to do: https://github.com/OhYea777/ForgeOCEP

It seems to work for the most part, however it probably needs some further testing and verifying that it registers blocks, items etc... properly

OhYea777:
There is one issue here https://github.com/MinecraftForge/MinecraftForge/blob/1.13-pre/src/main/java/net/minecraftforge/fml/packs/ResourcePackLoader.java#L72 when the folder name is the same, it messes with the loading of their resource packs

PR #544, opened 2019-01-10
Detect and add additional jvm args to client run configs

### LINKED ISSUE #540: [FG3.0] [Mac] GLFW windows may only be created on the main thread
I encountered an issue with mac and trying to run 1.13.
When trying to start the client everything runs until the game tries to create the window and then the following crash happens:
https://haste.gorymoon.se/haqusubuvo.mccrash

The fix for this is to include `-XstartOnFirstThread` in the JVM args when starting the client.
I tested this by adding the JVM args to the task in FG3.0 in [UserDevPlugin.java](https://github.com/MinecraftForge/ForgeGradle/blob/FG_3.0/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java) and it works as intended.
       
```java
task.setJvmArgs(Collections.singleton("-XstartOnFirstThread"));
```

Because I don't know if you want to add JVM args like you do normal args and environment variables or just add a special case I didn't make a pull for it.
Tested by loading FG as a `Composite Build` into a mdk workspace.

The code in LWJGL that needs this is in [EventLoop.java](https://github.com/LWJGL/lwjgl3/blob/015c9a95b5ac98fa53e7e5ba359541b12e6798de/modules/lwjgl/glfw/src/main/java/org/lwjgl/glfw/EventLoop.java)

#### Versions
Forge: 24.0.55-1.13-pre
Mappings: 20181008-1.13
FG: Latest from FG https://github.com/MinecraftForge/ForgeGradle/commit/56559d371b402463854bf667d7f05d501cf94a38

JDLogic:
Resolves #540. Tested on mac and windows 10.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
@@ -35,9 +35,12 @@ import groovy.lang.Closure;
 import groovy.lang.MissingPropertyException;
 
 public class RunConfig implements Serializable {
+    private static final String MCP_CLIENT_MAIN = "mcp.client.Start";
+    private static final String MC_CLIENT_MAIN = "net.minecraft.client.main.Main";
     private static final long serialVersionUID = 1L;
     private String main;
     private List<String> args;
+    private List<String> jvmArgs;
     private Map<String, String> env;
     private Map<String, String> props;
     private boolean singleInstance = false;
@@ -83,6 +86,18 @@ public class RunConfig implements Serializable {
         if (args == null)
             args = new ArrayList<>();
         return args;
+    }
+
+    public void jvmArg(String value) {
+        getJvmArgs().add(value);
+    }
+    public void jvmArgs(List<String> values) {
+        getJvmArgs().addAll(values);
+    }
+    public List<String> getJvmArgs() {
+        if (jvmArgs == null)
+            jvmArgs = new ArrayList<>();
+        return jvmArgs;
     }
 
     public void singleInstance(boolean value) {
@@ -194,6 +209,11 @@ public class RunConfig implements Serializable {
         String key = value.substring(1, value.length() - 1);
         String resolved = vars.get(key);
         return resolved == null ? value : resolved;
+    }
+
+    public boolean isClient() {
+        boolean isTargetClient = getEnvironment().getOrDefault("target", "").contains("client");
+        return isTargetClient || MCP_CLIENT_MAIN.equals(getMain()) || MC_CLIENT_MAIN.equals(getMain());
     }
 
     public static class Container {
--- a/src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
+++ b/src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
@@ -61,8 +61,11 @@ import com.google.common.collect.Lists;
 
 import javax.annotation.Nonnull;
 import java.io.File;
+import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 public class PatcherPlugin implements Plugin<Project> {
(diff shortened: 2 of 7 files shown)


LexManos:
Perhaps it should directly inject into the RunConfig instead of the IDE hacks, so that the run tasks work as well.

JDLogic:
Yeah, that's a good point. I'll try that instead.

JDLogic:
Ok, the jvm args are now added to each client RunConfig. It looks like an `afterEvaluate` block is needed though if they are used in a run task defined in the build file.

~Also, I don't know why the test build failed. It should be fine.~ nvm

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -566,4 +565,32 @@ private Project getMcpParent(Project project) {
...
+        });
+
+        VersionJson json = null;
+
+        try {
+            json = Utils.loadJson(extractNatives.getMeta(), VersionJson.class);
+        }
+        catch (IOException e) {}
+
+        List<String> additionalClientArgs = json != null ? json.getPlatformJvmArgs() : Collections.EMPTY_LIST;
```
liach: `Collection.emptyList()` instead (raw types)

src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
```diff
@@ -85,6 +88,18 @@ public void setArgs(List<String> values) {
         return args;
     }
 
+    public void jvmArg(String value) {
+        getJvmArgs().add(value);
+    }
+    public void setJvmArgs(List<String> values) {
```
liach: Can you rename this to `jvmArgs` to prevent confusions on whether this clears previous args?

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -201,13 +203,26 @@ public RenameJarInPlace create(String jarName) {
...
         });
 
+        VersionJson json = null;
+
+        try {
+            json = Utils.loadJson(extractNatives.getMeta(), VersionJson.class);
+        }
+        catch (IOException e) {}
+
+        List<String> additionalClientArgs = json != null ? json.getPlatformJvmArgs() : Collections.EMPTY_LIST;
```
liach: Same rawtypes

liach (commented):
Everything else look good except a few small details.

liach (approved):

GoryMoon:
@JDLogic Noticed now when testing this out that this falsely detects 64bit intel systems as 32bits.
One of the rules that it checks is:
```json
{
  "rules": [
    {
      "action": "allow",
      "os": {
        "arch": "x86"
      }
    }
  ],
  "value": "-Xss1M"
}
```
And for my system `System.getProperty("os.arch")` returns `x86_64` that matches the rules above with the used method
```java
System.out.println(Pattern.compile("x86").matcher("x86_64").find());
> true
```

JDLogic:
@GoryMoon Yes, I did notice that during my testing. Then I launched the game with the standard Mojang launcher and found it did the same thing.
```
/Applications/Minecraft.app/Contents/runtime/jre-x64/1.8.0_74/bin/java -XstartOnFirstThread -Xss1M -Djava.library.path=/var/folders/... -Dminecraft.launcher.brand=minecraft-launcher -Dminecraft.launcher.version=2.1.1433 ...
```
Therefore, I would say that the code here is working as intended.

Pokechu22:
The older java-based launcher (which can be decompiled) effectively uses `Pattern.compile(arch).matcher(System.getProperty("os.arch")).matches()`, not `find()`.  (I don't have a link on-hand to the bootstrap portion of this, but [here's the "inner" launcher](https://s3.amazonaws.com/Minecraft.Download/launcher/launcher.pack.lzma) where this code actually is).  I don't have any information about the newer launcher, and the newer one probably should be considered authoritative, but keep this in mind.

PR #545, opened 2019-01-19
Improving configuration management for Minecraft and Deobfuscated mods

marchermans:
### Reason for making PR:
It is not possible to add a deobfuscated Minecraft (or a deobfscated Mod for that matter) to a sourceSet that does not inherit from the main sourceSet.
This is caused by the `UserDevPlugin` that only creates a single `minecraft`-and `deobf`-configuration, from both of whom the `compile`-configuration extends, causing only the `main`-sourceset (and all its depends, like the `test`-sourceset) from being able to access Forge and Minecraft classes.

Following the Java convention for API's under gradle, an API should be put in a SourceSet named `api`, from which the `compile` configuration (and thus the main sourceset) needs to extend. Due to this api sourceset now being on a lower level then the main sourceset (and its minecraft configurations) the API sourceset is not able to access Minecraft or Forge classes, even though that is in most cases required. _An example would be an interface declaring some kind of object that would need to be registered to a ForgeRegistry, other simple examples come to mind easily._

### Implemented fix:
The solution is to follow the JavaPlugin convention scheme of creating configurations.
This PR will create two configuration named as such: _<sourceSetTaskName>Minecraft or minecraft (for the main sourceset) as well as <sourceSetTaskName>Deobf or deobf (for the main sourceset)_ for each sourceSet that is added to the `sourceSets {}` section of the build.gradle.

#### Required adaptations of the normal logic, to stay within the guidelines, of one MC version per Project.
When the project is now in its `afterEvaluate` phase and the relevant callback is made, we now need to check if a person added atleast one entry to any of the `minecraftXXXX` configurations, if all the configurations, that contain a minecraft dependency, contain exactly one entry, and that all the configurations that contain exactly one entry, contain all the same entries.

### Why not make all compileXXX configurations extends minecraft?
Cause this is not up to FG3 to decide, certain sourceset configuration might not need Minecraft, so it should not be added to its dependency list.

### Example:
This can now be used as follows:
``` groovy
dependencies {
    minecraft 'net.minecraftforge.test:forge:1.13-24.0.86-1.13-pre'
    apiMinecraft 'net.minecraftforge.test:forge:1.13-24.0.86-1.13-pre'
}
```

Also tested this with multiproject builds, and those seem to load as well, after a couple of runs of `gradle eclipse --no-daemon --refresh-dependencies`.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -387,7 +387,11 @@ public class UserDevPlugin implements Plugin<Project> {
     private ExternalModuleDependency generateDeobfuscatedMinecraftDependency(MinecraftUserRepo repo, Logger logger)
     {
         final String minecraftDeobfuscatedDependencyString = repo.getDependencyString();
-        logger.info(String.format("Creating new deobfuscated minecraft dependency: %s", minecraftDeobfuscatedDependencyString));
+
+        //Log to the logger which MC has been selected for debugging purposes.
+        //Required by lex to see which version is being loaded.
+        logger.lifecycle(String.format("Creating new deobfuscated minecraft dependency: %s", minecraftDeobfuscatedDependencyString));
+
         return (ExternalModuleDependency) repo.getProject().getDependencies().create(minecraftDeobfuscatedDependencyString);
     }
 }
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -62,7 +62,6 @@ public class UserDevPlugin implements Plugin<Project> {
     public void apply(@Nonnull Project project) {
         @SuppressWarnings("unused")
         final Logger logger = project.getLogger();
-        logger.warn("Applying ForgeGradle.");
         final UserDevExtension extension = project.getExtensions().create("minecraft", UserDevExtension.class, project);
         if (project.getPluginManager().findPlugin("java") == null) {
             project.getPluginManager().apply("java");
@@ -139,8 +138,6 @@ public class UserDevPlugin implements Plugin<Project> {
                                                               .collect(Collectors.toList());
 
             final List<Configuration> validMinecraftConfigurations = this.validateMinecraftDependencies(minecraftConfigurations, p.getLogger());
-
-            logger.warn("Found: " + validMinecraftConfigurations.stream().map(Object::toString).collect(Collectors.joining(", ")) + " as configuration containing minecraft.");
 
             final MinecraftUserRepo mcrepo = generateMinecraftUserRepo(p, extension, validMinecraftConfigurations);
             final ExternalModuleDependency deobfuscatedMinecraftDependency = generateDeobfuscatedMinecraftDependency(mcrepo, logger);
@@ -316,7 +313,6 @@ public class UserDevPlugin implements Plugin<Project> {
             //This configuration (and with that its sourceSet) does not depend on minecraft skip it.
             if (information.isEmpty())
             {
-                logger.warn("Skipping configuration: " + configuration.getName() + " no minecraft dependency information has been found.");
                 continue;
             }
 
@@ -331,7 +327,6 @@ public class UserDevPlugin implements Plugin<Project> {
                   validInformation));
             }
 
-            logger.warn("Adding configuration: " + configuration.getName() + " as a configuration containing a dependency on minecraft with information: " + information);
             validConfigurations.add(configuration);
         }
 
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -26,13 +26,11 @@ import org.gradle.api.NamedDomainObjectFactory;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
-import org.gradle.api.artifacts.Configuration;
-import org.gradle.api.artifacts.Dependency;
-import org.gradle.api.artifacts.DependencySet;
-import org.gradle.api.artifacts.ExternalModuleDependency;
+import org.gradle.api.artifacts.*;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.JavaExec;
+import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.bundling.Jar;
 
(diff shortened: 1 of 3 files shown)


marchermans:
As per request i clarified where the logging line from https://github.com/MinecraftForge/ForgeGradle/pull/545/files#diff-df0afbd40bfef0d804b53226a7095994L149 went: https://github.com/MinecraftForge/ForgeGradle/pull/545/files#diff-df0afbd40bfef0d804b53226a7095994R393

Attached to this comment is also a build.gradle file for an example project. This gradle file works with the default MDK project and creates an API sourceset that can access deobfuscated minecraft classes.
[build.zip](https://github.com/MinecraftForge/ForgeGradle/files/2798789/build.zip)

gabizou:
Doesn't seem to work if you have a project and subproject depend on Minecraft (whether one has one set of AT's and the other has more AT's for example) that are not straight dependencies like debobf dependencies.

gabizou:
Tested locally with some expansion mod that depended on the parent project with some changes (like adding more AT's on top of the parent's AT's). When going in vice versa (where parent depends on sub project that has some AT's + forge), understandably, FG is recreating the dependencies for both (since both have their hashes changed with different AT entries for both). I haven't had any issues except occasionally needing to refresh a few times.

I've even tested cases where one of the projects does NOT depend on Minecraft/Forge and it will still function normally. The only case where I feel there may be issues is when a project that depends on another project that provides the Minecraft dependency would potentially end up with an unresolved dependency in IDE, even though the dependency is provided for that project.

image

root build.gradle:
```

buildscript {
    repositories {
        mavenLocal()
        maven { url = 'https://files.minecraftforge.net/maven' }
        jcenter()
        mavenCentral()
    }
    dependencies {
        classpath group: 'net.minecraftforge.gradle', name: 'ForgeGradle', version: '3.+'
    }
}
apply plugin: 'java'
//apply plugin: 'net.minecraftforge.gradle'
apply plugin: 'eclipse'

ext.expansion = project('expansion')
version = '1.0'
group = 'com.yourname.modid' // http://maven.apache.org/guides/mini/guide-naming-conventions.html
archivesBaseName = 'modid'

sourceCompatibility = targetCompatibility = compileJava.sourceCompatibility = compileJava.targetCompatibility = '1.8' // Need this here so eclipse task generates correctly.
//
//minecraft {
//    mappings channel: 'snapshot', version: '20180921-1.13'
//    accessTransformer = file('common_at.cfg')
//    runs {
//        client = {
//            workingDirectory project.file('run').canonicalPath
//            source sourceSets.main
//        }
//        server = {
//            workingDirectory project.file('run').canonicalPath
//        }
//    }
//}

dependencies {
    compile(expansion)
//    minecraft 'net.minecraftforge.test:forge:1.13-24.0.116-1.13-pre'
}
```

expansion build.gradle:
```

apply plugin: 'java'
dependencies {
    minecraft 'net.minecraftforge.test:forge:1.13-24.0.116-1.13-pre'
//    compile(parent.project)
}

minecraft {
    mappings channel: 'snapshot', version: '20180921-1.13'

//    accessTransformer(parent.project.file('common_at.cfg'))
    accessTransformer(file('expansion_at.cfg'))
    runs {
        client = {
            workingDirectory project.file('run').canonicalPath
            source sourceSets.main
        }
        server = {
            workingDirectory project.file('run').canonicalPath
        }
    }
}

```

Naturally, this is with the root `settings.gradle` having an `include 'expansion'`
and the project file layout is like so:
layout

At the end of the day, this may be some kerfuffling with IntelliJ/gradle/caching that I'd need to clear out and restart the workspace again, but if it's testable and reproducible, this may be a corner case that won't be seen in the community, if at all.

marchermans:
@gabizou What is the exact issue?

gabizou:
@OrionDevelopment It's unable to resolve the runtime dependency of `net.minecraftforge.test:forge:1.13-24.0.116-1.13....` for the parent project. Being that it's a runtime dependency, I'm not sure if that's a different handling altogether from the artifact creation that this PR does achieve.

marchermans:
This PR did not change anything when it comes multi-project workspaces.
This allows you to add minecraft to a different sourceset, it does not deal with any multi-project related properties.


I tested multi-project workspaces just to be sure, but send me your project that you tested this with and i will try to fix it.

gabizou:
[forge-1.13-24.0.116-1.13-pre-mdk.zip](https://github.com/MinecraftForge/ForgeGradle/files/2804715/forge-1.13-24.0.116-1.13-pre-mdk.zip)


> I tested multi-project workspaces just to be sure, but send me your project that you tested this with and i will try to fix it.

It's definitely an anomaly in that the root project is depending on a sub project that includes the Minecraft dependency (forge in this case) while the root project doesn't have a minecraft dependency of it's own. In normal circumstances, I'd wager there's no shortage of projects that always have a minecraft dependency of their own when using FG.

marchermans:
Actually that was the case i tested
Root (without) splitted into two subprojects: first and second
First depends on minecraft with its api and main sourcesets
Seconds depends on minecraft with its main sourceset and on the api sourceset of the first

Will test tomorrow

marchermans:
@gabizou I tested your project.
A simple run: `gradle eclipse --no-daemon --refresh-dependencies` and then importing or refreshing the project in IntelliJ works without an error.
If i import it into IntelliJ directly without making a call that holds the `--no-daemon` flag does not work, and has never worked for me, on idea why.

Maybe somebody else can explain that for me and you?

gabizou:
While I still couldn't get IntelliJ to resolve the runtime dependency (since the `:expansion` project is a compile time dependency of the parent project, the dependencies of `:expansion` become runtime dependencies of the parent project), which is what the screenshots were trying to demonstrate, but I don't think this is an issue that the PR had altered in any way from previous behaviors. 

@LexManos I'd say this PR is fine to merge for what it's achieving with multiple source sets having minecraft dependencies. I'll try and narrow down my corner case when I have a chance and if it's necessary to resolve in the future, there's other things that take priority over this exception.

Raycoms:
@LexManos can this be merged?

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -251,4 +257,141 @@ private void createRunConfigsTasks(@Nonnull Project project, ExtractNatives extr
...
+     *
+     * @param project The project to generate for.
+     * @param extension The extensions with the project information.
+     * @param validMinecraftConfigurations The valid configurations.
+     * @return The minecraft user repo.
+     */
+    private MinecraftUserRepo generateMinecraftUserRepo(Project project, UserDevExtension extension, List<Configuration> validMinecraftConfigurations)
+    {
+        //We can do this since the configurations have been validated.
+        final Dependency dependency = new ArrayList<>(validMinecraftConfigurations.get(0).getDependencies()).get(0);
```
LexManos, line 376: Is there something that prevents them from setting the wrong value for the different configurations?
This just picks the first and goes with it. The first is whatever happens to be in their build.gradle.
What is the benefit of this, over being able to just have configurations.api.extendsFrom(configurations.minecraft) ?

Issue #550, opened 2019-02-05
Import hanging


kashike:
The project gets stuck synching - it has been running for nearly an hour at this point:
![x](https://i.imgur.com/YmuR9qf.png)

```
Parallel execution is an incubating feature.

> Configure project :SpongeCommon
New Dep: net.minecraft:server:1.13_mapped_snapshot_20180921-1.13_at_b27072188f97db9a65ca30c03163920147b4c84d

> Configure project :
New Dep: net.minecraftforge.test:forge:1.13-24.0.116-1.13-pre_mapped_snapshot_20180921-1.13_at_b27072188f97db9a65ca30c03163920147b4c84d

Deprecated Gradle features were used in this build, making it incompatible with Gradle 5.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/4.9/userguide/command_line_interface.html#sec:command_line_warnings

CONFIGURE SUCCESSFUL in 5s
```

It is "solved" by following these steps, though it may take more than one try:
1. Force-stop the refresh process in IntelliJ
2. Run `./gradlew build --refresh-dependencies`
3. Refresh IntelliJ project

LexManos:
IntelliJ has a habit of assigning very little ram and using random daemons for Gradle projects.
If you can find a way to debug this, and add more ram. Or at the very least post the console/build log 
 (one that isn't 1/2 deleted because gradle has a fancy self-deleting interface) instead of just random IntelliJ screenshots as they arnt helpful.

kashike:
There's no way to tell IntelliJ to use `--debug` or anything else, and it only hangs within IntelliJ. I don't use the default RAM, I have it set to a larger value. There isn't anything else I really *can* provide. :/

gabizou:
@kashike configure global gradle config to enable debugging, and you can attach a debugger (remote application) to debug itself), or configure the global gradle properties in your user folder to enable finer logging.

Aaron1011:
I'm experiencing the same issue.

Aaron1011:
I just got a stacktrace from the hanging gradle process: https://gist.github.com/Aaron1011/5328492a0aaa0e036e88d575d8f00ee6

Aaron1011:
This seems wrong: https://github.com/MinecraftForge/ForgeGradle/blob/FG_3.0/src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java#L79

[...]

PR #553, opened 2019-02-07
Fix issues with dependency resolution

Aaron1011:
This PR fixes several issues with dependency resolution:

* The deadlock from issue #550
* The UserDevPlugin crash from issue #551 
* An unnecessary computation of `version`, which could result in an `IndexOutOfBoundsException`

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
@@ -103,7 +103,6 @@ public class MinecraftRepo extends BaseRepo {
         String version = artifact.getVersion();
         String mappings = getMappings(version);
         if (mappings != null) {
-            version = version.substring(0, version.length() - mappings.length() + "_mapped_".length());
             return null; //We do not support mappings
         }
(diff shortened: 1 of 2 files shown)


LexManos:
the better fix would be to https://github.com/MinecraftForge/MCPConfig to define a config that has sided injects. Instead of just reverting something and calling it a 'fix'
Also 'files' needs a better name, and associated comments explaining the deadlock so we dont have to dig the git history for why it's written this way.
And again, you hardcode the client into those deps, {or more precisely I did, but this is a good enough time to fix it} It should be the requested side.

Aaron1011:
@LexManos: [This comment](https://github.com/MinecraftForge/ForgeGradle/blob/57aad122cc4b02cd6a8c660cf11be5cc29edead4/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java#L351) seems to indicate that `:client` is deliberately used for the ':extra' and ':data' dependencies. Should I change that as well?

Aaron1011:
Superseded by https://github.com/MinecraftForge/ForgeGradle/pull/554

PR #554, opened 2019-02-09
Ensure that gradle tasks are not run during dependency resolution

Aaron1011:
As described in
https://github.com/MinecraftForge/ForgeGradle/issues/550,
running Gradle tasks from a dependency resolution can easily lead to a
deadlock.

To avoid this issue, we perform all of the gradle task-related actions -
decompiling, recompiling, etc - up front, during plugin initialization.
This is accomplished by explicitly invoking BaseRepo#getArtifact for all
of the possible 'generated' artifacts. This ensures that all of the
caches are populated by the time dependency resolution actually runs.
With all requests for 'generated' files being served from various
caches, there's no longer any need to execute any new gradle tasks
during dependency resolution.

To ensure that the dependency resolver isn't actually generating any new
files, we set a flag 'alreadyBuiltDependencies' after the initial plugin
configuration. This flag is checked whenever a file is about to be
generated due to a cache miss - if it's set, an error is logged to
indicate the bug.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
@@ -83,6 +83,7 @@ public class MinecraftUserRepo extends BaseRepo {
     private MCP mcp;
     @SuppressWarnings("unused")
     private Repository repo;
+    private boolean alreadyBuiltDependencies = false;
 
     /* TODO:
      * Steps to produce each dep:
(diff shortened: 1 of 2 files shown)


Aaron1011:
Should I close this, or are there changes that I can make in order to make this mergeable?

LexManos:
as stated on discord, this will never be mergeable as long as your 'solution' is to resolve every possible artifact always.


If the deadlock is due to running gradle tasks inside the repos, then we've already discussed the solution. The only gradle task that is run is in compileJava. Replacing that should suffice for what you need.

Aaron1011:
Superseded by https://github.com/MinecraftForge/ForgeGradle/pull/555

PR #555, opened 2019-02-11
Fix depedency resolution issues, take 3

Aaron1011:
This PR fixes the deadlock issue in https://github.com/MinecraftForge/ForgeGradle/issues/550.

I've also fixed an issue with client-only injections - the previous fix didn't apply to MinecraftUserRepo

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
@@ -113,6 +113,7 @@ public class MinecraftUserRepo extends BaseRepo {
     private MCP mcp;
     @SuppressWarnings("unused")
     private Repository repo;
+    private Set<File> extraDataFiles;
 
     /* TODO:
      * Steps to produce each dep:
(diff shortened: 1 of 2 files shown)


src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
```diff
@@ -501,6 +531,18 @@ private File findRaw(String mapping) throws IOException {
                 while ((entry = zin.getNextEntry()) != null) {
                     if (!entry.getName().startsWith(prefix) || entry.isDirectory())
                         continue;
+
+                    // If an entry has a specific side in its name, don't apply
+                    // it when we're on the opposite side. Entries without a specific
+                    // side should always be applied
+                    if ("server".equals(NAME) && entry.getName().contains("/client/")) {
```
LexManos, line 598: Patchers are only joined. So this filter is unneeded. The InjectFunction is run on the MC code itself so that filter works fine on this repo as well.

Aaron1011: This filter is definitely necessary - I get a compilation error when using `UserDevPlugin` otherwise. The zip entries come directly from the `mcp` jar, which contains a [Start](https://github.com/MinecraftForge/MCPConfig/blob/2d1b27131651897791bdee0f1574f76467f7ebb0/versions/1.13/inject/mcp/client/Start.java) class that we don't want to inject into a server jar.

LexManos: Actually got time to verify this, and I was wrong I thought you were in the other loop where we inject the patcher's code.
This is just for the MCP jar so the filter is fine.

src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
```diff
@@ -973,7 +1072,42 @@ private File compileJava(File source, File... extraDeps) {
...
+                @Nullable
+                @Override
+                public OriginTaskExecutionMetadata getOriginExecutionMetadata() {
+                    return null;
+                }
+            });
+
+            Method compileMethod = JavaCompile.class.getDeclaredMethod("compile");
+            compileMethod.setAccessible(true);
+            compileMethod.invoke(compile);
```
LexManos: Considering the compile function is protected, we could make a subclass of JavaCompile and hard link it so that we get a compile error instead of runtime error if Gradle changes?

src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
```diff
@@ -176,6 +207,64 @@ public void validate(Configuration cfg, Map<String, RunConfig> runs, File native
...
+     * code will only be executed once, instead of during every call to
+     * compileJava
+     * @return
+     */
+    private Set<File> buildExtraDataFiles() {
+        Configuration cfg = project.getConfigurations().create(getNextCompileName());
+        List<String> deps = new ArrayList<>();
+        deps.add("net.minecraft:client:" + mcp.getMCVersion() + ":extra");
+        deps.add("net.minecraft:client:" + mcp.getMCVersion() + ":data");
+        deps.addAll(mcp.getLibraries());
```
LexManos, line 260: Just as a note, as I was the one to originally hardcode 'client'. It is it necessary to use the client, or should this be moved to the proper side as well? So Server only stuff gets the proper libraries.

Also, speaking with cpw, there is a issue with dedicated server and log4j with extra/data being separate {Log4j tries to grab the custom logger class from the same jar as the config, instead of from the classpath}. So it's on my TODO to merge the 'extra' and 'data' deps into just a single 'extra' jar.

PR #559, opened 2019-02-18
Replacing the raw main jar with the recompiled jar, when sources are requested.

marchermans:
The sources jar is generated from the recompiled jar instead of the raw jar.
This causes discrepancies when an IDE then tries to find the sources of certain classes with in that sources-main jar combination.

To fix these discrepancies the main raw jar is replaced with the recompiled jar, this way the sources jar, which is generated from the recompile jar is in sync.

Several precautions have been taken to prevent problems in a threaded environment:
A external FileLock is created to prevent other programs from modifying the File externally. And an in repo lock object is created to prevent the repo from writing to that jar in a multi threaded in jvm scenario.

I ran this solution for a couple of hours continuously creating and recreating, both sources and jars to figure out if this is causing any issues, so far none have surfaced, however i am still not sure if this is enough, and input on the solution used to replace the jars is highly welcome.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
@@ -74,6 +74,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Method;
+import java.nio.channels.FileLock;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


LexManos:
This is no longer necessary, I found the root cause. cpw was ALWAYS loading the raw jar in debug. So the class files were mismatched vs the sources. My change to find Forge via the ClassLoader will make it find whichever version is linked against. Which should be -recomp

marchermans:
Okey perfect.

PR #562, opened 2019-02-20
[FG_3.0] Remove dependency on 'eclipseClasspath' for 'genEclipseRuns' task

OhYea777:
These changes should allow FG to do what it currently does, with `genEclipseRuns`, but without the dependency on the `eclipseClasspath` task

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/EclipseHacks.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/EclipseHacks.java
@@ -27,6 +27,8 @@ import net.minecraftforge.gradle.common.task.DownloadAssets;
 import net.minecraftforge.gradle.common.task.ExtractNatives;
 import org.apache.commons.io.IOUtils;
 import org.gradle.api.Project;
+import org.gradle.api.Task;
+import org.gradle.api.tasks.TaskProvider;
 import org.gradle.plugins.ide.eclipse.GenerateEclipseClasspath;
 import org.xml.sax.SAXException;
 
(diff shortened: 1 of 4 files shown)


OhYea777:
Added a fix to clone the tokens map when setting tokens, accidentally forgot this earlier so it is overwriting it for all runs

LexManos:
Why? The hacks we do to the classpath is 1/2 the reason we have this code. And what use are runs without the workspace?

tterrag1098:
The eclipse gradle plugin will suffice to generate the project. All this dependency served to do is make it so your classpath was overwritten when all you wanted was to generate run configs. Of course it's still totally possible to add the eclipse plugin and run the `eclipse` task (you could even make `eclipse` depend on `genEclipseRuns` in your own buildscript).

OhYea777:
Yeah, this change will let you regen runs without wiping your entire classpath to its original form

LexManos:
As long as it works, then i'm fine with it. The need to nuke the classpath was more to make sure that it existed so we could edit it in the hacks method.

OhYea777:
Okay, so the issues that this PR fixes is:
- The `genEclipseRuns` task currently depends on the `eclipseClasspath` task, the problem with this is that every time the `eclipseClasspath` task is ran it nukes the classpath back to its original form
- When it sets the `MOD_CLASSES` environment variable, during the `genEclipseRuns` task, it writes the directories 4 separate times, e.g.:
``examplemod%%/bin/main;examplemod%%/bin/main;examplemod%%/bin/main;examplemod%%/bin/main``
This is because I had made an initial fix, then a secondary better fix, for the issue of Forge requiring the first (left side), to be a resource directory and then the rest (right side) to be the classes and any other directories. Having only a single directory only allows for scanning of resources and not classes
- Currently all **RunConfig**'s are passed a reference to the environment tokens used for replacement in vars. The issue with this is that, when they modify the tokens, the tokens are modified for any future **RunConfig**'s that use them

These are the changes I made in this PR to resolve these issues:
- I removed `genEclipseRuns`' dependency on the `eclipseClasspath` task. This works but, because of the way Eclipse works, only for source sets whose source directories already exist, currently the `eclipseClasspath` task has an additional hack to create these dirs. To fix this secondary issue I created a new, separate, task `makeSrcDirs` that ensures all directories for the defined source sets exist. I then added this as a dependency to the `eclipseClasspath` and `genEclipseRuns` tasks
- I removed the initial fix I had made to setting dual copies of directories in **MOD_CLASSES**, during `genEclipseRuns`, it now only appends two directories to satisfy Forge requiring _BOTH_ a resources and classes directory in order to scan properly
- Now when you set the tokens of a **RunConfig** it sets it as a clone of the tokens, this way it no longer mutates the original

OhYea777:
This _should_ also fix MinecraftForge/MinecraftForge#5556

PR #567, opened 2019-03-02
Ensure that the JDK version is within the supported range

gigaherz:
Minimum required is set to 1.8.0_52, up to but **not** including 11.0.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,6 +64,7 @@ license {
     }
     exclude '**/*.properties'
     exclude 'net/minecraftforge/gradle/common/diff/'
+    exclude 'net/minecraftforge/gradle/common/util/JavaVersionParser.java'
 }
 
 wrapper {
--- a/src/common/java/net/minecraftforge/gradle/common/util/JavaVersionParser.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/JavaVersionParser.java
@@ -1,0 +1,268 @@
+/*
+ * Copyright 2015-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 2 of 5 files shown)


ichttt:
Actually minimum is 101 as this is the first one that ships the LE certificates that forge file server uses

gigaherz:
Oh! Lex mentioned 52, I guess neither of us thought of Let's Encrypt. Fixing.

drekryan:
Actually the following commit fixed this in Forge so honestly a minimum of 52 (which MC ships with) is fine if I understand correctly. https://github.com/MinecraftForge/MinecraftForge/commit/d9c4446ccc0931ba1fdbbc5156c60740b97dea6e

ichttt:
This is ForgeGradle not Forge itself. FG also needs to pull stuff from the files server. So no, it needs at least 101 to use https to forge maven

bs2609:
To make things clear, the vanilla launcher downloads 8u**51**. A subversion of 52 does not correspond to any actual Java 8 release from Oracle (e.g. https://www.oracle.com/technetwork/java/javase/documentation/8u-relnotes-2225394.html).

gigaherz:
Oh so it is u51.... could have sworn it was 52. 

Remember that you can still use u51 in your run/launch configs, separate from the gradle jdk, if you are in a situation where u51 is needed. It won't work for `gradlew runClient`, though, but that's a very minor annoyance.

In any case, if Lex prefers it, I can set it to 51 and try to make use of that FixSSL hack within ForgeGradle.

LexManos:
i said 52 {meant 51} because Mojang ships 51. IF they get to this code. Then they have already gone past the LetsEncrypt issue. As they have downloaded the FG code.

PR #569, opened 2019-03-04
Add check for compatible Gradle version range.

gigaherz:


LexManos:
Have you tried pulling in a dynamic AF version based on the gradle version? If we can get that working we can expand the range. As it should only be AF that's the issue between versions.

gigaherz:
Oh right you did mention that. I'm not sure it's possible to pick different dependency versions for the buildscript, but I'll try.

LexManos:
Well it wouldnt be a different build script, it would be dynamically pulling in and loading the library.

covers1624:
Can you rebase this onto the `FG_4` branch and force Gradle 6.8+.

sciwhiz12:
The feature added by this PR is already implemented and merged in by #754.

Issue #574, opened 2019-03-16
Use compileOnly scope instead of compile for Minecraft and Forge dependencies


Cypher121:
According to Gradle's notes on the compileOnly scope (https://blog.gradle.org/introducing-compile-only-dependencies), one of the use-cases for it is as follows: 

> Dependencies whose API is required at compile time but whose implementation is to be provided by a consuming library, application or **runtime environment**.

Last part fits forge dependency for mods perfectly.

As it stands right now, every library and mod declares forge as a compile dependency, which means 2 things:
* It is declared a transitive dependency, causing mods to add it in addition to its own version of forge, unless one of them is explicitly removed.
* It is declared a runtime dependency, which somewhat confusingly means mods are expected to provide it themselves at runtime. That means shadow jar will by default attempt to include forge and all of its dependencies in the mod. That behavior is not easy to avoid, as it is supposed to be avoided by having dependency marked compileOnly

Please consider putting moving the minecraft configuration from compile to compileOnly.

P.S. https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph is a good reference on how dependency scopes are set up for java projects. Note that compile and runtime scopes are not there at all, as they have been deprecated since Gradle 3.0

Niko-sk2x:
Except that using compile only dependency here means that IDEs won't add Forge and Minecraft to run classpath. Because it's expected to be provided by something else.

While it's technically correct, it's a very useless kind of correct, because it would make it very hard to actually run mods in dev environment. The linked documentation even explicitly states that compileOnly dependencies won't be included in runtime classpath. And for the mentioned use case:
```

    Dependencies required at compile time but never required at runtime, such as source-only annotations or annotation processors;
    Dependencies required at compile time but required at runtime only when using certain features, a.k.a. optional dependencies;
    Dependencies whose API is required at compile time but whose implementation is to be provided by a consuming library, application or runtime environment.
```
While it technically fits the last one, I think it definitely goes the intention that can be seen in the previous 2 use cases.

And yes some IDEs have an option to include "provided" dependency scope in run classpath. But this means it would no longer be possible to have other mods as compileOnly dependency.

It's far easier to just get around shadowJar defaults once, than it is to get around IDEs each time you set up dev environment.

Cypher121:
[...]

Issue #571, opened 2019-03-09
[FG3] Deobf dependencies not resolved


Choonster:
After adding a `deobf` dependency like this in my **build.gradle**:
```gradle
dependencies {
    minecraft "net.minecraftforge:forge:${mc_version}-${forge_version}"

    deobf "mcp.mobius.waila:Hwyla:1.10.1-B53_1.13.2"
}
```

I get the following error when refreshing the Gradle project:
```text
Could not resolve: mcp.mobius.waila:Hwyla:1.10.1-B53_1.13.2_mapped_snapshot_20190308-1.13.1
```

Adding the same dependency as `compile` rather than `deobf` works without issue (apart from it still using SRG names).

LexManos:
Next time please provide the report for that custom dependency so I can properly test. But this should fix it.

Choonster:
Thanks for the fix, but I don't understand exactly you wanted me provide. Did you want a full reproduction, i.e. a full **build.gradle** that reproduces this problem?

Choonster:
I don't think this is working. You can see a **build.gradle** file to reproduce this and the relevant logs [here](https://gist.github.com/Choonster/441ed47db1ce715674ca45b2ea51d151).

* I renamed the **~/.gradle** directory to ensure that it wasn't a cache issue.
* I downloaded a fresh MDK (1.13.2-25.0.85) and added Hwyla as a deobf dependency in **build.gradle**. This can be seen in the Gist.
* I imported the new project into IDEA. This is the first log in the Gist.
* The import finished with an error, so I refreshed the project in IDEA's Gradle window. This is the second log in the Gist.
* Just to be sure, I ran `.\gradlew --refresh-dependencies --debug`. This is the third log in the Gist, though it doesn't seem to mention Hwyla at all.

At the end of this, Hwyla still can't be resolved:
![image](https://user-images.githubusercontent.com/878073/54487054-f3b54f80-48e4-11e9-911a-4ed0ea391eb7.png)

EdwinMindcraft:
Had the same issue, try adding a / at the end of your repo declaration
`maven { url = "http://dvs1.progwml6.com/files/maven" }`
becomes
`maven { url = "http://dvs1.progwml6.com/files/maven/" }`

Choonster:
[...]

Issue #369, opened 2016-05-12
deobfCompile/deobfProvided dependency source attachment issue


matthewprenger:
It looks like when you have a deobf\* dependency that has a classifier, like -api, the sources don't get downloaded and attached in idea like they would with the regular artifact.

mezz:
hi @DaemonUmbra, I see this is closed, was it fixed at some point or was it closed as part of a github issues cleanup?

DaemonUmbra:
Closed as part of cleanup, feel free to reopen if you want

mezz:
Hmm actually, @tterrag1098 did your commit (https://github.com/MinecraftForge/ForgeGradle/pull/637) fix this issue?

tterrag1098:
No, that PR was reverted. Pretty sure this is fixed in newer FG versions though.

PR #577, opened 2019-03-18
Implement deobf dependencies for all configurations

Cypher121:
Reimplement the way deobfuscated dependencies are handled. The main goal here is allowing users to put deobf dependencies in any scope (#574), instead of just `compile`. For example:

```gradle
dependencies {
    compileOnly fg.deobf("some:optional:dependency")
    api fg.deobf("base:library:1.2.3")
    implementation fg.deobf(files("utility.jar"))
}
```

Additional changes:

* Obfuscated dependencies are now resolved by gradle via a custom configuration, FG only deobfuscates them. As without those MavenArtifactDownloader only handles limited numbers of dependencies, like MCP mappings, it no longer needs to perfectly re-implement every feature of gradle's resolver, so issues like #571 shouldn't happen nearly as often.

* Due to above we can now handle most types of dependencies: the obfuscated artifact is loaded for us by gradle, so we only need to know how to remap it. Currently supports external artifacts (maven, ivy, client modules (not sure how well)) and local files, so vast majority of usecases.

* `deobf` configuration is still intact, now rerouted through the new process, so it supports all the new features as well.


src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
```diff
@@ -0,0 +1,45 @@
...
+
+package net.minecraftforge.gradle.userdev;
+
+import groovy.lang.GroovyObjectSupport;
+import net.minecraftforge.gradle.userdev.util.DependencyRemapper;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Dependency;
+
+public class DependencyManagementExtension extends GroovyObjectSupport {
+    public static final String EXTENSION_NAME = "fg";
```
Cypher121, line 29: Putting `deobf(...)` on top level would make it ambiguous with the deobf configuration. Similarly, the minecraft extension is also ambiguous with minecraft configuration. For now, made it `fg.deobf(...)`. Any other suggestions?

src/userdev/java/net/minecraftforge/gradle/userdev/ModRemapingRepo.java
```diff
@@ -1,241 +0,0 @@
...
-import net.minecraftforge.gradle.common.util.MavenArtifactDownloader;
-import net.minecraftforge.gradle.common.util.McpNames;
-import net.minecraftforge.gradle.common.util.SrgJarRenamer;
-import net.minecraftforge.gradle.common.util.Utils;
-
-/*
- * Takes in SRG names jars/sources and remaps them using MCPNames.
- * TODO: Direct file dep support?
- */
-public class ModRemapingRepo extends BaseRepo {
```
LexManos, line 52: did git mess up? This should be rename/modify not remove/add

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -154,18 +172,21 @@ public RenameJarInPlace create(String jarName) {
                 minecraft.getDependencies().add(ext);
             }
 
-            deps = deobf.getDependencies();
-            for (Dependency dep : deps.stream().collect(Collectors.toList())) {
-                if (!(dep instanceof ExternalModuleDependency)) //TODO: File deps as well.
-                    throw new IllegalArgumentException("deobf dependency must be a maven dependency. File deps are on the TODO");
-                deps.remove(dep);
+            DependencySet legacyDeps = deobfConfiguration.getDependencies();
```
LexManos: If this is a legacy list, then add a //TODO: Remove FG3.1 or something so we can remove it. I'd rather not have a long lasting list of depreciated code.

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
```diff
@@ -0,0 +1,148 @@
...
+        project.getLogger().debug("Deobfuscating binary file {} with mappings {}", original.getName(), mappings);
+
+        File names = findMapping(mappings);
+        if (names == null || !names.exists()) {
+            return null;
+        }
+
+        File cacheDir = getOrCreateCacheDir(cachePath);
+
+        String mappedName = addMappingsToName(original.getName(), mappings);
```
LexManos: This is nothing but asking for collisions by not supporting unique identifiers, this is the main reason why I never bothered to add direct file dep support.
Everything should go through ivy/maven format.

src/userdev/java/net/minecraftforge/gradle/userdev/util/DependencyRemapper.java
```diff
@@ -0,0 +1,92 @@
...
+        DefaultSelfResolvingDependency newDep = new DefaultSelfResolvingDependency(files);
+        mappingListeners.add(m ->
+                dependency.getFiles().getFiles().forEach(f ->
+                        {
+                            try {
+                                //use hash of the directory path to prevent name collisions, eg. lib1/output.jar and lib2/output.jar
+                                if (f.getName().endsWith("sources.jar")) {
+                                    files.from(deobfuscator.deobfSources(f, m,  "local_file", HashFunction.SHA1.hash(f.getParentFile().getAbsolutePath()), f.getName()));
+                                } else {
+                                    files.from(deobfuscator.deobfBinary(f, m, "local_file", HashFunction.SHA1.hash(f.getParentFile().getAbsolutePath()), f.getName()));
```
LexManos: Hash of the file contents would probably be better then the random/arbitrary path.
It's crappy, but it's what Gradle does internally.

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
```diff
@@ -0,0 +1,148 @@
...
+
+            Utils.updateHash(output, HashFunction.SHA1);
+            cache.save();
+        }
+
+        return output;
+    }
+
+    public File deobfSources(File original, String mappings, String... cachePath) throws IOException {
+        project.getLogger().debug("Deobfuscating sources file {} with mappings {}", original.getName(), mappings);
```
LexManos: This entire class seems to be duplicated code, You should prevent that duplication.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -49,25 +52,29 @@
...
         @SuppressWarnings("unused")
         final Logger logger = project.getLogger();
-        final UserDevExtension extension = project.getExtensions().create(UserDevExtension.class, UserDevExtension.EXTENSION_NAME, UserDevExtension.class, project);
+        final UserDevExtension extension = project.getExtensions().create(UserDevExtension.EXTENSION_NAME, UserDevExtension.class, project);
 
         if (project.getPluginManager().findPlugin("java") == null) {
             project.getPluginManager().apply("java");
         }
-        final File natives_folder = project.file("build/natives/");
+        final File nativesFolder = project.file("build/natives/");
```
LexManos, line 66: 0.o

LexManos:
Honestly, I would prefer we DONT give the controls back to gradle to download things. Because it's been known to be horrible. But I can see the benefits of generically supporting more repo formats. Local files also can be done via maven repos as is. And doing them directly means we need to make sure that collisions are prevented. As I said in my comment, best way would be by hashing the file contents. Tho it hurts organization.

We also need example usages, as well as if the old system is broken or not.

src/userdev/java/net/minecraftforge/gradle/userdev/ModRemapingRepo.java
Cypher121: Yeah, git can't handle rename + changes in one commit that well, and I forgot about that.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
Cypher121: Good 0.o or bad 0.o?

Cypher121:
Haven't really had issues with gradle downloading dependencies, except for the fact there aren't many ways to properly tell it to do so (hence the obf configuration). The way I see it, FG only adds mappings to things here, so for downloading `deobf "1:2:3"` and `scope fg.deobf("1.2.3")` should be as close to identical as possible. If it is identical, we can focus on the mapping process and not worry about debugging the downloads. If something breaks, it'd break with just gradle by itself, so that's gradle's fault.

As for the current system being broken, here's two examples, one for downloading, another for scopes:

[This project](https://github.com/DarkRoleplay/Dark-Roleplay-Medieval/tree/experimental/1.13.3) cannot build as it is, because MavenArtifactDownloader.manual is having trouble resolving the snapshot artifacts. Works fine with just compile dependencies, so letting gradle download it fixes the error. As such, I had no issues building it after switching to my build of FG with these changes. Now we could find the cause and update the downloader to handle that edge-case, but gradle already does it, so I really see no need.

As for scopes, here's how JEI docs recommend people to set up their dependencies for FG 2.x:
```gradle
dependencies {
  // compile against the JEI API but do not include it at runtime
  deobfProvided "mezz.jei:jei_${mc_version}:${jei_version}:api"
  // at runtime, use the full JEI jar
  runtime "mezz.jei:jei_${mc_version}:${jei_version}"
}
```
Meanwhile on FG3, the only possible configuration is
```gradle
dependencies {
  deobf "mezz.jei:jei_${mc_version}:${jei_version}" //equivalent to deobfCompile
}
```
Since we can't split that dependency into provided and runtime, we can't split it into api and full mod either, we have to go with one or the other for both compilation and runtime. If we go with API, the API will be loaded into runtime in userdev. We won't be able to add the actual mod in run/mods either, as it contains the API as well, so that's duplicate classes. So we have to use the full version for both compilation and runtime, polluting the compile classpath with JEI's internals, and making it hard to launch userdev without loading JEI (have to remove it from runtimeClasspath manually), so no testing whether your mod works when it's absent.

With the new configuration it's as easy as
```gradle
dependencies {
  compileOnly fg.deobf("mezz.jei:jei_${mc_version}:${jei_version}:api")

  runtimeOnly "mezz.jei:jei_${mc_version}:${jei_version}"
}
```
or just `compileOnly` and put the full jar in the mods.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
Cypher121: Added a log warning when configuration is used and TODO

MajorTuvok:
Sorry for the interruption from someone with not that much gradle experience, but isn't one supposed to avoid (it's deprecated not?) the compile dependency as of Gradle 3.0, and we are working with 4.9 here...?

Cypher121:
@MajorTuvok `compile` is deprecated, `compileOnly` isn't. current deobf configuration uses `compile` under the hood, so it'd be better to gradually get rid of it and switch to new configurations as described [here](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph)

TL;DR, `compile` used to mean "compile dependency, runtime dependency, and a transitive dependency for everyone using this project", while `compileOnly` just means "compile dependency"

MajorTuvok:
Thanks for clarifying @Cypher121 :). Sorry, that I didn't know that :(...

marchermans:
@Cypher121 I totally agree with this pr. On several points even.

Yes, we should never have handled file downloads ourselfs for anything. Regardless, if we had the time and with 2020 hindsight we should have defined repos for every single core artifact and a processor and or repo processor combo for anything that needs local processing. That said, what we have right now is a compromise and your pr is definitely an improvement in my opinion.

Secondly adding the other deobf configurations can be done much better and with a way that follows the general guidelines of how Java project plugins should create a configuration (using sourceset.forall()), although my pr actually was not needed with its initial goal, I created one that had a similar result, I am on mobile now so I can't find the damn number but it should be easy to find.

Tldr: agree with this change, but can be done without a major change to anything by following convention.

marchermans:
It is pr #545

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
LexManos: Bad 0.o, why change names it's just midding your commit.

Cypher121: Was named in snake case for some reason, so I decided it's a small enough change to just do it along the way. Want me to revert that?

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
```diff
@@ -0,0 +1,150 @@
...
+        String channel = mapping.substring(0, idx);
+        String version = mapping.substring(idx + 1);
+        String desc = "de.oceanlabs.mcp:mcp_" + channel + ":" + version + "@zip";
+        return MavenArtifactDownloader.manual(project, desc, false);
+    }
+
+    private static String getUniqueMappedName(File file, String mappings) throws IOException {
+        String originalBaseName = file.getName().substring(0, file.getName().lastIndexOf('.'));
+        String originalExtension = file.getName().substring(originalBaseName.length());
+        String hash = HashFunction.SHA1.hash(file);
```
LexManos: Why is the hash here? And this derps up the whole classifier vs version thing. Where you'd end up with things like "foo/1.0/foo-1.0-sources_mapped_blah.jar" instead of what the remapping repo is expecting: "foo/1.0_mapped_blah/foo-1.0_mapped_blah-sources.jar"

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
```diff
@@ -0,0 +1,150 @@
...
+            Utils.updateHash(output, HashFunction.SHA1);
+            cache.save();
+        }
+
+        return output;
+    }
+
+    private File getOrCreateCacheDir(String... cachePath) {
+        File cacheDir = new File(cacheRoot, String.join(File.separator, cachePath));
+        ;
```
LexManos: ...

LexManos:
Your opinions on wither we should use Gradle's internal downloader or not are irrelevant. I spent 3 weeks fighting with the damn gradle downloaded because it has horrible, undocumented, un-bypassible behaviors that fundamentally break core aspects of what FG does. So please stop parroting it.

And don't be spouting 'we' like you wrote/developed any of this. If you want to claim credit you should of been there at the beginning and helped out. FG is NOT going to be re-written YET AGAIN because an outsider thinks it should. Honestly this entire PR seems to be pushing twards more reliance on internal changing gradle API then I'd like. The SIMPLEST solution would be to just search for deobfX config names, and have them attach to X config. with 'deobf' defaulting to 'compile'. 

The point is, I have to maintain this, and I spent several weeks digging into gradle's internals trying to get the current system to work. I'd rather not introduce two majorly different codepaths to have to maintain. One of which relies on even more internal gradle API. Yes the current system is relying on internal API, but that's on the AF end. The FG end mostly uses public API. So if this were to be pulled it would need to nuke the old version and if feasible wrap the old version in a single block of well documented/deprecated code to delete in FG3.1

Overall i'm not seeing the benefits to this PR outweighing the costs.

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
Cypher121: This is used for the cache, not for the artifacts downloaded through maven and since this class is the only one accessing them, it works fine. The version Repo gets is used to find the artifact and mappings version, so it's not affected by this in any way.

That said, this way, the cache doesn't follow the maven structure that way, but I didn't think much of it, since local files won't be following that in any case, and the cache is just internals. If that's an issue I could restore the way maven artifacts were stored before (so no hash for them) and have hashes only for local files. Should I?

LexManos: Use the maven format I have speced for the maven formats.
use a {ObjectHash}/filename_mapped_blah format for flat files. 

Tho HONESTLY I think we should nuke any concept of using a file based deobfed dep. As all mods SHOULD be releasing as uniquely identified maven artifacts. And thus we should encourage people to use that in dev as well. What setups do you think would be good to have a flat file system? Why is that better then a maven structure?

Cypher121:
> And don't be spouting 'we' like you wrote/developed any of this. If you want to claim credit 

Didn't intend to.

I also tried avoiding the internal APIs as much as possible. The only place where any internal APIs are used is when making a local file dependency in DependencyRemapper and I wrote a note on why that had to be done as a comment. Rest follows the way gradle would usually be told to download things: by resolving configurations.

> So if this were to be pulled it would need to nuke the old version and if feasible wrap the old version in a single block of well documented/deprecated code to delete in FG3.1

This is basically done already. The only things remaining from the old version are the deobfConfiguration in the Plugin and findPom in Repo, the latter of which I didn't touch cause it's not implemented. Will try to condense the configuration related code and update findPom as much as possible.

LexManos:
>This is basically done already. The only things remaining from the old version are the deobfConfiguration in the Plugin and findPom in Repo, the latter of which I didn't touch cause it's not implemented. Will try to condense the configuration related code and update findPom as much as possible.

That isn't a single block of code, that is the entire repo class and the wrapper code in the apply section. What you should do is delete the repo code and replace that with whatever "compile fg.deobf('maven:artifact:version')" would be int he new system. 

>That means either calling Dependency#copy and then changing parameters in the copy or using gradle classes starting with Default, e.g. DefaultSelfResolvingDependency

And copying/changing using non-internal API is not a option? Yes welcome to my hell dealing with gradel's shitty API.

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
Cypher121: > Use the maven format I have speced for the maven formats
Sure.

As for file deobf deps, it's basically an option for when no better ones are available. 

All mods should be releasing as maven artifacts, but unfortunately that's probably never gonna happen. So if a user wants to depend on someone else's mod that's released as a file (say on curseforge), but not as a maven artifact, they'd have to either try to get it through CF's maven, which is shaky at best (spaces in filenames just break it), or, if it's open-source, clone it, set up publishing and publish to local. And who knows what kind of shitty un-reproducible build that other mod has. In cases like that, it's way simpler to just download and use it as a file dependency.

LexManos: All files can be referenced as a maven artifact. It just needs to be in the correct folder structure.
As for curseforge, there already exists projects to turn CurseForge into a maven repo.

Cypher121:
> And copying/changing using non-internal API is not a option? Yes welcome to my hell dealing with gradel's shitty API.

Yep, not the best API I've seen. Could be worse though. I'll try opening an issue with them on letting plugins provide some kind of dependency transformer, but who knows when/if they'll agree to impl it.

And most of the old ModRemapingRepo code is gone, the only things that are left (apart from findPom) are now in Deobfuscator and only deal with downloading mcp mappings. Nothing to change there, aside from maybe letting gradle download the mappings as well? Not sure if there's even a point in that, as it's a fixed file that's always on a known url.

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
Cypher121: So we don't handle local files and tell people that if they want to use one, they can add it with a `flatDir` repo? Not as convenient for users, but it does remove the only part of the PR that depends on gradle internal API.

LexManos: Yes. At the end of the day this *shouldn't* be a major use case. As modders *should* be publishing to public mavens. Hell Forge's maven is fairly public. We just need to add a signup system. And i'd rather get rid of internal API dependencies as much as possible.

Cypher121: Fair. Removing file handling then.

Cypher121:
This is about as done as I can get it. All deprecated code in one place, only depends on public gradle api.

src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
```diff
@@ -0,0 +1,147 @@
...
+        version = version.substring(0, version.length() - (mappings.length() + "_mapped_".length()));
+        String classifier = artifact.getClassifier() == null ? "" : artifact.getClassifier();
+        String ext = artifact.getExtension();
+
+        Artifact unmappedArtifact = Artifact.from(artifact).withVersion(version);
+
+        debug("  " + REPO_NAME + " Request: " + clean(artifact) + " Mapping: " + mappings);
+
+        if ("pom".equals(ext)) {
+            //return findPom(orig, mappings); //TODO: Unsupported for now, transitive deobfed deps? Need full xml reader?
```
LexManos: That needs to be filled out.

LexManos:
Looks better nothing jumps out as using bad gradle API.. well.. as best we can in this hacky situation.
So, what is the way we'd need to do it to get rid of the fg. prefix? Just so that's documented for 3.1.

Can you look into the pom? Should be just pulling in the pom, and replacing the version string?

But besides that, looks fine.

Cypher121:
For removing the fg object, there doesn't seem to be any kind of The Right Way to do it, since [Gradle devs prefer people to keep those inside extension objects](https://discuss.gradle.org/t/exposing-methods-from-plugin/12275/2), but the below code works (moved from DependencyManagementExtension).

    project.getExtensions().add("deobf", new Closure<Dependency>(project) {
        public Dependency doCall(Object dependency) {
            Dependency baseDependency = project.getDependencies().create(dependency);
            project.getConfigurations().getByName(UserDevPlugin.OBF).getDependencies().add(baseDependency);
            return remapper.remap(baseDependency);
        }
    });

POM handling done, just replacing the version. The only other option seems to be to change all dependencies' versions so they all get deobf-ed and I'm not sure if that'd break anything (probably would).

Cypher121:
Waiting for review

mezz:
I tested this dependency configuration locally using the latest Forge versions:
```
    // compile against the JEI API but do not include it at runtime
    compileOnly fg.deobf("mezz.jei:jei-1.13.2:5.0.0.15:api")
    // at runtime, use the full JEI jar
    runtimeOnly fg.deobf("mezz.jei:jei-1.13.2:5.0.0.15")
```
It is mostly working as expected. I think it is at the point where it's very usable 🎉 

There is an issue with the API sources however.
FG creates a copy of the mapped API jar and calls it
`jei-1.13.2-5.0.0.15_mapped_snapshot_20180921-1.13.jar`
and uses it as the sources for 
`jei-1.13.2-5.0.0.15_mapped_snapshot_20180921-1.13-api.jar`
which isn't right.

I think the mapped api jar should be using a mapped version of the `-sources` jar from maven. I'm not entirely clear on what is responsible here, since sources/javadocs are requested by the IDE afaik. I supposed I can just include all that in the api jar together.

Cypher121:
The issue here seems to be that, since IDEs just request sources directly from gradle using the mapped version, that sources artifact never makes it into the configuration where all obfuscated artifacts are downloaded, as `deobf` is never called. Not sure what the best approach here would be, aside from having users explicitly specify which deps to download sources for. Will experiment with it a little more to see if there's a good solution for that.

mezz:
There is an existing issue open for that problem, so I do not think it needs to be addressed in this PR.
https://github.com/MinecraftForge/ForgeGradle/issues/369

I tried to see if I could work around the issue by explicitly asking for the sources and javadocs like this:
```
    compileOnly fg.deobf("mezz.jei:jei-1.13.2:5.0.0.15:api")
    compileOnly fg.deobf("mezz.jei:jei-1.13.2:5.0.0.15:sources")
    compileOnly fg.deobf("mezz.jei:jei-1.13.2:5.0.0.15:javadoc")
```
but it is hanging forever at `Gradle: Resolve dependencies of:__obfuscated`

I am not sure if that is an issue with this PR or not.
In any case I think this is good to merge as-is, any remaining issues I can find are pre-existing or very minor.

Cypher121:
Not sure what's with the hangup, can't reproduce myself.

However, there is something else: there is apparently a public API for artifact transforms in Gradle, which has been there since 3.4, but never was documented, over at https://docs.gradle.org/4.9/javadoc/org/gradle/api/artifacts/transform/ArtifactTransform.html

It's working, but I'm not entirely sure if the way it's used fits this case. I'm currently discussing this with gradle devs over on their community slack, so I'd like to wait with this PR until there's a definite answer. Might be there's a way to get rid of the version-changing altogether, which would solve a few issues (source downloads, POM generation).

Cypher121:
Ok, seems like it won't work properly in 4.9 due to a bug, so that's out until FG updates to newest gradle. There's a documentation PR here, for future reference, https://github.com/gradle/gradle/pull/8957 Would recommend updating once the thing works right, it's pretty much perfect for deobf.

For now though, this should work well enough.

mezz:
For reference, what is the bug, and which version of Gradle will properly support what we want to do with artifact transforms?

Cypher121:
1. https://github.com/gradle/gradle/issues/9050 . Makes it impossible to add an attribute to artifacts, except adding it to all artifacts of the same type (e.g. all jar files), so can't mark only some deps for deobfuscation

2. probably 5.3.x? whenever the issue above gets fixed

3. the idea was to add an Attribute to dependencies marked for deobfuscation and have a TransformAction which downloads the mappings and deobfuscates the given files. This would let gradle handle every step except deobfuscation itself, while keeping the dependency at same coordinates, so there are no issues with the (mapped) version being used not actually existing.

There's a good sample [here](https://github.com/gradle/artifact-transform-samples/tree/master/cacheable-transforms). That one makes a new JAR file with some package names changed, while FG would make a new JAR file with deobfuscated contents.

mezz:
Thanks for the info.
I have been looking through the sample code there and it seems very promising.

PR #586, opened 2019-04-28
Fixing Task iteration in an InitScript.

marchermans:
Hello,

This PR fixes a problem that occurs when an InitScript loops over the tasks and causes them to be realised before ForgeGradles afterEvaluate has been called.

This particular problem happens when a ProjectEvaluationListener is registered by an InitScript (like by the buildsystem TeamCity, which does this to listen to Test results) whos afterEvaluate method loops over all tasks in a project causing the extractSrg task to be realised, before FG3s afterEvaluate is triggered (since it is registered after the ProjectEvaluationListener) , which fails with an NPE, as can be seen here: https://gist.github.com/OrionDevelopment/d468fc30bfaea6f94736017f742367ce .

This problem can be solved in two ways (based on StackOverflow searches and Gradle Forum, visits and questions):
The first one is by using a callback, and lazily set the value. This is the solution that was used in this PR. It is not the most elegant solution IMO, but since it is only required in one task, it was what I opted for here.

The other solution would be to use a set of additional tasks to configure the tasks that are actually executed, by registering them properly and setting their AST configuration the execution order guarantees that our tasks would be configured properly. This would be much more work then was needed to solve my problem, so it was not attempted here. But I wanted to leave this bit of knowledge behind in case somebody else needs it later.

For people interested the actual NPE is caused by the DownloadMavenAritfact task, because its _artifact field is null when the extractSrg task is realised, due to our afterEvaluate block not having triggered yet, when the realisation of the extractSrg task asks the downloadMcpConfig task for its output file to link them together properly.

Hoping for a quick merge,

Orion

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ def gitVersion() {
     def raw = grgit.describe(longDescr: true, tags: true)
     def desc = (raw == null ? 'unknown-unknown-unknown' : raw).split('-') as List
     def hash = desc.remove(desc.size() - 1)
-    def offset = desc.remove(desc.size() - 1) + "00"
+    def offset = desc.remove(desc.size() - 1)
     def tag = desc.join('-')
     def branch = grgit.branch.current().name    
     return "${tag}.${offset}" //${t -> if (branch != 'master') t << '-' + branch}"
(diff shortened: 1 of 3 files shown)


mezz (changes requested):
Looks good to me.

I'm assuming `getConfig` is not called often enough to warrant any optimization.
I think it may simplify the code if you removed `config` and replaced it entirely with `configSupplier`. 
Then the default `configSupplier` can just supply `null` and most of the logic goes away.

PR #587, opened 2019-04-29
Make EclipseHacks code nicer

Pokechu22:
Gradle provides [an API](https://docs.gradle.org/current/dsl/org.gradle.plugins.ide.eclipse.model.EclipseModel.html) for working with the eclipse code; there is no need to manually mess with the generated XML.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/EclipseHacks.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/EclipseHacks.java
@@ -20,89 +20,58 @@
 
 package net.minecraftforge.gradle.common.util;
 
-import groovy.util.Node;
-import groovy.util.XmlParser;
-import groovy.xml.XmlUtil;
 import net.minecraftforge.gradle.common.task.DownloadAssets;
 import net.minecraftforge.gradle.common.task.ExtractNatives;
-import org.apache.commons.io.IOUtils;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.tasks.TaskProvider;
+import org.gradle.plugins.ide.api.XmlFileContentMerger;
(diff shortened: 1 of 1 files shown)


src/common/java/net/minecraftforge/gradle/common/util/EclipseHacks.java
```diff
@@ -20,89 +20,58 @@
...
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
+import java.util.Iterator;
 import java.util.Set;
 
 public class EclipseHacks {
 
-    @SuppressWarnings({"unchecked", "UnstableApiUsage"})
```
Pokechu22, line 51: Since eclipse does not support `UnstableApiUsage` I'm not sure if it's actually logical to remove this.  Everything is normal API, but I might have missed something.

src/common/java/net/minecraftforge/gradle/common/util/EclipseHacks.java
```diff
@@ -20,89 +20,58 @@
...
+        classpathMerger.whenMerged(obj -> {
+            Classpath classpath = (Classpath)obj;
+            Set<String> paths = new HashSet<>();
+            Iterator<ClasspathEntry> itr = classpath.getEntries().iterator();
+            while (itr.hasNext()) {
+                ClasspathEntry entry = itr.next();
+                if (entry instanceof SourceFolder) {
+                    SourceFolder sf = (SourceFolder)entry; 
+                    if (!paths.add(sf.getPath())) {
+                        //Eclipse likes to duplicate things... No idea why, let's kill them off
```
Pokechu22, line 65: This probably has to deal with making changes in the IDE, and then because one version does not match exactly, it recreates the original version in addition.  But I'm not 100% sure of that.  Removing them is probably still fine.

PR #589, opened 2019-05-07
Add support for VSCode run configs

robin4002:
This PR add `genVSCodeRuns` to create runs file for VS Code. This makes modding on VS Code easier and more convenient.

As Visual Studio Code already import the gradle project himself when you add the project folder to the workspace, typing the command is the only step to take.

VScoderuns

Since the current code was only design to generate a run file in the XML format, I refactored the code to make it a little more flexible.

Note: The Java extension for VSCode use Eclipse JDT Language Server, the behavior is therefore similar to Eclipse. Binary file are compiled to the folder `bin/` so I also need to remap `MOD_CLASSES`.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/IDEUtils.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/IDEUtils.java
@@ -22,6 +22,11 @@ package net.minecraftforge.gradle.common.util;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+
 import org.apache.commons.lang3.tuple.ImmutableTriple;
 import org.apache.commons.lang3.tuple.Triple;
 import org.gradle.api.Project;
@@ -46,6 +51,9 @@ import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
 import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.Writer;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


mezz:
This looks good but will require some testing. Can you provide assurance that it does not break the run configurations for other IDEs?
Also, please update this to resolve merge conflicts.

robin4002:
Rebased.

I generated run files for Eclipse and Idea with the current version of ForgeGradle, renamed it, then I compiled my modified version of ForgeGradle, replaced the file in `.gradle/caches/modules-2/files-2.1/net.minecraftforge.gradle/ForgeGradle/3.0.127` by my custom version and generated run files again to compare it.
And I can confirm that the files are identical :
```bash
robin@robin-X550IU:~/Documents/forge-1.13.2-25.0.215-mdk$ diff runClient.launch runClient-before.launch 
robin@robin-X550IU:~/Documents/forge-1.13.2-25.0.215-mdk$ diff runServer.launch run
run/                     runClient-before.launch  runClient.launch         runServer-before.launch  runServer.launch         
robin@robin-X550IU:~/Documents/forge-1.13.2-25.0.215-mdk$ diff runServer.launch runServer-before.launch 
robin@robin-X550IU:~/Documents/forge-1.13.2-25.0.215-mdk$ diff .idea/runConfigurations/runClient.xml .idea/runConfigurations-before/runClient.xml
robin@robin-X550IU:~/Documents/forge-1.13.2-25.0.215-mdk$ diff .idea/runConfigurations/runServer.xml .idea/runConfigurations-before/runServer.xml
```

I also tried to launch from eclipse, as expected it's working. I haven't tested idea because I don't have it, but it should also work since the generated files are exactly the same. But if you still want a real test, I can ask a friend to test.

mezz:
It doesn't get much better than "identical", I don't think you have much to worry about there.
Thanks for testing it out thoroughly.

mezz (approved):

Unnoen:
The gradle tasks `:prepareRunClient` and `:prepareRunServer` (respectively) need to be run before the launch configuration as they are responsible for compiling the mod.

A simple method is to generate a `tasks.json` containing tasks that will call the needed commands and in `launch.json` using `"preLaunchTask": "taskNameWeGenerated"`


image
image

Unnoen:
Also, `${workspaceFolder}` should be used where possible.
 `"nativesDirectory": "${workspaceFolder}\\build\\natives"` for example.

robin4002:
No, it's useless and ineffective to run this task before every run.
Like a said before, VSCode use Eclipse JDT Language Server for the Java support, so the backend is the same and all behavior are identical.
When you import a Java project, all java source file are automatically compiled into the `bin/main` folder. When you save a modified source file, it is compiled automatically. (In VSCode you can also force compilation to be redone with maj + alt + b.)

In your setup you have to do it because your MOD_CLASSES is configured to use binaries / resources files that's are in `build/classes` and `build/resources`.
But if you change this to `bin/main` you won't need to run prepareRunClient/Server.


Otherwise concerning `${workspaceFolder}`, it will indeed be better as this will allow to move the workspace without breaking the launch configurations. I will improve that when I have time.

Unnoen:
Ah you are correct, my bad.

Nightenom:
Does this update or overwrite the file? Since now I have some local changes like permanent username or server nogui

robin4002:
It overwrites the file, like other IDE's runs generation.
If you want custom args, you should put it inside `build.gradle`.

Pokechu22:
If I recall correctly, the eclipse one and perhaps also the idea one do attempt to merge the existing content, though sometimes they aren't super successful at that (the whole `EclipseHacks` file is there to mitigate an issue where it duplicates some things instead of merging them).

robin4002:
Are you sure ? I try to test this, but if I run again genEclipseRuns after modifying the value inside `<stringAttribute key="org.eclipse.jdt.launching.PROGRAM_ARGUMENTS" value=""/>`, it return back to the default value. 
I did my test with the last version of FG (`3.0.130`), on `forge-1.13.2-25.0.191-mdk` and `forge-1.14.3-27.0.15-mdk` (the two mdk I have inside my test workspace).

Anyway I can still try to implement a merge for vscode gen

Crimix:
Any news on when this is merged and coming? 😄

robin4002:
Rebased.
Also I did a little modification, now the directory of the workspace is replaced by the variable `${workspaceFolder}` as suggested @Unnoen.

PR #617, opened 2019-08-27
Add class javadoc injection

JoJoDeveloping:
This allows classes to also have Javadoc documentation, provided the bot provides the appropriate data in the exports.


williewillus:
:itshappening:

LexManos:
Do you have a test case that shows it actually functioning and adding all necessary class cases?
Root Classes/Enums/Interfaces/Annotations
Nested Classes/Enums/Interfaces/Annotations
What about named nested method classes/enums?

JoJoDeveloping:
[Here](https://github.com/MinecraftForge/ForgeGradle/files/3726366/classes.csv.zip) is a test classes.csv (zipped because github is stupid) that can be added to the mcp bot exports to add a few test javadocs.
Minecraft does not actually define new @interfaces/annotations so they can't be tested, but it's not a problem if they wouldn't work

PR #618, opened 2019-08-27
Add support for extraParams.txt format

JoJoDeveloping:
Requires a change to the mcp config data exports from now on.
See https://github.com/ModCoderPack/MCInjector/pull/9


sciwhiz12:
This seems to have been force-pushed to an unrelated commit.

JoJoDeveloping:
Closed because it's no longer relevant.

PR #619, opened 2019-08-29
Update to Srg2Source V5

JDLogic:
Update Srg2Source to take advantage of the rewrite to RangeExtractor which signicantly reduces the runtime of the extract rangemap tasks. The measurements below are run time of the `extractForgeRangemap` task using the latest Forge 1.12.x branch.

```
Average time of 5 runs:
legacy (before) - 288289ms
batched (after) -   8416ms
```

Also note that the task to patch the JDT classes were taken from the Srg2Source repo, and I take no credit for it.


LexManos:
This shouldn't break existing builds as its just a internal dep that we shade. It might break some that are using internal classes... but... why would they?

PR #622, opened 2019-09-05
Replace HackyJavaCompile with direct JavaCompiler interface

### LINKED ISSUE #620: [FG3] Importing into IntelliJ broken
It seems like after an initial import, FG 3.0 completely breaks when trying to import to IntelliJ. The JavaCompile for resolving the primary artifact fails without end, even after wiping the `build` directory. It seems like Gradle 5.6.1 doesn't like how FG is compiling things.

https://gist.github.com/kenzierocks/c38032da8a7657ecf1aade05e39b6c87

This is not a daemon issue as with #563, as I disabled the daemon as best I could (IntelliJ still starts one because the tooling API relies on it).

octylFractal:
This PR removes the `HackyJavaCompile` task in favor of directly invoking the built-in `JavaCompiler` interface. This should be much safer & compatible than trying to hook into Gradle's task system at inappropriate times, and allows FG to work on Gradle 5.6.1 again.

Edit: This also fixes #620


src/userdev/java/net/minecraftforge/gradle/userdev/util/NonTaskJavaCompiler.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private JavaCompiler getJavaCompilerSafe() {
+        try {
+            return new JavaHomeBasedJavaCompilerFactory().create();
+        } catch (Exception e) {
+            // Perhaps Gradle internals were refactored. Use a simpler but maybe incorrect method.
+            logger.info("Unable to get Java compiler via Gradle, using ToolProvider", e);
+            return ToolProvider.getSystemJavaCompiler();
+        }
```
LexManos: The problem with this is javax.tools isn't a standard package that is included in every JDK. We would need to figure out a way to make sure it exists.
As for JavaHomeBasedJavaCompilerFactory, that is internal gradle API, and thus we are trying our best to stay away from it.

octylFractal: I'm pretty sure that Gradle itself relies on the existence of `javax.tools`, so it would probably not be good for someone to try and build without it present anyways. I've refactored to add a command fallback anyways, since that's the best we can go for if there is no `javax.tools`.

I've also re-implemented Gradle's internal JAVA_HOME compiler resolution, so we can avoid that API as well. Although, I'm not sure if including all this extra resolution code is worth not touching the rarely-modified `Jvm` class instead.

Niko-sk2x:
Wouldn't that also fix the issue with gradle daemon? As I remember, the issue with it was trying to run tasks when resolving dependencies.

octylFractal:
No, because `JarExec` is another root cause of that one from the stack traces I can find. It might reduce the chances of it happening, though.

LexManos:
What would it take to move the JarExec tasks to not use the gradle task system?
Also, I am weary of making major changes like this, if people can test this and respond back if it works or not. Testing is simple, copy, publishToMavenLocal, delete cached files from .gradle/cache/forge_gradle/userdev_repo/ and run setup.

octylFractal:
`JarExec` would probably be pretty simple in combination with this change -- we can just re-use the existing JAVA_HOME detection and call Java ourselves.

Personally, I've tested this change on Ubuntu 19.04, with both the default built-in system, and a patch to force the command-line `javac` version. It would probably be good if we can get some confirmation from Windows & Mac users, esp. Windows because it's got some special cases.

octylFractal:
https://github.com/kenzierocks/ForgeGradle/compare/bugfix/gradle-5.6-compat...kenzierocks:bugfix/gradle-daemon-compat for an quick write-up as to what it might look like to fix `JarExec` non-task cases. I can make it part of this PR if that's desirable.

Niko-sk2x:
This PR appears to work for me on windows 10 with adoptopenjdk installed (I haven't tested forcing different possibilities).
The linked gradle daemon compat breaks with this (it's not because of mingw, same happens in normal cmd, JDK bin directory is definitely in %PATH%): https://gist.githubusercontent.com/Barteks2x/7ff3e6a5ae3734b4661d0c995f239411/raw/3cea0e62fcb4209e8c614e34f1502da535eef06d/gistfile1.txt

octylFractal:
@Barteks2x pushed a potential fix for that, if you wouldn't mind trying again.

Niko-sk2x:
Still the same, even with JAVA_HOME set. I think the reason may be here https://github.com/kenzierocks/ForgeGradle/blob/290c2d20f9100c0ed4d1e0f8b23f3b160fcb298a/src/common/java/net/minecraftforge/gradle/common/util/Utils.java#L538 as `os.name` is `Windows 10` which does not contain the string `windows`.

Update: after some testing, it appears to work on windows after adding `.toLowerCase(Locale.ROOT)` in the linked place.

Niko-sk2x:
Also it doesn't in any way consider intellij setting for gradle JVM, if I set it to some custom JVM location, it will still try to find JDK in path and java_home. Would it be possible to get the java executable the current process is using? I think this would be the most reliable option if possible.

octylFractal:
The way to get it via the current process is `java. home`, that's already how Gradle does this. I'd be surprised if that's not working, that should be how it is. Nice catch on the windows detection, I'll fix it soon.

Edit: what's probably really wrong is I should be using JAVA_HOME environment variable before `java.home`.

Niko-sk2x:
This is why it doesn't recognize java installation configured in intellij: https://github.com/kenzierocks/ForgeGradle/blob/290c2d20f9100c0ed4d1e0f8b23f3b160fcb298a/src/common/java/net/minecraftforge/gradle/common/util/Jvm.java#L61
It doesn't use `.exe` on windows.

octylFractal:
Both branches should now have fixes for that.

Niko-sk2x:
Appears to work so fa with the system JVM. Both this PR and the gradle daemon compat. Gradle daemon also doesn't appear to cause any issues with it (daemon still definitely breaks things with just this PR).

But it STILL doesn't properly detect JVM from java.home (both this pr and gradle daemon compat branch), because the result of this subpath call https://github.com/MinecraftForge/ForgeGradle/blob/2ecd06024ba9f6e279ab211992d2d058d6a47dd8/src/userdev/java/net/minecraftforge/gradle/userdev/util/Jvm.java#L81 turns
 ` C:\Program Files\AdoptOpenJDK\jdk-8.0.222.10-hotspot_testcopy\lib\tools.jar` 
into
 `Program Files\AdoptOpenJDK\jdk-8.0.222.10-hotspot_testcopy`

octylFractal:
That method really doesn't behave like I thought it would. Strange that it won't copy the root element. Fixed in both branches.

Niko-sk2x:
It seems to work for me by default, both branches. The daemon compat executes java from java.home just fine.

But with this PR, if I force searching for java compiler in java home, I get this https://gist.githubusercontent.com/Barteks2x/467dc9c4ccff159b196dc94f44cef9f0/raw/be2bd0af4c7fe70e2b2ba7728c22e5ce1b8b7d99/gistfile1.txt
I have no idea why or how to fix it, maybe too long comandline arguments? (turns out I didn't really test that code path before with this PR, I just *thought* I did. I only really tested it with gradle daemon compat branch)

octylFractal:
That appears to just be a Windows limitation (https://stackoverflow.com/a/10598074/436524), the solution I implemented is to put the classpath & sources into a `javac` argfile, so that limitation can't be hit again.

Niko-sk2x:
The result is that the compiler fails because the arguments are wrong (it sees the classpath as invalid argument).
[javac log](https://gist.github.com/Barteks2x/2220fc0828a1aeebc8658facb7c726ed)
[generated argfile](https://gist.github.com/Barteks2x/14f3fd3058c319971c0f831d8ec07c89)

Update: it seems to be fixed by changing https://github.com/MinecraftForge/ForgeGradle/blob/84072a5150862bb76cb9c348c6d078abcac0ea2e/src/userdev/java/net/minecraftforge/gradle/userdev/util/CommandCompatJavaCompiler.java#L85-L96 to just output the lines directly, without any processing (is that even necessary with argfile, as it's no longer restricted to commandline limitations?)

octylFractal:
Yes, I accidentally left a `skip(1)` in there. Sorry for all the back-and-forth, I haven't had time to test any of the changes myself so I'm just pushing what I have, to re-test myself later.

marchermans:
@octylFractal Are you still available to work on this? If so could you update/rebase this?

@Barteks2x How is your state on this? Is this working?

Niko-sk2x:
I am unable to reproduce the bug that this fixes anymore even without this PR so I lost interest in it. Last time I tested it, it worked.

But I also remember asking lex about this PR on discord, with the end result being a no, because it was impossible to guarantee it would work in every case.

octylFractal:
Yes, gabizou created an alternative fix that allowed it to work again on some Gradle versions, but I would expect it to break again without warning. Once Gradle 6.7 is out I'd be interested in updating this PR to be an alternative method for 6.7+, as the issue of finding the compiler is now no longer there! In 6.7, Gradle exposed their tool chain lookup utility, so we are free to grab the compiler in the same way they do.

marchermans:
Then I will close this PR.

Issue #624, opened 2019-09-10
Find an alternative to 'registering' a custom AbstractArtifactRepository


Aaron1011:
Currently, ForgeGradle (through Artifactural) ['registers' a custom AbstractArtifactRepository with Gradle](https://github.com/MinecraftForge/Artifactural/blob/master/src/gradlecomp/java/com/amadornes/artifactural/gradle/GradleRepositoryAdapter.java#L87-L108). This a huge hack - Gradle does not support registering custom artifact repositories, which requires us to use a combination of reflection and internal Gradle apis.

This has the effect of tying all users of ForgeGradle to a specific minor (or even point) release of Gradle. Upgrading Gradle becomes very difficult, as there's no guarantee that ForgeGradle will work at all.

This problem is only going to get worse over time. Since Artifactural is using internal Gradle APIs, Gradle is free to change those APIs at any time. We've already dealt with this once - a change to a constructor signature required [a horrible build-time bytecode hack](https://github.com/MinecraftForge/Artifactural/blob/d3c01e7be2247cce9f6fe3456db73dd0a7ef1ccf/build.gradle#L91) in order to a produce a ForgeGradle jar compatible with multiple versions of Gradle. As new versions of Gradle are released, ForgeGradle will need to choose between dropping support for old minor (or even patch!) releases, or adding even more version-specific hacks to hold things together.

We should investigate officially supported alternatives to the current custom artifact repository hack. If acceptable solutions exist, I will try to write up a Gradle feature request describing the changes that would be necessary.

LexManos:
I would LOVE to move away from ugly hacks, but at the time this was written Gradle had nothing.
Gradle 5.5 introduced [Artifact Transformers](https://docs.gradle.org/5.5/userguide/dependency_management_attribute_based_matching.html#sec:abm_artifact_transforms), but 4.10+ have caching issues that I don't want to deal with {It deletes random files from the cache in the name of 'automatic cleaning'}. As well as the transformers not being fully functional/fleshed out last time I looked at them.

As it sits, I'm jusst really burnt out on trying to get Gradle to do what SHOULD be a simple concept of 'let me make code that generates a dependency', but it comes with the baggage of the other crap in Gradle's dep management.

DenWav:
[...]

PR #625, opened 2019-09-10
Fix HackyCompile for Gradle 5.1+ to ignore compilation time reporting.

gabizou:
An annoyance with some of the Gradle changes involves [this change](https://github.com/gradle/gradle/commit/4457941d9896528a1adb3703e42128c9891b5beb#diff-e970d36a1f73d09c05bdd69a7a585f92L152) causing a `JavaCompile` task to try and register itself as being performance logged. The issue is that with how hacky the compilation task is performed, [something something NPE from missing listeners](https://gist.github.com/gabizou/58dcd667333282fdaa5ac33ca7ca385f#file-spongeforge-output-L134)  will happen with Gradle 5.1+.

Yes, it's been discussed about getting [a replacement](https://github.com/MinecraftForge/ForgeGradle/pull/622), and there's a [discussion that's cropped up from the issue above](https://github.com/MinecraftForge/ForgeGradle/issues/624#issuecomment-529747404), this PR simply puts a bandaid fix that works all the way up to Gradle 5.6.2 (latest release), tested on Windows 10, and macOS with Gradle versions 5.2.1, 5.1.1, 5.5, and 5.6.2. This will at least buy us some time to eventually proceed with a way forward for using [artifact transformers](https://docs.gradle.org/5.5/userguide/dependency_management_attribute_based_matching.html#sec:abm_artifact_transforms), which is likely going to need some more time to migrate to (and of course, requires that a lot of people again migrate up Gradle versions).

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/tasks/HackyJavaCompile.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/tasks/HackyJavaCompile.java
@@ -21,9 +21,18 @@
 package net.minecraftforge.gradle.userdev.tasks;
 
 import org.gradle.api.file.FileCollection;
+import org.gradle.api.internal.tasks.compile.CleaningJavaCompiler;
+import org.gradle.api.internal.tasks.compile.DefaultJavaCompileSpec;
+import org.gradle.api.internal.tasks.compile.JavaCompileSpec;
+import org.gradle.api.tasks.WorkResult;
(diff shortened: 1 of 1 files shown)


clienthax:
Tested on gradle 5.1.1, works.

LexManos:
This will work as a interm solution until https://github.com/MinecraftForge/ForgeGradle/pull/622 gets fully tested and the java location issues get resolved.

PR #628, opened 2019-09-23
Add updateMappings task to userdev

JDLogic:
Adds a userdev version of the updateMappings task as a convenience for modders. Also changes the dependency deobfuscator to use the fake repos for getting the mapping information.

In the case that this is used to update a mod to use Mojang's official mappings, I wasn't sure if we should output a warning about the use of those mappings. As in, they are for _internal reference_ purposes and sources referencing those names should not be published.


LexManos:
yes, you should add the warning about the license on the official mappings... I didn't get around to doing that.
Should probably add it as a every-run nag if the mappings are set to official for now as well.

JDLogic:
Ok, I added a warning about the license for the official mappings. I also added a message to point issues with source renaming to the Srg2Source repo as I could see people making issues here about already known bugs. This is what is looks like if a user runs the updateMappings task with the official mappings as the target.
UpdateMappings Warning

The every run warning when using the official mappings is basically the same and looks like this:
Every run warning

PR #630, opened 2019-10-08
Fix Obfuscation deadlock with IntelliJ

Asherslab:
With the help of @OrionDevelopment we were able to find and hopefully eliminate the issue of some IDEs (sometimes) deadlocking when using `fg.deobf` in gradle configs.

Description from Orion, because he's a lot better at describing this stuff:

Due to the fact that IDEA (and possibly also eclipse) resolve dependency information in parallel for each module, during the build finishing phase the daemon worker thread holds a lock on the notification system. This lock prevents the worker thread (which is spawned by the daemon worker thread to handle the resolving of dependency information) from notifying the IDEA handler of the impending resolution of dependencies (since the DeobfusctingRepo has not been called yet, its origin has not been resolved, during the same phase that the other configurations got resolved).

This PR locks down the access to the origin configuration as well as force the resolution of its dependencies early, so that this happens in afterEvaluate, which means that the lock on the notification system is not yet held by the DaemonWorker, since the build has not finished yet, allowing IDEA to resolve the dependencies properly in the build finished phase, preventing the deadlock form occurring.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -178,6 +178,9 @@ public class UserDevPlugin implements Plugin<Project> {
 
             if (!internalObfConfiguration.getDependencies().isEmpty()) {
                 deobfrepo = new DeobfuscatingRepo(project, internalObfConfiguration, deobfuscator);
+                if (deobfrepo.getResolvedOrigin() == null) {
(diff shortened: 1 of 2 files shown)


williewillus:
Seems to resolve my deadlock issues on my laptop

Asherslab:
I've been using it for about 4 days now and i've not have the issues ever since (after previously having it 100% of the time), also haven't had any other issues

Tfarcenim:
Same, solved my occasional deadlock issues when using JEI.

PR #631, opened 2019-10-20
Changing the repository and caching scheme to take individual projects, within a workspace, into account.

marchermans:
Continuing from the debugging me and @Asherslab did in #630, i continued trying to fix the usage of FG3.+ inside IDEA.

## Error description:
While doing this I came across the use case of child projects. Initially all seemed well, projects loaded and dependencies where actually loaded up properly. However as soon as at least one child-project differed from the others (say for example it used a custom access transformer) the current system breaks.

Currently FG generates the required artifacts and sources into its own cache, additionally only one instance of MCPRepo and MinecraftRepo exist (regardless of how many projects FG is applied to in a workspace) and both the cache entries and the Repos are created only for the first project to which FG is applied. So its configuration (basically anything that influences the created deobfuscated minecraft jar, like the access transformer) is used to generate the repos as well as all cache entries. If any project, that comes after, the initial one, has a different configuration, the initially created repo (for the first project, with its configuration) tries to lookup the files, notices it has them in the cache, and returns the already created file. Regardless of the fact that the configuration in the secondary project requires a different artifact to be generated.

## Error resolution:
This PR adds two changes:
 1) It changes the cache base path to include a projects name (and that of its parents)
 2) MinecraftRepo and MCPRepo are now individually created for each project and stored in a concurrent map, to allow for different configuration to generate artifacts properly

## Resolution consequences:
 1) First and foremost, every project now generates its own artifacts in its own cache. This has some significant side-effects, especially when it comes to performance in multi project workspaces. Each project now has to generate its own dependencies, and MC has to be decompiled and deobfuscated for each project individually. Since gradle does not do this in parallel, even if the parallel processing option is enabled, this might take a bit of time depending on the amount of projects in the workspace.
 2) Each project has its own cache (by design), if the cache for one of the projects breaks, it can be easily recovered.

## Why do I need this PR?
Additionally to get FG to work in a multi project per workspace scenario, I actually have a workspace like this. My project called 'BlockOut' actually has its components split apart into several child projects, as such (as of writing) it has 6 projects within one workspace, which have FG applied (it actually contains a lot more projects, but most of them do not need FG so they do not have it). The problem arose, when I could not apply different AT files, or even different configurations and get IDEA to recognize the new changes, so I dug into the Caches and saw to no additional files where generated. Some debugging later and I came up with this solution.

Greets,

Orion

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MinecraftRepo.java
@@ -26,15 +26,13 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
-import java.util.Enumeration;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import java.util.zip.ZipOutputStream;
 
+import com.google.common.collect.Maps;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.gradle.api.Project;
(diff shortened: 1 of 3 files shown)


ichttt:
Sorry, missclick on phone

marchermans:
@ichttt It is fine. Lex already reacted and he is completly opposed to this solution and does not see this as a Problem in FG3, even though FG3 does not follow Gradle Spec with regards to Mutli-Project workspaces here.

In understand that Splitting apart the Cache is not an Optimal solution here, I totally agree. But I mean, Right now it is not working at all. Cause two Projects with different ATs cause the same collision to appear constantly overriding each other.

PR #633, opened 2019-11-13
Fix license in FG_1.2

anatawa12:
ForgeGradle was re-licensed to LGPL in d57f4a2 but the pom file still says this FG is licensed under EPL. This fixes this problem

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -161,8 +161,8 @@ uploadArchives {
 
                         licenses {
                             license {
-                                name 'Eclipse Public License (EPL), Version 1.0'
-                                url 'http://www.eclipse.org/legal/epl-v10.html'
+			        name 'Lesser GNU Public License, Version 2.1'
+                                url 'https://www.gnu.org/licenses/lgpl-2.1.html'
                                 distribution 'repo'
                             }
                         }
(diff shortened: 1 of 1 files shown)


DaemonUmbra:
Shouldn't this branch be considered retired?

anatawa12:
there is some commit for FG_1.2 in this year and updated SNAPSHOT so I think this is not retired.

srs-bsns:
Those commits are irrelevant. They were only made because of changes made in FG 2.3 that affected all previous FG versions. (Removing version resolution through an external json file), and they had to be made so that 1.7.10 dev environments would continue working.

This change is rather pointless though.

LexManos:
The real question is.. who actually reads the damn pom file for license info?
Everyone is right tho, FG < 2.3 is effectively EOL except for fundamentally breaking bugs.
Honestly we *could* move all MCP/Forge data into FG3 data format. So that people can use the new system.
It just requires work that is not important and thus not even on my TODO list.
It doesn't really matter, we can pull this. it won't effect anything of importance.

anatawa12:
> who actually reads the damn pom file for license info? 

I often check build.gradle or pom to check license because the license sometimes modified. if the pom or build.gradle have link to oridinal license, I think the license is not modified. 

> Those commits are irrelevant

OKey. I understand.

PR #635, opened 2019-11-13
Fix classes javadoc breakage

JoJoDeveloping:
Fix bug in previous PR ( #617 )

We previously failed hard when we couldn't keep track of indentation. Now we just fail softly and warn the user that there is likely an issue with class javadoc comments.

We could also use "debug" or "info" instead of "warn" for the logger or just stop adding class javadocs when it fails, instead of trying to recover

### DIFF
--- a/.github/support.yml
+++ b/.github/support.yml
@@ -1,0 +1,16 @@
+# Configuration for support-requests - https://github.com/dessant/support-requests
+
+# Label used to mark issues as support requests
+supportLabel: Forum
+# Comment to post on issues marked as support requests. Add a link
+# to a support page, or set to `false` to disable
+supportComment: >
+  👋 We use the issue tracker exclusively for final bug reports and feature requests.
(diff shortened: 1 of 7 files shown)


d0sboots:
What will it take to get this merged in? The bug (that this looks like it fixes) prevents "gradlew eclipse" from working in 1.13,2, 1.14.2, and 1.14.3.

JoJoDeveloping:
It is complete from my side, and with it setting up the MDK should work again (at least on 1.14.2, which I tested).
Note that these versions are unsupported, though.

covers1624:
Can you please rebase this on top of the `FG_4` branch.

You also don't need to pass the Logger object around just do `private static final Logger LOGGER = Logging.getLogger(MyClass.class);`

JoJoDeveloping:
Closed as this is no longer up to date

PR #637, opened 2019-11-25
Fix fg.deobf dependencies not having attached (mapped) sources

tterrag1098:
Forces a :sources dependency into the obf configuration so that later IDE requests can see it. Also fixes a typo in the source deobfuscator.


LexManos:
As discussed on Discord, the better solution is to move back to MavenArtifactDownloader instead of letting Gradle's black magic deal with the internals.
This system may function, but it's brought to light issues with https://github.com/MinecraftForge/ForgeGradle/pull/577

PR #645, opened 2019-12-30
Properly cache deobfMcMCP to improve performance

newbthenewbd:
Currently, when a project is being built with the use of ForgeGradle-2.3-SNAPSHOT, it is checked whether the project has been setup with `setupDecompWorkspace`. Unless that happens to be the case, the task `deobfMcMCP` is delegated for execution, along with the tasks it depends on.

During the execution phase of the Gradle lifecycle, the mentioned tasks may be marked as `SKIPPED` if they have already been performed earlier and their results have been cached for the project. However, it seems that merely determining these facts at this point may take Gradle 10 seconds or more - scaling to quite a lot of time, and, presumably, processing power as well.

This pull request aims to fix the described issue by making ForgeGradle check for the existence of the output of the `deobfMcMCP` task, instead of the entire decompiled workspace, before the execution phase starts. I attach a small comparison of the practical performance of the two methods on the same machine, each starting with a clean build of a small project, followed by two cached ones:

* [Old](https://pastebin.com/raw/4zLEpcyU): Clean build **44s**, cached builds **13s** and **12s**, respectively, for a cached average of **12.5s**;
* [New](https://pastebin.com/raw/VKGFSQPe): Clean build **31s**, cached builds **3s** and **2s**, respectively, for a cached average of **2.5s**.

(The seemingly improved clean build times are probably incidental, perhaps dependent on my network speed or some Gradle daemon intrinsics, however the improvement that follows for the cached builds is fully reproducible on my system, for all projects that I have thus far tested.)

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/user/UserBasePlugin.java
+++ b/src/main/java/net/minecraftforge/gradle/user/UserBasePlugin.java
@@ -300,6 +300,13 @@ public abstract class UserBasePlugin<T extends UserBaseExtension> extends BasePl
     {
         madeDecompTasks = true; // to guard against stupid programmers
 
+        final Object deobfBinJar = chooseDeobfOutput(globalPattern, localPattern, "Bin", "");
+        final Object deobfDecompJar = chooseDeobfOutput(globalPattern, localPattern, "", "srgBin");
+        final Object decompJar = chooseDeobfOutput(globalPattern, localPattern, "", "decomp");
+        final Object postDecompJar = chooseDeobfOutput(globalPattern, localPattern, "", "decompFixed");
+        final Object remapped = chooseDeobfOutput(globalPattern, localPattern, "Src", "sources");
+        final Object recompiledJar = chooseDeobfOutput(globalPattern, localPattern, "Src", "");
+
         final DeobfuscateJar deobfBin = makeTask(TASK_DEOBF_BIN, DeobfuscateJar.class);
         {
             deobfBin.setSrg(delayedFile(SRG_NOTCH_TO_MCP));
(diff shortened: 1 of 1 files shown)


0xilly:
It's been explained multiple times to you in the discord why this isn't an option or ideal. Randomly hacking away tasks for your edge case isn't going to happen Gradle takes time to check if a task needs to be run we don't control that.

newbthenewbd:
I would be thankful for a quote of the `multiple times`. To the best of my present knowledge, the only time that something to this extent has been recently said on the Forge Project's Discord channel was affirmed by me... And was referring to manually excluding `deobfMcMCP` with each cached `gradle build`. This pull request is the result of Yours truly taking a look in ForgeGradle's code as to why the issue actually occurred. Indeed, I could have not received prior negative feedback about this particular change - simply because I have not publicly shared any information about it prior to sending this very request...

If You also took the time to take a single look at the proposed code, You might've noticed that the _current_ version is, by Your definition, already a hack, and quite a bit hackier than what I propose - instead of deciding whether `deobfMcMCP` should be executed depending on whether that very `deobfMcMCP` has been executed yet and its output is available, it decides depending on the output of the much heavier `setupDecompWorkspace`. I am sad to conclude that this must have had not happened.

Unless, of course, I am somehow horribly wrong and the change breaks some functionality that I had not tested or might otherwise deteriorate user experience? If that is the case, I am naturally open to hear about it - for the first time.

Should I however get no objective review of the pull request, yet have it declined still, I will have to conclude that Minecraft Forge is not a serious place to attempt to contribute improvements to, for free.

PR #646, opened 2020-01-06
Support third party mappings channels

tterrag1098:
Removes the pointless limitation on mapping providers. Will now simply warn for "unknown" channels, but still attempt a maven lookup for them. This allows third party sources to provide mcp artifacts for custom channels. Effectively implements the TODO, assuming whatever other project provides a maven artifact matching the MCP format.


LexManos:
Other providers should not be publishing under MCP's name.
We can discuss your intentions on discord so i can get your idea better.

PR #648, opened 2020-01-13
Rework run config generation.

gigaherz:
I have reworked run config generation logic, so that calculating MOD_CLASSES environment is now less of a mess.
~~This requires [a corresponding PR to forge](https://github.com/MinecraftForge/MinecraftForge/pull/6431), since the `source_roots` replacement token is no longer needed, and in fact breaks the run config generation process.~~ The PR still exists, but it's now a minor cleanup.


LexManos:
Not acceptable if it doesn't load old data.

gigaherz:
I have reworked it again to play nice with the tokens stuff, AND added a workaround for that pointless "dummy" value that should have never existed. This way it won't choke if it finds a version of forge with the "dummy" BS in it, and it should still work with any configs that make use of the sources_root replacement.

PR #649, opened 2020-01-20
introduce Travis

hannesa2:
Additional you have to add `ForgeGradle` to your [Travis](https://travis-ci.org)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+sudo: required
+dist: trusty
+
+language: java
+jdk: oraclejdk8
+
+#branches:
+#  only:
+#    - master
+
+before_install:
+  - java -version
+
+script:
+  - ./gradlew clean build
+
+git:
+  depth: false
+
+before_cache:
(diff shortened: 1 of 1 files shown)


hannesa2:
Ahh, I see you use Jenkins right now. But I can't see results.
I still prefer transparent Travis

Unnoen:
The choice for Jenkins was made during the 1.13 rewrite, there's some private information in the builds such as publishing to maven and webhooks which is why it's not public, it may be in the future.

Unrelated, but I noticed your issue in the MinecraftForge repo earlier, if you want more information and discussion [our Discord server](https://discord.gg/UvedJ9m) would be better suited for that.

hannesa2:
> it may be in the future.

Maybe the future is now ?

In Travis you can add some secret variables
image

which only the maintainer can see. This solves the problem about private information

Unnoen:
Like I said, possibly in the future, we are always looking at improving things, especially the build pipeline, however it's not as simple as that right now.

PR #650, opened 2020-01-20
upgrade to Gradle 5

hannesa2:
Even https://github.com/MinecraftForge/ForgeGradle/pull/550 has some sense or not, It should not prevent an upgrade to Gradle > 5.1.0


Unnoen:
We're sticking with 4.9 for the moment as some newer Gradle versions have features that can mess with certain tasks (like build caching). 

We know that 4.9 is stable for ForgeGradle (and it was written for it) and thus we will be sticking to that for the foreseeable future, at least until a ForgeGradle update/rewrite.

You are welcome to use different versions locally if you wish.

PR #657, opened 2020-02-20
Fix gradle run tasks getting garbage instead of the right args

gigaherz:
I missed a toArray().

Also, fix intellij delegation detection logic... again.


gigaherz:
As per tterrag's request. I fixed the issue where args with spaces are not properly delimited by our run generators.

pupnewfster:
Tested this and can confirm it works.

tterrag1098 (approved):

Issue #658, opened 2020-02-21
ForgeGradle 1.2 HTTPS repo GET failure - error 501 HTTPS required


JackCarterSmith:
When using Forge 1.7.10, I got errors "501 HTTPS required" when gradle initialising. I think need to update link in constant file.

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'test'.
> Could not resolve all dependencies for configuration ':classpath'.
   > Could not resolve net.md-5:SpecialSource:1.7.3.
     Required by:
         :test:unspecified > net.minecraftforge.gradle:ForgeGradle:1.2-SNAPSHOT
      > Could not GET 'http://repo1.maven.org/maven2/net/md-5/SpecialSource/1.7.3/SpecialSource-1.7.3.pom'. Received status code 501 from server: HTTPS Required
   > Could not resolve org.apache.httpcomponents:httpclient:4.3.3.
     Required by:
         :test:unspecified > net.minecraftforge.gradle:ForgeGradle:1.2-SNAPSHOT
      > Could not HEAD 'http://repo1.maven.org/maven2/org/apache/httpcomponents/httpclient/4.3.3/httpclient-4.3.3.pom'. Received status code 501 from server: HTTPS Required
   > Could not resolve org.apache.httpcomponents:httpmime:4.3.3.
     Required by:
         :test:unspecified > net.minecraftforge.gradle:ForgeGradle:1.2-SNAPSHOT
      > Could not HEAD 'http://repo1.maven.org/maven2/org/apache/httpcomponents/httpmime/4.3.3/httpmime-4.3.3.pom'. Received status code 501 from server: HTTPS Required
```

DaemonUmbra:
Please update to a modern version to receive support.

JackCarterSmith:
I've found a fix using maven.minecraftforge.net endpoint. Like the 1.12.2 version does, place it in the build.gradle file:

```
buildscript {
    repositories {
	maven { url = 'https://maven.minecraftforge.net/' }       <-- Added this line before mavenCentral()
        mavenCentral()
    }
    dependencies {
        classpath 'net.minecraftforge.gradle:ForgeGradle:1.2-SNAPSHOT'
    }
}
```

I'm not sure I'm doing it very well, but it's worked for me!
**It's can be usefull to put this patch in released src zip on forge website.**

PR #659, opened 2020-02-21
Migrated to HTTPS support for Maven repo

JackCarterSmith:
Fix errors "501 HTTPS  required" when using gradle

#658

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/common/Constants.java
+++ b/src/main/java/net/minecraftforge/gradle/common/Constants.java
@@ -50,13 +50,13 @@ public class Constants
     public static final Closure<Boolean> CALL_FALSE = new Closure<Boolean>(null){ public Boolean call(Object o){ return false; }};
 
     // urls
-    public static final String MC_JSON_URL      = "http://s3.amazonaws.com/Minecraft.Download/versions/{MC_VERSION}/{MC_VERSION}.json";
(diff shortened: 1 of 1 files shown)


DaemonUmbra:
FG1.2 is no longer being developed.
If you REALLY want to use 1.7.10, feel free to update the 1.7.10 ForgeDev workspace to use FG3.

LexManos:
Note: This actually  WOULDN'T fix the error. As you'll notice the server throwing the error ISNT one of the ones you've changed. But yes, if you want any sense of being accepted. The only option is to put in the work to modernize the old branch to fg3.

JackCarterSmith:
OK, no problem.
I'm still working on one of my mods in 1.7.10 when suddenly no longer possible to use gradle because of this error... You mean that it is possible to use FG-3 with MC 1.7.10?
If I'm not mistaken, https://files.minecraftforge.net/maven refers to http://repo1.maven.org/maven2/, hence the link with the issue #658

DaemonUmbra:
It's possible, but a few required extra steps are yet to be taken and nobody has stepped up to do it.

LexManos:
> If I'm not mistaken, https://files.minecraftforge.net/maven refers to http://repo1.maven.org/maven2/

You are mistaken, they are two completely different servers, that only show up in the stack trace because both are checked for the artifact.

And yes, it's possible to use literally any version of Minecraft {or any other java project} with FG3. Someone just needs to make the datasets needed. Nobdoy have bothered to because old versions are old and nobody cares.

PR #662, opened 2020-02-28
Added basic mirroring functionality to ForgeGradle

FledgeXu:
Hello, the maintainers of the ForgeGradle.
I add a new feature that allows the user to use the mirrors repositories instead of the original maven repositories which sometimes is slow in some countries and regions to the ForgeGradle.
Besides, allowing the user to use the mirrors repositories also can help the people who still work on the old Forge Version.
If there are any problems in the code, I'm glad to fix it.
People just need to add one line to the `build.gradle`'s `Minecraft` closure to enable this function :
*removed links to questionable sites*
Have a good day.


LexManos:
You can already add extra maven repos to your own build script. And only Mojang has legal rights to distribute their material so we will not be adding a way to bypass their sites.
Nothing in this pr cant be do e already. And pulling this pr would both be encouraging bad practices and add bloat to fg which we would have to maintain. So no

FledgeXu:
Ok, Thanks for your reply.

PR #674, opened 2020-05-08
Fix deprecation (Due to be removed in Gradle 6.0)

QartemisT:


LexManos:
You're gunna need to be a lot more descriptive then this.
As for Gradle 6 compatibility, there are quite a few other things we need to test before we even think about moving twards that officially.

I would LOVE to get rid of our Artifactual hacks, but as it sits, even in Gradle 6 there doesn't seem to be anything that provides the functionality we need.

loordgek:
https://docs.gradle.org/5.5/release-notes.html#transforming-dependency-artifacts-on-resolution
does this help ?

LexManos:
Not in the slightest, we've already looked into that system and it is not functional for our needs.

QartemisT:
Exact words from gradle's output:

```
The TaskContainer.remove(Object) method has been deprecated.
    This is scheduled to be removed in Gradle 6.0. Prefer disabling the task instead,
    see Task.setEnabled(boolean).
```

covers1624:
This has been resolved in FG 4.

PR #676, opened 2020-05-12
Gradle has removed http support.

tbringuier:
We need a gradle version higher than 3.0 to get working ForgeGradle for Minecraft 1.7.10-1614 or other old versions wich are constantly used today.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14-bin.zip
+distributionUrl=https://services.gradle.org/distributions/gradle-3.0-bin.zip


FinnT730:
Anything under 1.14.4 and 1.15.2 is unsupported.
Please update to receive support.

tbringuier:
But a lot of people are today using Minecraft 1.7.10.
It's only a rebuild and publishing the new version.
There's no one other things to patch.

FinnT730:
That is a. old excuse. No support means no support.
If the Forge team would build another version of 1.7.10, then everyone will come back and beg for support, like they always do.

Rongmario:
@MichelBaie you should be able to use all the way upto `4.4.1` last time I checked. You'd have to manually edit your `gradle-wrapper.properties`. That should fix some bugs and speed up setups and builds.

tbringuier:
But a lot of people doesn't know that and are creating threads on forums. MinecraftForge must do an annoucement about that. Yesterday I thought that version was definitely dead.

Rongmario:
Yeah it's unfortunate, MinecraftForge indeed isn't supporting anything below 1.14 now.

FinnT730:
I wish people ON the Forums would read: https://www.minecraftforge.net/forum/topic/79304-112114-and-lts-system/

Since it has been out for a while now.

tbringuier:
It's sad :(

DaemonUmbra:
Tagging with Forum to lock, if you do go to the forums the answer will not change

LexManos:
We don't have the bot on this repo so it won't auto lock so i'll do that.
But before I do.
To make it clear, no we have not done anything on our end to break older versions.
However we are also not doing anything on our end to FIX older versions when other systems break.
In this case gradle/maven central no longer accepting non-HTTPS requests.
No this is not a simple 'a rebuild and publishing' There are many things that need be done correctly.
A [MCPConfig](https://github.com/MinecraftForge/MCPConfig) version has to be created.
Forge needs need to be [Updated to FG3](https://github.com/MinecraftForge/MinecraftForge/compare/863ab2ca184cf2e2dfa134d07bfc20d6a9a6a4e8...1.12.x)
New dependencies have to be deployed on our maven.

I already did the [hard parts](https://github.com/MinecraftForge/ForgeGradle/commit/83993e9b64a59ac1f8ae01728ccc1ff2e40a2b31). It's now on all of the supposed mass of modders still developing for older version to show that they exist and are willing to be helpful. 

Anyways, thread locked, there are obvious better places to discuss this, do your research fist and he helpful and we may be helpful in return.

PR #682, opened 2020-05-28
Document required gradle version

Phlosioneer:
I couldn't find the supported gradle versions anywhere.
I'm unsure if the lower bound of 5.0 is correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,4 @@ ForgeGradle
 
 Minecraft mod development framework used by Forge and FML for the gradle build system
 
-Requires Gradle 5.0 - 5.6.4
+Requires Gradle 4.0 - 5.6.4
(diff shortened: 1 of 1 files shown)


Pokechu22:
Forge itself uses 4.10.3 (and prior to MinecraftForge/MinecraftForge@23541ab17519e5b1e39cb30c6c04cf8d2178da63 used 4.9).

README.md
```diff
@@ -4,3 +4,5 @@ ForgeGradle
 [![Join the chat at https://gitter.im/MinecraftForge/ForgeGradle](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/MinecraftForge/ForgeGradle?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
 Minecraft mod development framework used by Forge and FML for the gradle build system
+
+Requires Gradle 4.0 - 5.6.4
```
twixthehero, line 8:
```suggestion
Requires Gradle 4.0 - 4.10.3
```

LexManos:
The limiting factor is Artifactural, not FG.
The range of gradle versions change quite a bit depending on when people feel like testing it.
We can recommend a version but listing a range isn't gunna be feasible.

PR #686, opened 2020-06-21
Allow disabling forced System.exit for run config tasks

diesieben07:
Allow disabling the forced `System.exit` workaround at the end of run tasks, as it causes Gradle to think the task failed, making it unsuitable for use in e.g. CI.
To disable / configure this behavior run configurations can specify `forceExit false`.


gigaherz (approved):
Looks good to me. 👍

LexManos:
Sure, its in there because gradle is dumb.
If you wanna disable it yourself feel free.

PR #687, opened 2020-06-21
Dump patch reject files on failure.

covers1624:
`ApplyPatches` task by default dumps patches to `build/<taskName>/rejects.zip`, when `-PUPDATING` is set, it dumps them to `rejects/` of the current project, (`Forge/projects/forge/rejects`). 
`PatchFunction` currently doesn't save its patches, don't really know if that's something that's needed. 
`LocalContext` patches, dump their rejects next to the output file.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/diff/PatchFile.java
+++ b/src/common/java/net/minecraftforge/gradle/common/diff/PatchFile.java
@@ -9,30 +9,24 @@ import java.nio.charset.StandardCharsets;
 
 public class PatchFile {
 
-    public static PatchFile from(String name, String string) {
-        return new PatchFile(name, () -> new ByteArrayInputStream(string.getBytes(StandardCharsets.UTF_8)), false);
+    public static PatchFile from(String string) {
+        return new PatchFile(() -> new ByteArrayInputStream(string.getBytes(StandardCharsets.UTF_8)), false);
     }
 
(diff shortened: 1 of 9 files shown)


src/common/java/net/minecraftforge/gradle/common/diff/PatchFile.java
```diff
@@ -9,26 +9,32 @@
 
 public class PatchFile {
 
-    public static PatchFile from(String string) {
-        return new PatchFile(() -> new ByteArrayInputStream(string.getBytes(StandardCharsets.UTF_8)), false);
+    public static PatchFile from(String name, String string) {
+        return new PatchFile(name, () -> new ByteArrayInputStream(string.getBytes(StandardCharsets.UTF_8)), false);
```
LexManos: These are all breaking, any particular reason? you need the input name? The target name is where rejects should go.
As in if there are multiple files targeted in a single patch it should spit out multiple rej files.

covers1624: The reason was so that rejects from a single patch file get mapped to a corresponding rej file. I'll change this though.

covers1624: Fixed, turns out the way I was doing it was totally redundant, got a little confused it seems, diff4j is ass.

PR #688, opened 2020-06-23
Strip JAR signatures when using deobfuscating libraries

### LINKED ISSUE #636: Signed dependencies break remapping in some circumstances.
I've ran into an issue on FG 3+ where having a class from a signed dependency in the hierarchy of my class will cause instance members to fail remapping. 

For example I have a [TileEntity](https://github.com/Darkhax-Minecraft/Dark-Utilities/blob/3313d0f3f9b0bad3527387e78e304fa3a7a672b9/src/main/java/net/darkhax/darkutils/features/slimecrucible/TileEntitySlimeCrucible.java) class in my mod which extends the [TileEntityBasicTickable](https://github.com/Darkhax-Minecraft/Bookshelf/blob/6504639554f211115b372a93043b8218fa15ef3b/src/main/java/net/darkhax/bookshelf/block/tileentity/TileEntityBasicTickable.java) class from the Bookshelf mod which is signed. This results in inherited fields and methods from vanilla to fail remapping, so things like `this.getPos()` and `this.getWorld()` will not be remapped to their srg names. The can be confirmed by compiling against Bookshelf `4.3.62` instead of `4.3.61`. The only difference with build 62 is that signing was temporarily disabled. An example of the decompiled error class can be found [here](https://gist.github.com/Darkhax/4750529cb79632faf87afeaa8dfe5f29)

I got some help from Paul with debugging this issue and he found that an exception was being swallowed from special sources [here](https://github.com/md-5/SpecialSource/blob/378cd784325120cd6885a2e091c1ea0cb1394c39/src/main/java/net/md_5/specialsource/provider/ClassLoaderProvider.java#L74). He modified the code to print the stack trace and it provides this.

```
java.lang.SecurityException: SHA-256 digest error for net/darkhax/bookshelf/block/tileentity/TileEntityBasicTickable.class
    at sun.security.util.ManifestEntryVerifier.verify(ManifestEntryVerifier.java:223)
    at java.util.jar.JarVerifier.processEntry(JarVerifier.java:243)
    at java.util.jar.JarVerifier.update(JarVerifier.java:230)
    at java.util.jar.JarVerifier$VerifierStream.read(JarVerifier.java:484)
    at java.io.FilterInputStream.read(FilterInputStream.java:133)
    at org.objectweb.asm.ClassReader.readStream(ClassReader.java:270)
    at org.objectweb.asm.ClassReader.<init>(ClassReader.java:237)
    at net.md_5.specialsource.provider.ClassLoaderProvider.getParents(ClassLoaderProvider.java:77)
```

The jar signatures are initially correct, however the deobf process changes the file contents. The deobf version is being used during remap but retains the old signature data. 

From debug logs this is the jar being used here. 
`\.gradle\caches\forge_gradle\deobf_dependencies\net\darkhax\bookshelf\Bookshelf-1.14.4\4.3.61_mapped_snapshot_20191121-1.14.3\Bookshelf-1.14.4-4.3.61_mapped_snapshot_20191121-1.14.3.jar`

TheSilkMiner:
The presence of a signature inside a JAR prevents tampering, and so the deobfuscation process of Forge creates an effectively invalid JAR that will cause issues in various cases (e.g. when reobfuscating).
With this PR, ForgeGradle should correctly strip the signatures off of deobfuscated JARs, mitigating the above issues.

Should close MinecraftForge/ForgeGradle#636
See also MinecraftForge/InstallerTools#2

This PR is currently marked as a draft since it requires a bump to the InstallerTools dependency definition inside ForgeGradle itself.

Attempting to replicate the issue linked above gave no result: the JAR got reobfuscated correctly (checked via compiling an example mod and then decompiling it via FernFlower and CPR).

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
@@ -102,7 +102,7 @@ public class Utils {
     public static final String SPECIALSOURCE = "net.md-5:SpecialSource:1.8.3:shaded";
     public static final String SRG2SOURCE =  "net.minecraftforge:Srg2Source:5.+:fatjar";
     public static final String SIDESTRIPPER = "net.minecraftforge:mergetool:1.0.7:fatjar";
-    public static final String INSTALLERTOOLS = "net.minecraftforge:installertools:1.1.7:fatjar";
+    public static final String INSTALLERTOOLS = "net.minecraftforge:installertools:1.1.10:fatjar";
     public static final long ZIPTIME = 628041600000L;
     public static final TimeZone GMT = TimeZone.getTimeZone("GMT");
     public static final String OFFICIAL_MAPPING_USAGE =
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/tasks/RenameJarSrg2Mcp.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/tasks/RenameJarSrg2Mcp.java
@@ -42,7 +42,7 @@ public class RenameJarSrg2Mcp extends JarExec {
 
     public RenameJarSrg2Mcp() {
         tool = Utils.INSTALLERTOOLS;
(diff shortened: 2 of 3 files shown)


TheSilkMiner:
Is there anything else that needs to get addressed?

PR #689, opened 2020-06-28
Add multithread download for downloadAssets.

FledgeXu:
Downloading assets from Minecraft is always a task that needs a lot of time. However, Minecraft's assets are small and independent files that suit for multithread downloading. This PR adds the multithread downloading to `downloadAssets` task.


LasmGratel:
It will show great improvement in a high latency network

LexManos:
Not a fan of your auto-formatter redoing imports. Be sure not to do that next time.
But this works.

FledgeXu:
> Not a fan of your auto-formatter redoing imports. Be sure not to do that next time.
> But this works.

Ok, I will remember that.

PR #694, opened 2020-07-06
Allow specifying the subproject a mod uses in the run configuration

Fadenfire:
I have a multi-project workspace in IntelliJ where there's one forge sub-project that I generate run configurations from, and sub-projects for each mod. The forge sub-project depends on all the mod sub-projects. In the forge sub-project I have this in the `minecraft` block:
```gradle
runs {
    client {
        //Other configuration
        
        modProjects.each { mod ->
            mods.create(mod.name) {
                source mod.sourceSets.main
            }
        }
    }
}
```
`modProjects` is a list of all the mod sub-projects. I can then run `genIntellijRuns` and it generates run configurations with paths to the gradle build directory for each sub-project in the `MOD_CLASSES` environment variable.

The issue is when I switch IntelliJ to compile using its own compiler instead of using gradle, then `genIntellijRuns` produces incorrect paths in `MOD_CLASSES`. ForgeGradle already detects when IntelliJ is using its own compiler and will switch to using IntelliJ's build directory instead of gradle's build directory, but it constructs the paths relative to the *current* project instead of the project the mod is actually in. This means that all paths in `MOD_CLASSES` just point to the forge project, not any of the mod sub-projects.

What this pull request does is add a `modProject` property to the mods section of the runs block. So I would change the above code to:
```gradle
runs {
    client {
        //Other configuration
        
        modProjects.each { mod ->
            mods.create(mod.name) {
                modProject mod
                source mod.sourceSets.main
            }
        }
    }
}
```
Now ForgeGradle knows which project that mod belongs to and will construct the paths for `MOD_CLASSES` properly. If the `modProject` property is omitted then it'll default to the current project, which is the old behavior.

I don't know if this is the best solution to this problem, but it's the best I could come up with. It would probably be better if the project could be extracted from the source sets instead of adding a new property, but I couldn't find a way to get the project from a `SourceSet` instance using the gradle API.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/ModConfig.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/ModConfig.java
@@ -32,21 +32,38 @@ import java.util.List;
 
 public class ModConfig extends GroovyObjectSupport {
 
-    private transient final Project project;
+    private transient final Project rootProject;
 
     private final String name;
+    private Project modProject;
     private FileCollection resources;
     private FileCollection classes;
 
     private List<SourceSet> sources;
 
-    public ModConfig(@Nonnull final Project project, @Nonnull final String name) {
-        this.project = project;
+    public ModConfig(@Nonnull final Project rootProject, @Nonnull final String name) {
+        this.rootProject = rootProject;
         this.name = name;
     }
 
     public String getName() {
         return name;
+    }
+    
+    public void setModProject(Project project) {
+        modProject = project;
+    }
+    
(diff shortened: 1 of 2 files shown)


LexManos:
Gunna tag giga with this, As i d not use intellij, so it's on others to test and approve intellij related things.

SizableShrimp:
Should be fixed by ed208677be58388176fd168bdf84c1f74456250e and 5665f587843bb299ab3ceee2e22042c1c2960547. Also this is targeting FG 3 anyways.

PR #704, opened 2020-08-10
Stop double build from IntelliJ run configurations (see #703)

### LINKED ISSUE #703: [6.0] Runs generated by genIntellijRuns make the IDE build twice
Run configurations generated by `genIntellijRuns` build the mod twice before running, which is useless. Right now the generated configurations have this:
Annotation 2020-08-10 124737
As `prepareRun[...]` builds the mod for running from the IDE, the build task before it is unecessary and should not be generated at all. Right now I have to remove the unnecessary build task from each run configuration manually each time I regenerate the run configurations, which is annoying.

There is one caveat though:<br>
The mod class path (which is specified in an environment variable instead of the JDK class path property) that is configured in a generated run configuration seems to be different depending on this setting:
Annotation 2020-08-10 130117
Found at 'Settings > Build, Execution and Deployment > Build Tools > Gradle', this setting specifies whether the IDE should build with Gradle or with the builtin compiler:
- When this setting is set to 'Gradle', the class path in generated run configurations is set to the outputs of the Gradle compiler (somewhere in the `build` directory).
- When this setting is set to 'IntelliJ IDEA', the class path in generated run configurations is set to the outputs of the IntelliJ compiler (somewhere in the `out` directory).

There can be several solutions now:
1. The build task is removed from the generated run configurations. In this case, the generated run configurations should **_always_** specify the outputs of the _Gradle compiler_ as classpath, since `prepareRun[...]` always uses the Gradle compiler.
2. The `prepareRun[...]` does not compile the mod anymore (does not depend on `classes`). In this case, `run[...]` (the Gradle task) _should_ depend on `classes` so that sources are still compiled before running from Gradle (instead of the run configuration). This is a recommended solution since modders may want to modify what compiler builds the mod (with the setting shown above).
3. The `prepareRun[...]` task is removed from the run configurations. In this case, `genIntellijRuns` should do all preparing configurations such as downloading assets. I discourage this solution though as I don't know what additional configurations `prepareRun[...]` does that need to be done every time before running.

FoxSamu:
Fixes #703
I don't know how this is going to work with Eclipse and VSCode, but it seems that these also run the compiler twice before running. PLEASE VERIFY BEFORE MERGING!

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
@@ -169,7 +169,7 @@ public abstract class RunConfigGenerator
 
         TaskProvider<Task> prepareRun = project.getTasks().register("prepare" + Utils.capitalize(runConfig.getTaskName()), Task.class, task -> {
             task.setGroup(RunConfig.RUNS_GROUP);
-            task.dependsOn(prepareRuns, runConfig.getAllSources().stream().map(SourceSet::getClassesTaskName).toArray());
+            task.dependsOn(prepareRuns);
 
             File workDir = new File(runConfig.getWorkingDirectory());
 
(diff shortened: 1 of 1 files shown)


SizableShrimp:
This is an interesting one. We will investigate this and see if it should be applied to FG 5. Closing for now.

PR #705, opened 2020-08-12
Add timeout to copyURLToFile.

FledgeXu:
According to [document](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/src-html/org/apache/commons/io/FileUtils.html#line.1477):
> Warning: this method does not set a connection or read timeout and thus might block forever. Use {@link #copyURLToFile(URL, File, int, int)} with reasonable timeouts to prevent this.

We should use `copyURLToFile(URL, File, int, int)` instead of `copyURLToFile(URL, File)` to avoid forever blocked download.


FledgeXu:
Resolve the conflicts with  #707

FledgeXu:
Merge the Last Commit.

covers1624:
Can you please rebase your changes onto the `FG_4` branch.

PR #706, opened 2020-08-14
Scan Minecraft official launcher's cache, before download Client.jar

FledgeXu:
This PR lets ForgeGradle scan the Official Launcher's cache before download `client.jar` to avoid unnecessarily downloading.

I have already tested it on Windows, Linux, and macOS.


FledgeXu:
#707
Should close.

PR #684, opened 2020-06-05
Scan Minecraft install location for assets before downloading.

Unnoen:
This PR makes ForgeGradle scan the common install directory for Minecraft to pull assets from it before defaulting back to downloading them.

I have also changed the download location to use HTTPS.

Since I am unable to test on other OSs, others will have to implement the paths required for them if they want this functionality.

PR #701, opened 2020-08-06
Add Hash Checking for downloading Assets.

FledgeXu:
In some cases, for example in Chinese stupid terrible and under censorship networks,  FG will download broken assets which will cause textures loading warnings and errors.

We should check assets hash before and after it is downloaded.

PR #707, opened 2020-08-14
Check local install for assets and client.jar. Hash all assets. Closes #684 #701 #706

Unnoen:
This PR implements #684 #701 #706, the latter two by @FledgeXu.

I fixed some typos/clarity issues and also reduced the threads for asset downloading from 16 to 8.

Testing would be appreciated.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
@@ -20,6 +20,7 @@
 
 package net.minecraftforge.gradle.common.task;
 
+import net.minecraftforge.gradle.common.util.HashFunction;
 import net.minecraftforge.gradle.common.util.Utils;
 import net.minecraftforge.gradle.common.util.VersionJson;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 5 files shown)


src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
```diff
@@ -48,21 +49,32 @@ public void run() throws IOException, InterruptedException {
...
-                        getProject().getLogger().lifecycle("Downloading: " + url + " Asset: " + key);
-                        FileUtils.copyURLToFile(url, target, 10_000, 5_000);
-
+                        File localFile = FileUtils.getFile(assetsPath + File.separator + asset.getPath());
+                        if (localFile.exists()) {
+                            getProject().getLogger().lifecycle("Copying local object: " + asset.getPath() + " Asset: " + key);
+                            FileUtils.copyFile(localFile, target);
+                        } else {
+                            getProject().getLogger().lifecycle("Downloading: " + url + " Asset: " + key);
+                            FileUtils.copyURLToFile(url, target, 10_000, 5_000);
```
FledgeXu, line 68: I'm considering that we should use a unified timeout value instead of hardcode.
#705

FledgeXu (commented):
We maybe also should fix all of `FileUtils.copyURLToFile`.
#705

src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
```diff
@@ -48,21 +49,32 @@ public void run() throws IOException, InterruptedException {
         AssetIndex index = Utils.loadJson(getIndex(), AssetIndex.class);
         List<String> keys = new ArrayList<>(index.objects.keySet());
         Collections.sort(keys);
-        ExecutorService executorService = Executors.newFixedThreadPool(16);
-        CopyOnWriteArrayList<String> downloadingFailedURL = new CopyOnWriteArrayList<>();
+        File assetsPath = new File(Utils.getMCDir(), "/assets/objects");
+        ExecutorService executorService = Executors.newFixedThreadPool(8);
```
FledgeXu, line 53: I recently read Loom's Code, Maybe We can increase download Threads depends on the user's processer.
like:
```
ExecutorService executorService = Executors.newFixedThreadPool(Math.min(8, Math.max(Runtime.getRuntime().availableProcessors() / 2, 1)));
```
https://github.com/FabricMC/fabric-loom/blob/8b1268abe741d9c578a654c6a82b9b7788c43bb9/src/main/java/net/fabricmc/loom/providers/MinecraftAssetsProvider.java#L87

PR #710, opened 2020-08-17
Switch to DiffPatch

covers1624:
Switch to [DiffPatch](https://github.com/TheCBProject/DiffPatch).

- Entirely removes diff4j and all FG's custom stuff surrounding it.
- Adds a dependency on [DiffPatch](https://github.com/TheCBProject/DiffPatch).

Benifits:
- Centralized patch implementation usable across all projects.
- Support for auto hunk headers, removing large amounts of ever-changing useless junk from Forge patches (underscore for start2, `@@ -start1,len1 _,len2 @@`)
- Far better Fuzzy patch implementation. ([1.15 to 1.16 output diff](https://gist.github.com/covers1624/733ae1bd13e956f03c6d77a022530c04) and [DiffPatch log](https://gist.github.com/covers1624/c5c43e602e06501b1ea53d833e519207))

Currently, DiffPatch is on my own maven until we get review out of the way and throw them on Forge's maven.

Recommended bump to FG 3.1 as this slightly changes the generated patch output in some cases ([chooses different context](https://gist.github.com/covers1624/99cae202f54c193361de063b331f213a)).
Patches are 'baked' for userDev jars to maintain compat with older FG versions.


LexManos:
canonicalizeAccess is something that needs to be kept as discussed on discord, as modders have access transformers which will effect the patches.
It needs to be aseperate setting beside 'fuzz'. As mod workspaces should allow access changes, but disallow white space/offsets.

src/mcp/java/net/minecraftforge/gradle/mcp/function/PatchFunction.java
```diff
@@ -20,105 +20,73 @@
...
-            try {
-                return IOUtils.toString(zip.getInputStream(e));
-            } catch (IOException ex) {
-                throw new RuntimeException(ex);
-            }
-        }));
+    public void initialize(MCPEnvironment environment, ZipFile zip) throws IOException {
+        patchesDir = new File(environment.getWorkingDir(), "patches");
+        Utils.extractDirectory(e -> new File(patchesDir, e.substring(path.length())), zip, path);
     }
```
LexManos: I specifically did not want to extract these files, as it's a lot of small disc churn. Our entire process is about working in zip files, not exploded directories.

src/mcp/java/net/minecraftforge/gradle/mcp/function/PatchFunction.java
```diff
@@ -20,105 +20,73 @@
...
 
-
     @Override
     public File execute(MCPEnvironment environment) throws Exception {
         File input = (File)environment.getArguments().get("input");
+        environment.project.getLogger().lifecycle(input.getAbsolutePath());
         File output = environment.getFile("output.jar");
```
LexManos, line 50: Debug text

src/mcp/java/net/minecraftforge/gradle/mcp/function/PatchFunction.java
```diff
@@ -20,105 +20,73 @@
...
+        File rejects = environment.getFile("rejects.zip");
 
         File hashFile = environment.getFile("lastinput.sha1");
         HashStore hashStore = new HashStore(environment.project).load(hashFile);
         hashStore.add(input);
-        patches.forEach((path,data) -> hashStore.add(path, data));
+        Files.walk(patchesDir.toPath())//
+                .filter(Files::isRegularFile)//
+                .filter(Files::exists)//
+                .forEach(e -> hashStore.add(patchesDir.toPath().relativize(e).toString(), e.toFile()));
```
LexManos, line 56: Should be in memory, built like we do before.

src/mcp/java/net/minecraftforge/gradle/mcp/function/PatchFunction.java
```diff
@@ -20,105 +20,73 @@
...
-                context.save(output);
-                hashStore.save(hashFile);
-            }
+        PatchAction action = new PatchAction(environment.project);
+        action.setBase(input);
+        action.setPatches(patchesDir);
+        action.setOutput(output);
+        action.setPatchMode(PatchMode.FUZZY);//TODO Fuzzy-Access
+        //action.setMinFuzzQuality(0.98F);//Some access changes might cause the quality to be lower than this.
+        action.setMaxFuzzOffset(0);//Make sure fuzzing is an exact line match
```
LexManos: Yup, need a FUZZY access only mode.

src/mcp/java/net/minecraftforge/gradle/mcp/function/PatchFunction.java
```diff
@@ -20,105 +20,73 @@
...
-
         return output;
     }
 
     @Override
-    public  void cleanup(MCPEnvironment environment) {
-        this.patches.clear();
+    public void cleanup(MCPEnvironment environment) throws Exception {
+        Utils.deleteFolder(patchesDir.toPath());
     }
```
LexManos: This is why we work in memory, very short lived files are killer on harddrives.

build.gradle
```diff
@@ -85,6 +85,7 @@ dependencies {
     commonImplementation 'org.apache.maven:maven-artifact:3.6.0'
     commonImplementation 'org.apache.httpcomponents:httpclient:4.3.3'
     commonImplementation 'net.minecraftforge:srgutils:0.2.12'
+    commonImplementation 'codechicken:DiffPatchGradle:1.0.0.+'
```
LexManos: I would prefer if we pulled in the library itself, and implemented the tasks themselves instead of going the indirect route.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -586,12 +590,6 @@ public void apply(@Nonnull Project project) {
             if (userdevConfig.get().getSource() == null) {
                 userdevConfig.get().setSource(project.getGroup().toString() + ':' + sourcesJar.get().getBaseName() + ':' + project.getVersion() + ':' + sourcesJar.get().getClassifier() + '@' + sourcesJar.get().getExtension());
             }
-            if (!"a/".contentEquals(genConfig.get().getOriginalPrefix())) {
-                userdevConfig.get().setPatchesOriginalPrefix(genConfig.get().getOriginalPrefix());
-            }
-            if (!"b/".contentEquals(genConfig.get().getModifiedPrefix())) {
-                userdevConfig.get().setPatchesModifiedPrefix(genConfig.get().getModifiedPrefix());
-            }
```
LexManos, line 602: We need these prefixes for patch gen for old patches so it doesn't diff the crap out of existing ForgeDev workspaces.

src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
```diff
@@ -1020,82 +1016,54 @@ private File findPatched(boolean generate) throws IOException {
...
                         patcher = patcher.getParent();
                     }
+                    cache.save();
+                    Utils.updateHash(patched, HashFunction.SHA1);
                 }
-
-                cache.save();
-                Utils.updateHash(patched, HashFunction.SHA1);
+            } finally {
+                //Cleanup the patcher temp outputs manually as the daemon exists.
```
LexManos: Again, working in memory is important so we don't have to do this.

LexManos (commented):
Binary compat is a important thing, even if it is just adding a bunch of bouncer lines like you already do for most of the tasks. Also, we would need to specify somewhere in the userdev config what type of patch files we are using. As the old implementation {and other consumers of the userdev jar} have no idea about this relative header... Unless we want to bake the headers during userdev jar creation...

covers1624:
This should be ready for more review, all comments have been addressed.

PR #712, opened 2020-08-20
Fix HackyCompile for Gradle 5.1

Elec332:
This was fixed for version 5.1+, but it seems that "5.1.0" does not trigger when using Gradle version 5.1
(The compare puts "5.1" as a lower version than "5.1.0")

This pull request adresses this issue.


covers1624:
This has been resolved with FG 4.

PR #715, opened 2020-09-01
Fix the problem when local assets file broke, FG will  be stuck.

FledgeXu:
[Original Problem (Chinese)](https://v2mcdev.com/t/topic/825)

## There is a problem:
 1. ForgeGradle will check if the local assets exist. If these files exist, FG does not do hash checks and just uses them.
 2. Many third-party launchers will use the same local assets dictionary and they do not do hash check too when they download assets.
 3. When FG meets these broken local assets files, FG will be stuck.

The fix is easy, We just need do hash checks before copying files.

Btw, I changed the thread pools' size, Now Thread pools will automatically change according to your computer's process.


covers1624:
Can you please rebase your changes onto the `FG_4` branch.

PR #719, opened 2020-09-13
Remove deprecated method call TaskContainer.remove(Object)

egorponomaryov:
According to [official Gradle docs](https://docs.gradle.org/5.0/release-notes.html) usage of `TaskContainer.remove(Object)` method is deprecated in Gradle 5.0 and starting with Gradle 6.0 is error. Preferred way to replace this method is to use `Task.setEnabled(false)`.


LexManos:
That would be cool if we gave a crap about gradle 6+.
There are many other reasons were stuck on 4/5. Honestly I dont care to look at hire versions until they develop a system that makes our internal hacks not needed.

PR #722, opened 2020-10-03
Some "working" Gradle 6 support

gabizou:
[Artifactural](https://github.com/MinecraftForge/Artifactural/pull/2) | **ForgeGradle**

This adds on some hackiness to bridge support for Gradle 6.x where changes to Gradle are referenced in the commit adding the workarounds. This still functions in Gradle 4.9/5.x, but specifically does allow for Gradle 6.x to start resolving the project.

There is an odd effect that `net.minecraft:mappings-snapshot:<version>_<at>` is not being resolved correctly, but as that is a resolution issue on the IDE (IntelliJ), it does not appear to affect dependency building and project management any other way. When I get to being able to put more time into that, I will. For now, this just focuses on getting FG to work with newer versions of Gradle where feasible.

Please note that *Artifactural* is needing an update as well, and is linked above for its own changes (since it extends internal gradle repository interfaces that have changed in Gradle 6).

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/task/JarExec.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/JarExec.java
@@ -42,6 +42,7 @@ import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.TaskAction;
 
 import net.minecraftforge.gradle.common.util.MavenArtifactDownloader;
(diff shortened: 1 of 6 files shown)


gabizou:
Closing as it's been manually updated/merged/squashed in 8f62e2d64538e871cfacdcee5e53c93f65ac571f

PR #727, opened 2020-11-04
Added fg.deobf source-loading support

TimeConqueror:
* Based on tterrag works
fg.deobf now provides boolean parameter, which represents if sources should be also loaded (actually, it just _tries_ to be loaded). By default it is set to false, so it wouldn't be able to break something.
Tested with the dependencies from reports in discord, which have led to reverting tterrag's commit. No errors found.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
@@ -41,9 +41,25 @@ public class DependencyManagementExtension extends GroovyObjectSupport {
         return deobf(dependency, null);
     }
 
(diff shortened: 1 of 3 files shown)


BrainStone:
Would be much apprechiated if that also fetched the ``javadoc`` artifacts as some mods don't have a ``sources`` artifact available.

BrainStone:
Ok. While this gernerally works I came across a few issues:

- Javadocs are not downloaded if available
- It errors when a source jar is not available when downloading it is enabled
- Source jars are always downloaded instead of just when needed. Source jars are not needed on CI systems so they should only be downloaded when requested (by the IDEs for example)

sciwhiz12:
This seems to have been implemented and merged in by #746.

PR #728, opened 2020-11-05
Fix corruption of Unicode characters on Windows systems

hakanai:
I had an issue where a corrupted copy of ClientHooks.java was found inside a jar file:

...\build\tmp\expandedArchives\forge-1.16.3-34.1.0_mapped_snapshot_20201028-1.16.3-sources.jar_d036fe06dff3dbdda94c44ac227acd65\net\minecraftforge\fml\client\ClientHooks.java:183: error: unmappable character (0x81) for encoding UTF-8
        return description.endsWith(":NOFML?��?��r") ? description.substring(0, description.length() - 8)+"?��?��r" : description;

Best practices for file I/O streams is always to specify the character set. I think it's reasonable to just enforce UTF-8 in all situations, so I fixed that.

To find other cases, I added forbidden-apis to the project, and fixed all the other things it found.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,7 @@ plugins {
     id 'org.ajoberstar.grgit' version '3.0.0'
     id 'com.github.ben-manes.versions' version '0.20.0'
   //id 'com.github.johnrengelman.shadow' version '2.0.4'
+    id 'de.thetaphi.forbiddenapis' version '3.1'
 }
 
 apply plugin: 'java'
@@ -128,4 +129,10 @@ publishing {
             }
         }
     }
+}
+
+forbiddenApis {
+    // Would add 'jdk-reflection' too but HackyJavaCompile currently requires it.
+    bundledSignatures = [ 'jdk-unsafe', 'jdk-deprecated', 'jdk-non-portable', 'commons-io-unsafe-2.4' ]
+    ignoreFailures = false
 }
--- a/src/common/java/net/minecraftforge/gradle/common/config/Config.java
+++ b/src/common/java/net/minecraftforge/gradle/common/config/Config.java
@@ -24,6 +24,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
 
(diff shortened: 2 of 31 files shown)


build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
     id 'org.ajoberstar.grgit' version '3.0.0'
     id 'com.github.ben-manes.versions' version '0.20.0'
   //id 'com.github.johnrengelman.shadow' version '2.0.4'
+    id 'de.thetaphi.forbiddenapis' version '3.1'
```
3TUSK, line 6: Irrelevant change?

hakanai: No, it's relevant.

TheCurle:
UTF-8 is a change more suited for omnibus.
You've essentially made 2 different PRs here and separating them is hard.
The addition of the extra plugin with next to no explanation (one sentence isn't enough) is enough to dismiss this out of hand.
This'll take some severe convincing to be worth a pull.

hakanai:
So you want the introduction of the checks and the fixes for the checks in two separate PRs? I can do that.

Oh, also, what is omnibus in this context?

TheCurle:
I want you to explain why you added this plugin that we have no control over.
What does it do? How did it help you? Why does it belong here?

I'd suggest leaving the UTF-8 fixes until the next Omnibus (collection of small changes) PR comes along.

hakanai:
What it does is causes the build to fail if someone adds a call to Java API which is deemed unsafe, such as forgetting to pass the `Charset` parameter to a `getBytes` call. i.e., it prevents new code from reintroducing the bug that I fixed.

TheCurle:
The problem is that most of ForgeGradle works on practices that most external auditing services would call "unsafe", because it's job is literally to do low level file access to change the code of a compiled program...

There are a thousand plugins that can be used for that, and this arbitrarily chosen one would have to go through some auditing itself before we can use it in this project..

hakanai:
That's fine by me. We went through all the same things at work and eventually settled on this one. But it's up to the project to decide how to detect issues. Error Prone is another thing worth looking into. We actually run both at work.

But no, that sort of "unsafe" is different too. The sort of unsafe I'm talking about is code which behaves differently just because the user is on a Windows machine, or because the user is Turkish.

PR #729, opened 2020-11-05
Fix corruption of Unicode characters on Windows systems

hakanai:
(derived from previous PR, includes the cure but not the prevention)

I had an issue where a corrupted copy of ClientHooks.java was found inside
a jar file:

...\build\tmp\expandedArchives\forge-1.16.3-34.1.0_mapped_snapshot_20201028-1.16.3-sources.jar_d036fe06dff3dbdda94c44ac227acd65\net\minecraftforge\fml\client\ClientHooks.java:183:
error: unmappable character (0x81) for encoding UTF-8
return description.endsWith(":NOFML??r") ? description.substring(0,
description.length() - 8)+"??r" : description;

Best practices for file I/O streams is always to specify the character
set. I think it's reasonable to just enforce UTF-8 in all situations, so I
fixed that.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/task/ArchiveChecksum.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/ArchiveChecksum.java
@@ -29,12 +29,11 @@ import com.google.common.collect.Maps;
 import com.google.common.hash.Hashing;
 import com.google.common.hash.HashingInputStream;
 
-import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.Map;
 import java.util.function.Supplier;
@@ -88,7 +87,7 @@ public class ArchiveChecksum extends DefaultTask {
             }
         }
 
-        try(PrintWriter out = new PrintWriter(new BufferedWriter(new FileWriter(getOutput())))) {
+        try(PrintWriter out = new PrintWriter(getOutput(), StandardCharsets.UTF_8.name())) {
             checksums.forEach((name, hash) -> {
                 out.write(hash);
                 out.write(' ');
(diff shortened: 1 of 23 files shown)


hakanai:
The branch with the build script changes is here: https://github.com/MinecraftForge/ForgeGradle/compare/FG_3.0...ephemeral-laboratories:encoding-bug-prevention?expand=1

I won't submit a PR at this point because first the project should decide for itself whether it wants checks like this as part of its build.

covers1624:
Can you please Rebase your changes onto the `FG_4` branch.

hakanai:
Not sure what was going on there. GitHub told me I'd added a workflow file, then when I tried to undo that it said I had deleted it, then I had to undo that.

src/common/java/net/minecraftforge/gradle/common/task/ArchiveChecksum.java
```diff
@@ -29,12 +29,8 @@
 import com.google.common.hash.Hashing;
 import com.google.common.hash.HashingInputStream;
 
-import java.io.BufferedWriter;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.PrintWriter;
+import java.io.*;
```
covers1624: You have heaps of * import collapses throughout here, we don't collapse imports to *, can you un collapses all these please?

covers1624 (changes requested):

src/common/java/net/minecraftforge/gradle/common/task/ArchiveChecksum.java
hakanai: Existing imports in several other files contradict that claim, but in any case, done.

SizableShrimp:
This needs retargetting and rebasing on FG 5 to be considered.

hakanai:
I've decided to close the PR.

A few reasons:
- I found that I had some calls which weren't just fixing the encoding but other locale-related issues (i.e., I should have cut the PR down further anyway)
- New calls to the bad methods appeared and I am not that keen on fixing new bad calls every release only to find that the PR isn't accepted by the next major release anyway (before attempting to fix this again, some sort of static checks to detect the issue really need to be in place or people will keep reimplementing the same issue)
- I have stopped supporting Forge in my own mods, so this issue no longer directly affects me

PR #730, opened 2020-11-12
Introduced sourceFileEncoding setting to config.json

hakanai:
- `sourceFileEncoding` setting added to `UserdevConfigV2`.
  Kept as `String` to make serialisation easier. Could be Charset
  but would then require custom serialisation logic.
- `sourceFileEncoding` added to `TaskGenerateUserdevConfig`.
  This one is a `Charset` because type safety in build scripts
  is a good idea, but I added an overload taking `String` for
  convenience of people who don't care.
- `MinecraftUserRepo` now uses the config to get the encoding
  when writing .java files. Some other cases of writing to other
  files in the same class were fixed to use UTF-8 instead.
- `Deobfuscator` now uses UTF-8 - it looks like the string it's
  writing was originally read as UTF-8 in the method called on 
  the line above.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
@@ -1155,7 +1155,9 @@ public class MinecraftUserRepo extends BaseRepo {
                     zout.putNextEntry(Utils.getStableEntry(name));
 
                     if (name.endsWith(".java")) {
-                        String mapped = map.rename(zin, addJavadocs && vanilla.contains(name.substring(0, name.length() - 5)));
+                        String mapped = map.rename(zin,
+                                addJavadocs && vanilla.contains(name.substring(0, name.length() - 5)),
+                                true, sourceFileCharset);
                         IOUtils.write(mapped, zout, sourceFileCharset);
                     } else {
                         IOUtils.copy(zin, zout);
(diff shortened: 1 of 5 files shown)


src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
```diff
@@ -1152,7 +1156,7 @@ private File findSource(String mapping, boolean generate) throws IOException {
 
                     if (name.endsWith(".java")) {
                         String mapped = map.rename(zin, addJavadocs && vanilla.contains(name.substring(0, name.length() - 5)));
```
LexManos: Need to pass in to McpNames, it defaults to UTF8 right now, but we need to make sure we use the correct encoding.
https://github.com/MinecraftForge/ForgeGradle/blob/c077542cef99ffadaf19b434eddf1e676f76c612/src/common/java/net/minecraftforge/gradle/common/util/McpNames.java#L106

hakanai: This call is easy to change but there is another use from `TaskApplyMappings` which currently has nothing to do with the user repo. What should be done about that call?

LexManos: You could add it as an argument to that task. However the current Patcher stuff is only using it for clean. Which should be UTF8. But it could probably use the full info from the parent patchers.  Also, be sure to add it as a argument on the end, and a default. Just like normal API extension.

PR #733, opened 2020-11-21
Add an OFFLINE_MODE to Utils class.

jcollard:
Added a `Utils.initUtils(Project p)` method which can be used to read 
in a Project's parameters (and other Gradle information) for use in 
configuration.

Additionally, added an OFFLINE_MODE value that is populated by the
`--offline` flag. The OFFLINE_MODE flag is used in several locations 
throughout the Util class to prevent network access when the flag is set.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
@@ -88,6 +88,7 @@ import java.util.zip.ZipFile;
 public class Utils {
     private static final boolean ENABLE_TEST_CERTS = Boolean.parseBoolean(System.getProperty("net.minecraftforge.gradle.test_certs", "true"));
     private static final boolean ENABLE_TEST_JAVA  = Boolean.parseBoolean(System.getProperty("net.minecraftforge.gradle.test_java", "true"));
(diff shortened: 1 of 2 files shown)


jcollard:
This feature is very useful for me as I don't always have access to the internet. I tried to make it as non-intrusive as possible. If you're not using the `--offline` flag, it has no effect. 

I understand the team is busy and working on more important features that is why I took the time to write this myself. I'm really not trying to make more work for anyone. I absolutely appreciate what the Forge Team has done and continues to do. I am very grateful for it.

Is there anything I can do to help add this feature? It really is convenient for me and I'm sure there are others who would appreciate it.

TheCurle:
First off, this PR is in bad form - the advertised feature is merely an "additionally" in the description, which gives the effect that it was merely a second thought.

The offline mode checking doesn't look for anything in particular. It's to the effect of "if there's a file, use it. I don't care what it is, whether it's useful, whether it's safe, or whether it's even got anything inside it." 

This is a half baked solution and needs a lot more work before it can be considered seriously.

jcollard:
I apologize if this was perceived as bad form. That wasn't my intent.

That's good feedback; I appreciate it and will try to make something more robust.

Thanks for your time.

PR #737, opened 2020-12-23
Allow insertion of javadocs using a special token, and SRG->MCP renaming in javadocs

sciwhiz12:
Currently, there are certain difficulties with javadocs for MCP methods, fields, and parameters which make it hard to properly create and maintain.
1. **Classes and parameters do not have javadocs.**
   Although there is latent code in ForgeGradle that supports them, classes do not have an associated `classes.csv`, so there exists no way to associate javadoc with classes.
   Parameters meanwhile, do have an associated file - `params.csv` - but it doesn't contain a `desc` field. This is more easily remedied by just adding the `desc` field (and FG will automatically pick up the javadocs), but this is the burden of the creator of the MCP mappings exports.

2. **Referenced methods/fields/params' MCP names must be hardcoded into the javadoc.**
   This means that if javadoc for method A were created that references method B, and method B's MCP name is changed, all references to method B's MCP name in javadocs of all other methods/fields/params _have to be_ updated manually.

   While yes, some would say this can be solved with just <kbd>CTRL</kbd>+<kbd>F</kbd> (find and replace), it suffers from a critical flaw: all similarly-named methods/fields (or those with a partial match) will also be affected.

3. **Parameter javadoc isn't inserted automatically into method javadocs.**
   Javadocs have the `@param` tag, which is used to describe the parameters of methods (and also generic types of methods/classes). Because ForgeGradle only currently directly inserts the javadoc lines (+ some line wrapping and newline adjustments), parameters of methods must be manually documented in the method javadoc itself (which falls prey to the above point).

This PR aims to partially relieve points 1 and 3, and fully relieve point 2. ForgeGradle will now preprocess the javadocs (for methods, classes, fields), where additional processing steps can be done of the javadoc line before being linewrapped. 
 - **There is now a special javadoc tag, `@fg.insertDoc`, which can be used to insert the javadoc of a SRG member.**
    Currently, the syntax is `{@fg.insertDoc <srg name> [description]}`. `<srg name>` refers to any SRG name, of which it's javadoc will be used to replace the whole tag. This allows inclusion of parameter javadocs within the method javadoc seamlessly. If the given SRG name does not have documentation or does not exist, then the tag will be replaced with an empty string.

    The optional `description` (which extends up to the end of the tag) is a semi-temporary measure due to there being no `desc` field for the paramaters CSV file. If `description` exists, and the given SRG name does not have documentation or does not exist, then the value of `description` will be used to replace the tag instead. If both `description` and the SRG name fail, then the tag will be replaced with an empty string.

 - **SRG to MCP renaming will now be performed on javadoc lines.** (this happens after replacement of `@fg.insertDoc` tags)
    This uses the same technology as the SRG to MCP renaming done on all lines of the java, but now also applied to the javadoc lines too. Although this does suffer from the fate of being unable to reference SRG names in javadocs, there should be little/none intended use of raw SRG names in MC javadocs.

I have provided javadocs for the method which performs the preprocessing.
If requested, I will maybe consider adding some kind of argument to the `minecraft` block/extension so preprocessing of javadocs may be configured by the user (in the event that some kind of conflict happens with the e.g. SRG to MCP renaming).

This does not provide a perfect solution to the three problems. The ideal solution would be that javadocs are created dynamically by the exporter, using other methods of supplying information to build each javadoc. But that solution is hoped to be realized with the creation of MMMS. For now, this presents a more integrated solution until MMMS is fully deployed and ready.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/McpNames.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/McpNames.java
@@ -28,6 +28,7 @@ import java.io.StringReader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


Pokechu22:
Issue 3 is actually not a problem.  MCPBot already supports javadocs for parameters through `sp`/`gp`; they just end up in the method CSV in the export.  Changing the way it's exported would be possible, but would require bot changes which are probably not going to happen for a while.

SRG to MCP remapping is a good thing though.  (I haven't looked over your implementation though.)

TheCurle:
As far as we're concerned, MCPBot is deprecated and there are multiple projects (MMS, Mapsy, etc) with the express purpose of replacing it. Adjustments are unlikely in any form, which is part of why it does not support 1.16+

A more modern intake system with a more distinct format is accepted. We can adjust to whichever the community decides is better. 

So, this is a community thing. Would you prefer better documentation?

sciwhiz12:
Closing as newer solutions have been found for the above problems. Namely:
- The use of Mojang names (which at the time of initial writing of this PR was in a state of dubiousness) means point 2 is a non-issue (as Mojang names will never change within a given Minecraft version);
- Points 1 and 3 are solved in ForgeGradle 5 with the new channel providers system, allowing external providers of documentation (such as the [ParchmentMC project](https://parchmentmc.org/)) to provide their data (which includes class javadocs and method javadocs with parameters included) into FG, which then performs the insertion.

PR #738, opened 2021-01-18
Bump project to Gradle 6.8 and upgrade ForgeGradle 4 as a version

gabizou:
(this needs to basically make an FG_4.0 branch for version naming)


LexManos:
Pulled in https://github.com/MinecraftForge/ForgeGradle/commit/8f62e2d64538e871cfacdcee5e53c93f65ac571f

PR #743, opened 2021-01-21
Use the java-gradle-plugin's gradlePlugin block

ramidzkh:
Has the benefit that there is extended metadata, which allows for modern plugin application
```gradle
plugins {
    id 'net.minecraftforge.gradle' version '4.0.8'
}
```
with some code in the settings.gradle
```gradle
pluginManagement {
    repositories {
        gradlePluginPortal()
        jcenter()

        maven {
            name = "MinecraftForge"
            url = uri("https://files.minecraftforge.net/maven/")
        }
    }
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,6 +64,40 @@ license {
     exclude 'net/minecraftforge/gradle/common/util/JavaVersionParser.java'
 }
 
+gradlePlugin {
+    plugins {
+        mcp0 {
+            id = 'net.minecraftforge.gradle.forgedev.mcp'
+            implementationClass = 'net.minecraftforge.gradle.mcp.MCPPlugin'
+        }
+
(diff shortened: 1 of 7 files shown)


covers1624:
This will not be happening as you cannot specify a "4.0.+" dependency with the 'Moddern' plugin application system.

ramidzkh:
> This will not be happening as you cannot specify a "4.0.+" dependency with the 'Moddern' plugin application system.

Builds shouldn't be using dynamic plugin versions in the first place. The build might depend on something not fully documented (which is a large potion of FG), and if that changes in the plugin, builds will randomly start failing for some people and not others

Anyway, if dynamic versions are required, you can either stick to the legacy application method, or, (with some changes to the publishing) use `-SNAPSHOT`

PR #744, opened 2021-01-21
Use Project#javaexec instead of creating a JavaExec task

shedaniel:


covers1624:
This looks good, have you checked for any other dynamic task creation?

Have you published a local build (add `-local` or something to the version string temporarily), and setup a fresh forge workspace & generated a build?

shedaniel:
There are still two instances of dynamic tasks, both of them being HackyJavaCompile, which I don't want to get myself into. I have already tested this in a local forge instance.

PR #746, opened 2021-02-17
Fix fg.deobf source not loading

IzzelAliz:
The `ResolvedConfiguration` contains only jars so download it manually by MavenArtifactDownloader per IDE request.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
@@ -127,6 +127,7 @@ public class DeobfuscatingRepo extends BaseRepo {
 
     private File findSource(Artifact artifact, String mapping) throws IOException {
         File origFile = MavenArtifactDownloader.manual(project, artifact.getDescriptor(), false);
+        if (origFile == null) return null;
 
         return deobfuscator.deobfSources(origFile, mapping, getArtifactPath(artifact, mapping));
     }
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
@@ -126,12 +126,7 @@ public class DeobfuscatingRepo extends BaseRepo {
     }
 
     private File findSource(Artifact artifact, String mapping) throws IOException {
-        Optional<File> orig = findArtifactFile(artifact);
-        if (!orig.isPresent()) {
-            return null;
-        }
-
-        File origFile = orig.get();
+        File origFile = MavenArtifactDownloader.manual(project, artifact.getDescriptor(), false);
 
         return deobfuscator.deobfSources(origFile, mapping, getArtifactPath(artifact, mapping));
     }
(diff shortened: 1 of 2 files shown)


src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
```diff
@@ -126,12 +126,8 @@ private File findRaw(Artifact artifact, String mapping) throws IOException {
     }
 
     private File findSource(Artifact artifact, String mapping) throws IOException {
-        Optional<File> orig = findArtifactFile(artifact);
-        if (!orig.isPresent()) {
-            return null;
-        }
-
-        File origFile = orig.get();
+        File origFile = MavenArtifactDownloader.manual(project, artifact.getDescriptor(), false);
```
marchermans, line 129: So this indeed does solve the problem, but would download the file everytime it is requested. Or at least makes an attempt todo it.

In my eyes it would be better if this is used in combination with the old technique.
In the sense that this new way is used when then optional is empty.

Would that be an option?

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
```diff
@@ -158,7 +158,7 @@ public File deobfSources(File original, String mappings, String... cachePath) th
         if (!cache.isSame() || !output.exists()) {
             McpNames map = McpNames.load(names);
 
-            try (ZipInputStream zin = new ZipInputStream(new FileInputStream(input));
+            try (ZipInputStream zin = new ZipInputStream(new FileInputStream(original));
```
marchermans, line 161: Why? Add a comment why this change is needed. As far as I can see this change is not needed.

marchermans (changes requested):

src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
IzzelAliz: No, the old mechanic cannot retrieve sources unless a souces classifier artifact is explicitly requested(added to configuration). Same rule apply on pom file, so transitive dep is not resolving too.

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
IzzelAliz: The input file is not exist when this line is reached. It is broken and I have no idea why previous contributor would write this.

src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
IzzelAliz: And IDE (idea in my case) would only request sources when it is missing, so it won't make too much network connections. Have no idea how Eclipse attach sources but Lex said Eclipse is working iirc.

marchermans: Okey perfect.
Resolves the issue then.

src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java
marchermans: Okey Perfect resolves the issue.

marchermans (approved):

PR #747, opened 2021-02-18
[WIP] Artifact Transformers

LexManos:
Stuff @OrionDevelopment and I have been talking about.
This completely removes fg.dobf hacks in favor of Gradle's Artifact Transformers.
Unfortunately, because gradle is gradle this requires some hacks to get it to work for sources/javadocs.
Stashing here while I work on other things.
Hence WIP


LexManos:
Small update/reminderr for myself.
The reason this hasn't been worked on much, and isn't replacing fg.deobf is because it does not function on dependencies that have classifiers, or extensions. Why? Because Gradle gods decided that for some unknown damn reason.
So there are two aproaches:
1) Find a way to make it work with classifiers, I dont think this is possible as the workaround we use for that is to put the source/javadoc deps into their own config and make that config have the transformer. The reason I don't want to do this is because it means ALL the deps in that config will be transformed. Which for MOST users and MOST deps is completely a waste of time. {We don't need to remap gson/guava/lwjgl...}

2) Let both systems live side by side for all of eternity.. But give a warning for people who use the old fg.deobf system to move to non-classifiers and the new system. This is bad due to the maintenance nightmare. But it may mean that in like 5 or 6 years we could drop fg.deobf as everyone has moved on...

marchermans:
I thought I had this working on my end......

I will check tomorrow

PR #749, opened 2021-02-22
Set aPrefix and bPrefix when building patch operation

tie:
See https://github.com/MinecraftForge/ForgeGradle/issues/748#issuecomment-783071882


covers1624:
Guess I missed that, looks good.

PR #753, opened 2021-03-09
Add Gradle Version Tests (FG4)

AterAnimAvis:
As requested by Lex in [Discord](https://discord.com/channels/313125603924639766/725850371834118214/818759728879632396)

Extends the already present Java Environment checks

Sample output
```
    Build file 'Example\build.gradle' line: 14

    A problem occurred evaluating root project 'Example'.
    > Failed to apply plugin [id 'net.minecraftforge.gradle']
       > Found Gradle version Gradle 4.10.3. Minimum required is Gradle 6.8.1.

    * Try:
    Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
@@ -514,19 +514,20 @@ public class Utils {
         return toCapitalize.length() > 1 ? toCapitalize.substring(0, 1).toUpperCase() + toCapitalize.substring(1) : toCapitalize;
     }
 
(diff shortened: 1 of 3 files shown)


sciwhiz12:
Note that this won't affect substantially users trying to upgrade to FG4, because Gradle will crash, with or without this patch applied, if they update the plugin to FG4 _before_ updating Gradle to 6.8.1.

The recommended method to upgrade to FG4 will always be to upgrade the Gradle wrapper first (via `gradlew wrapper --gradle-version=6.8.1`), then changing the ForgeGradle version in the `buildscript` block to `4.1.+`.

LexManos:
Isnt the entire point of his test that it wont sci? As shown in his example?
The point of my request is to make some method of "hey you should update gradle" besides the crash of method not found in AF.

PR #755, opened 2021-03-11
Fix issues when artifacts are being downloaded concurrently

### LINKED ISSUE #697: Potential deobf Concurrency Issue
We *sometimes* see ForgeGradle builds fail on our CI with the following stack trace, when running the "runData" goal:

```
> Task :runData
Error getting artifact: mezz.jei:jei-1.15.2:6.0.2.12_mapped_snapshot_20200715-1.15.1:null@jar from  DeobfuscatingRepo
java.util.zip.ZipException: error in opening zip file
	at java.util.zip.ZipFile.open(Native Method)
	at java.util.zip.ZipFile.<init>(ZipFile.java:230)
	at java.util.zip.ZipFile.<init>(ZipFile.java:160)
	at java.util.jar.JarFile.<init>(JarFile.java:168)
	at java.util.jar.JarFile.<init>(JarFile.java:132)
	at net.minecraftforge.gradle.common.task.JarExec.apply(JarExec.java:61)
	at net.minecraftforge.gradle.userdev.util.Deobfuscator.deobfBinary(Deobfuscator.java:130)
	at net.minecraftforge.gradle.userdev.util.DeobfuscatingRepo.findRaw(DeobfuscatingRepo.java:125)
	at net.minecraftforge.gradle.userdev.util.DeobfuscatingRepo.findFile(DeobfuscatingRepo.java:81)
	at net.minecraftforge.gradle.common.util.BaseRepo.getArtifact(BaseRepo.java:81)
	at net.minecraftforge.gradle.common.util.BaseRepo.getArtifact(BaseRepo.java:41)
	at net.minecraftforge.gradle.common.util.BaseRepo$Builder$1.lambda$getArtifact$0(BaseRepo.java:127)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1361)
	at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126)
	at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:499)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:486)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
	at java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:152)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:531)
	at net.minecraftforge.gradle.common.util.BaseRepo$Builder$1.getArtifact(BaseRepo.java:127)
	at net.minecraftforge.gradle.common.util.BaseRepo$Builder$1.getArtifact(BaseRepo.java:124)
[...]
```

We have two dependencies using fg.deobf:
```
    // Runtime, Mods
    runtimeOnly fg.deobf("mezz.jei:jei-${minecraft_version}:${jei_version}")
    runtimeOnly fg.deobf("mcjty.theoneprobe:TheOneProbe-${minecraft_release}:${minecraft_release}-${top_version}")
```

When debugging this on my machine, I noticed the following:
- Gradle will resolve these dependencies concurrently (Build operations threadpool)
- If the installertools fatjar is not present locally, multiple threads may start downloading it concurrently
- There seems to be no synchronization to prevent this

My current hypothesis is: Two threads start downloading the same jar file into the same location (maven_downloader cache), Jar is corrupted -> boom.

Potential solutions I see: nuclear option is synchronizing the methods in MavenArtifactDownloader. Alternatively, the existing field `CACHE` could be converted into a LoadingCache. I think Guava will take of the underlying concurrency, but I'd have to check that first.

p.s.: Alternatively we could exclude these mods from the runData task (which we would prefer); but we have not found a way to do that.

shartte:
Fixes #697: Concurrent attempts at downloading artifacts will be synchronized to avoid corrupted files on disk

To actually reproduce this issue, the cache needs to be cleaned first (the one in %USERPROFILE%\.gradle).
In addition, usually multiple dependencies are needed, but since it's a race condition it's non-deterministic when it happens.

The output that indicates the function was called concurrently and this fix actually prevented potential data corruption looks like this:
```
Downloading net.minecraft:mappings_official:1.16.5@zip
Waiting for download of net.minecraft:mappings_official:1.16.5@zip on other thread
Downloading net.minecraftforge:installertools:1.1.10:fatjar
Waiting for download of net.minecraftforge:installertools:1.1.10:fatjar on other thread
```

For our builds on AE2 this mostly happened on the CI/CD since it starts with a clean cache everytime. On top of that this only happened for us when we had at least _two_ fg.deobf dependencies since those semeed to be processed concurrently.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
@@ -103,6 +103,8 @@ public class MavenArtifactDownloader {
          * try to download the same artifact.
          * The first thread registers a future that other threads can wait on.
          * Once it finishes, the future will be removed and subsequent calls will use the CACHE instead.
+         * We use all parameters of the function as the key here to prevent subtle bugs where the same artifact
+         * is looked up simultaneously with different resolver-options, leading only to one attempt being made.
          */
         DownloadKey downloadKey = new DownloadKey(project, artifact, changing, generated, gradle, manual);
         CompletableFuture<File> future;
--- a/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
@@ -111,7 +111,7 @@ public class MavenArtifactDownloader {
             if (activeDownload != null) {
                 // Some other thread is already working downloading this exact artifact, wait for it to finish
                 try {
-                    project.getLogger().lifecycle("Waiting for download of {} on other thread", artifact);
+                    project.getLogger().info("Waiting for download of {} on other thread", artifact);
                     return activeDownload.get();
                 } catch (InterruptedException e) {
                     throw new RuntimeException(e);
@@ -123,7 +123,7 @@ public class MavenArtifactDownloader {
                     }
                 }
             } else {
-                project.getLogger().lifecycle("Downloading {}", artifact);
+                project.getLogger().info("Downloading {}", artifact);
                 // We're the first thread to download the artifact, make sure concurrent downloads just wait for us
                 future = new CompletableFuture<>();
                 ACTIVE_DOWNLOADS.put(downloadKey, future);
--- a/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
@@ -49,12 +49,22 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


LexManos:
This looks okay, except for the download key. If your concern is that it is overwriting the two files at once. It'll do it again if the context of the call is not the exact same. So that key should probably just be the artifact/path.

shartte:
Yes, that is true, I did it this way so it could be placed at the very top, making it easier to follow whats going on.

Looking at this with some distance, Path based would be best, but would make it necessary to move this logic into the download subfunctions (since they seem to determine the path).
I could however use the Artifact, since that's the key also used for `CACHE`. Not perfect, but probably better.

Thoughts?

shartte:
I took another hard look at this and decided to just comment on _why_ the entire parameter list is used as a composite key:
Sadly there's no guarantee that for any given artifact X, the function is only called with the same tuple of (changing, generated, gradle, manual), and I want to avoid skipping certain lookups (i.e. the gradle one) just because it happened concurrently with another type of lookup that had no result.

It'd probably require greater cleanup to this class to avoid that problem entirely (i.e. ensure different sets of lookup types are not used for the same artifact)

PR #752, opened 2021-02-24
Fix incorrect mcp_config version used in notchObf binpatch

tie:
Fixes https://github.com/MinecraftForge/ForgeGradle/issues/750

Issue #750, opened 2021-02-22
Yet another issue with run tasks


tie:
Running `gradle runClient` on 1.12 project with empty caches (as in `rm -rf ~/.gradle`) fails with the same error as in #748. Importing 1.12 project in IntelliJ does some weird Gradle magic incantations and things “just work”. See also https://github.com/MinecraftForge/ForgeGradle/issues/748#issuecomment-783112444

I can reproduce the issue with the latest FG 4.0.19 and older versions down to 3.0.166 (older than that probably just don’t work with 1.12 because I was getting userdev dependency errors). That is to say that I couldn’t track down the commit (if there is any) that broke things.

<details>
<summary>Minimal build.gradle</summary>

```
buildscript {
    repositories {
        maven { url = 'https://files.minecraftforge.net/maven' }
        jcenter()
        mavenCentral()
    }
    dependencies {
        classpath 'net.minecraftforge.gradle:ForgeGradle:4.0.19' // down to 3.0.166
    }
}
apply plugin: 'net.minecraftforge.gradle'

sourceCompatibility = targetCompatibility = compileJava.sourceCompatibility = compileJava.targetCompatibility = '1.8'

minecraft {
    mappings channel: 'stable', version: '39-1.12'
    runs {
        client {
        }
    }
}

// See https://github.com/MinecraftForge/ForgeGradle/issues/627
configurations.all {
    resolutionStrategy {
        force 'org.lwjgl.lwjgl:lwjgl-platform:2.9.4-nightly-20150209'
    }
}

dependencies {
    minecraft 'net.minecraftforge:forge:1.12.2-14.23.5.2854'
}
```
</details>

<details>
<summary>Logs</summary>

- [idea sync](https://github.com/MinecraftForge/ForgeGradle/files/6022929/00-idea-project-import.log)
- [idea runClient](https://github.com/MinecraftForge/ForgeGradle/files/6022930/01-idea-run-client.log)
- [gradle runClient](https://github.com/MinecraftForge/ForgeGradle/files/6022932/10-gradle-run-client.log)
</details>

---

The suspicious bits extracted from logs:

<details>
<summary>Log (IntelliJ IDEA, project import)</summary>

```
MinecraftUserRepo Request: net.minecraftforge:forge:1.12.2-14.23.5.2854_mapped_stable_39-1.12:sources@jar
[...]

PR #762, opened 2021-04-01
Fix issues with clean build

semoro:
I guess all those jars should be rebuilt if doesn't exist, not opposite

Otherwise, for me, it was causing failures in clean build on a clean system 

Probably related to #752, #750


LexManos:
No, the logic for all of these is correct.
Calling these gets with generate == false is done during the chain to get the bin.
generate is then set to true when the src is requested.
This allows those who are not using an IDE {and thus don't have gradle find the source artifact} to skip the decompile process and use the faster binpatching process.

Seems you should use the forums/discord to actually figure out what your problem actually is.

PR #763, opened 2021-04-10
Add support for Gradle 7.0

sciwhiz12:
This PR adds support for Gradle 7.0 (while maintaining compatibility with Gradle 6.8.1).

### Summary
- Fix `HackyJavaCompile` to work on Gradle 7.0
- Adds `@Internal` annotations to non-annotated getters
- Remove use of deprecated method

### Testing
A build of ForgeGradle with these changes are available at my maven, `https://sciwhiz12.tk/maven`, the `tk.sciwhiz12.gradle:ForgeGradle:4.1.9` artifact.

This has been tested by gigaherz (as testified in the [`#gradle` channel of The Forge Project Discord server](https://discord.com/channels/313125603924639766/801175194298744902/830431346987892787)) to work with [Gradle 7.0; JDK 8u212] and [Gradle 6.8.3, JDK 15].

Note that this is not compatible by default to run on Java 16, due to the use of reflection by [Artifactural](https://github.com/MinecraftForge/Artifactural/blob/master/src/java9/net/minecraftforge/artifactural/gradle/ModifierAccess.java) (which is done to JDK modules which are strongly encapsulated in JDK 16). 

Using the following in the `gradle.properties` will allow this build of ForgeGradle to work on Java 16 (but building fails due to a bug with ForgeFlower and Java 16):
```properties
org.gradle.jvmargs=-Xmx3G -XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang.reflect=ALL-UNNAMED
```

> **Note that ForgeGradle does not officially support building/running with Java 16, as the official Java target is Java 8, the same as Minecraft Forge and Minecraft.**

### In-depth
- **Fix `HackyJavaCompile` to work on Gradle 7.0**
  _(ignore the first commit, the second commit is the one that contains the correct changes)_

Because of the removal of `JavaToolChainInternal` in Gradle 7.0, a bit of refactoring had to be done to accomodate the changes. Instead of manually creating the `CleaningJavaCompiler` (and `DefaultDeleter`) ourselves, we now call `JavaCompile#createCompiler` (a package-private method).

However, `JavaCompile#createCompiler` has different method signatures between 6.8.x and 7.0; notably, 7.0 does not have the `JavaCompileSpec` parameter (as the legacy java toolchain was removed, the `JavaToolChainInternal`, in favor of the new java toolchains system).

As a workaround, it will first check for the `createCompiler` with the parameter (6.8.x), then the one without the parameter (7.0), and fail otherwise (newer version which changes their internals again).
(See `HackyJavaCompile#createCompiler` for the relevant code)

- **Adds `@Internal` annotations to non-annotated getters**

Gradle 7.0 introduced runtime validation of tasks. One validation check is ensuring that all task properties (Gradle defines a task property as something which has a public getter; for example, a `getSources()` method creates a `sources` task property) are marked with the proper annotations.

Some ForgeGradle tasks had getters which were unannotated, which failed this validation check. To fix this, the [`@Internal` annotation](https://docs.gradle.org/7.0/javadoc/org/gradle/api/tasks/Internal.html) is added to those getters, to prevent them from running afoul of the task property annotations runtime validation check.

- **Remove use of deprecated method**

Changes the use of the deprecated `AbstractArchiveTask#getArchivePath()` with the equivalent property `AbstractArchiveTask#getArchiveFile()`. 

Note that a deprecation warning is now present for `groovy.util.XMLParser` when building with the Gradle 7.0 API, but that is due to it having been moved to `groovy.xml` in Groovy 3. A comment has been placed to point towards replacing the reference whenever ForgeGradle drops support for 6.8.x.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
@@ -26,6 +26,7 @@ import net.minecraftforge.gradle.common.util.VersionJson;
 import org.apache.commons.io.FileUtils;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.tasks.InputFile;
+import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.TaskAction;
 
 import java.io.File;
@@ -108,6 +109,7 @@ public class DownloadAssets extends DefaultTask {
         this.meta = value;
     }
 
+    @Internal
     public File getOutput() {
         return Utils.getCache(getProject(), "assets");
     }
--- a/src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/DownloadAssets.java
@@ -26,7 +26,7 @@ import net.minecraftforge.gradle.common.util.VersionJson;
 import org.apache.commons.io.FileUtils;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.tasks.InputFile;
-import org.gradle.api.tasks.Internal;
+import org.gradle.api.tasks.OutputDirectory;
 import org.gradle.api.tasks.TaskAction;
 
 import java.io.File;
@@ -109,7 +109,7 @@ public class DownloadAssets extends DefaultTask {
         this.meta = value;
     }
 
-    @Internal
+    @OutputDirectory
     public File getOutput() {
         return Utils.getCache(getProject(), "assets");
     }
--- a/src/common/java/net/minecraftforge/gradle/common/task/DownloadMCMeta.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/DownloadMCMeta.java
@@ -23,6 +23,7 @@ package net.minecraftforge.gradle.common.task;
 import org.apache.commons.io.FileUtils;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction;
 
@@ -62,6 +63,7 @@ public class DownloadMCMeta extends DefaultTask {
         return mcVersion;
     }
 
+    @Internal
     public File getManifest() {
         return manifest;
     }
--- a/src/common/java/net/minecraftforge/gradle/common/task/DownloadMavenArtifact.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/DownloadMavenArtifact.java
@@ -25,6 +25,7 @@ import net.minecraftforge.gradle.common.util.MavenArtifactDownloader;
 import org.apache.commons.io.FileUtils;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction;
 
@@ -43,6 +44,7 @@ public class DownloadMavenArtifact extends DefaultTask {
         getOutputs().upToDateWhen(task -> false); //We need to always ask, in case the file on maven/our local MinecraftRepo has changed.
     }
 
+    @Internal
     public String getResolvedVersion() {
         return MavenArtifactDownloader.getVersion(getProject(), _artifact.getDescriptor());
     }
--- a/src/common/java/net/minecraftforge/gradle/common/task/JarExec.java
+++ b/src/common/java/net/minecraftforge/gradle/common/task/JarExec.java
@@ -37,6 +37,7 @@ import org.gradle.api.file.FileCollection;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.InputFiles;
+import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.TaskAction;
 
@@ -112,6 +113,7 @@ public class JarExec extends DefaultTask {
     protected void postProcess(File log) {
     }
 
+    @Internal
     public String getResolvedVersion() {
         return MavenArtifactDownloader.getVersion(getProject(), getTool());
     }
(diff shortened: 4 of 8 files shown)


scrouthtv:
How do I use it?

I changed my `build.gradle` like this:
```
 buildscript {
     repositories {
         maven { url = 'https://files.minecraftforge.net/maven' }
+        // https://github.com/MinecraftForge/ForgeGradle/pull/763
+        maven { url = 'https://sciwhiz12.tk/maven' }
         mavenCentral()
     }
     dependencies {
-        classpath group: 'net.minecraftforge.gradle', name: 'ForgeGradle', version: '4.1.+', changing: true
+        classpath group: 'tk.sciwhiz12.gradle', name: 'ForgeGradle', version: '4.1.9', changing: true
     }
 }
-apply plugin: 'net.minecraftforge.gradle'
+apply plugin: 'tk.sciwhiz12.gradle'
 // Only edit below this line, the above code adds and enables the necessary things for Forge to be setup.
```
but compilation fails:
```

* What went wrong:
A problem occurred evaluating root project 'BigMachines'.
> Plugin with id 'tk.sciwhiz12.gradle' not found.

```

I don't know anything about Gradle so thanks for any help.

*EDIT*:
```
-apply plugin: 'net.minecraftforge.gradle'
+apply plugin: 'tk.sciwhiz12.gradle'
```
this was my problem. Undoing this change fixed it.

LexManos:
This isn't too invasive, the issue's I'm having is 

1) do we want to go down the road of hacks per Gradle version again? When do we define what versions we break. MOST modders don't give a crap about the Gradle version as long as it works. So technically we could bump it to FG5 and drop 6.8 compatibility. But that seems very early to do right now. 

2) This is still reflection and accessing internal things which breaks on J16 without manual interactions by the modder {adding the opens arguments to gradle} So the major benefit of moving to G7 {J16 support} is moot. Is there a system where we can get around the J16 opens issues without breaking old JDKs?

The main concern in J16, which is something we will have to support in the future, MC 1.17 will most likely be the stage where we officially do those changes. We just need to figure out how to do them for users {Both players in the client/server, and developers in FG} without it being a major headake.

sciwhiz12:
Adressing your points:

1. _"Do we want to go down the road of hacks per Gradle version again?"_
   It would depend on exactly what hacks are needed. Currently, ForgeGradle relies on two reflection hacks: a) `HackyJavaCompile` to compile classes while dependency resolving (otherwise, according to the comment, it would deadlock), and b) Artifactural's hacks [to modify the `MavenResolver`][mavenresolverhacks].
   
   In my opinion, if a major version requires a small amount of modification to these hacks, then we should try to support it. More major hacks (such as another `Gradle410RepositoryAdapater` situation in Artifactual) would entail not trying to support it until it can be decided to up our major version and break compatibility again.
   
   _Small note:_ The reason why Gradle 7.0 caused us to modify `HackyJavaCompile` is because of their internal refactoring to eliminate their legacy `JavaToolChain` code and move all the way to using their java toolchains feature internally. Unless Gradle plans on refactoring that or their maven resolving subsystems, it would be a fair bet that these hacks will stay working for a fair while.

2. _"This is still reflection and accessing internal things which breaks on J16 without manual interactions by the modder."_
   A point to note is (as of my investigations when I made this PR) Gradle itself does not use the Java 9 module system (no `module-infos` are present). Reflection is only denied access for modules which do not export/open the package being reflected into. This means that the hack in FG, namely the `HackyJavaCompile` task, is not affected.

   However, what is affected (and the reason why FG crashes on Java 16 and Gradle 7.+) is the reflection hacks in Artifactual, the previously mentioned reflection into `MavenResolver`. The problem there is that to get around the restriction of not being able to reflect and modify values in `final` fields, Artifactural reflects into the JDK's `Field` class to change the `modifiers` field, to bypass the restriction on changing values in `final` fields through reflection.

   In my (limited) investigation, replacing this dirty hack would need reimplementation of some parts of `MavenResolver`. I cannot attest whether that is possible or not without actually attempting, but it could be something to look into in the near future.

3. _"Is there a system where we can get around the J16 opens issues without breaking old JDKs?"_
   The `org.gradle.jvmargs` properties line should work on even Java 8, due to the `-XX:+IgnoreUnrecognizedVMOptions` part. Outside of that, the only option I know of is to eliminate the reflection hacks (in Artifactural, at least) entirely.

I would also note that while the current situation prevents ForgeGradle from running on Java 16, Gradle itself is still able to compile and test for Java 16+, even when running on Java 8. While the possibility of Mojang bumping its Java version is still just a possibility, having ForgeGradle ready to support at least compiling Java 16 code by supporting Gradle 7.0 is still a good idea.

[mavenresolverhacks]: https://github.com/MinecraftForge/Artifactural/blob/4b39e3af8d3c0fe9f0ae03391715ff586bccf56d/src/gradlecomp/java/net/minecraftforge/artifactural/gradle/GradleRepositoryAdapter.java#L143

LexManos:
So, after discussing with people on discord and deciding how we want to deal with future gradle breaks and things.
We are not going to support multiple major builds of Gradle with the same version of FG.
Thus, the prompt for FG5. We will try and keep up to date with as much of Gradle as we can, and if they do something that breaks our hacks then we will bump the FG version and limit backwards compatibility.
Also, as with each FG major version bump, the first builds until a RB/Minor revision are open to breaking changes.

This may give issues to users who have 3rd party build script extensions/libraries. Unfortunately there isn't much we can do about that without going down the rabbit hole of massive overlaid hacks like we had in FG2.

With FG3+'s switch to a data driven system, We will strive to make all MC/Forge versions {that are designed for FG3+} work on Modern FG's. There is an effort to migrate old versions {those built before FG3} to the data driven system. If you would like too help, see the #retrogradle channel in the Forge Discord.

We are also adopting a similar LTS system for FG as we do MC. Where we will support the current version, and the last version. 3rd parties are free to continue making PRs/fixes to legacy versions but the core team are going to focus on the main branches.

scrouthtv:
Is FG5 already available? There is a bug in Gradle that has been fixed in 7.0 and requires me to use 7.0+ (https://github.com/gradle/gradle/issues/14094), that's why I came here in the first place.

LexManos:
Things auto-build for every commit we push to github. See the #jernkins and #github channels on our discord.
However, a quick skim of that issue does not have anything to do with MC development as it's only certain repos which are not necessary for our dev.
Either way, as it's a dev build, use at your own risk.

scrouthtv:
Thanks a lot

PR #767, opened 2021-04-21
Update MinecraftForge maven URL

Hiroku:
For those of us living in the past


LexManos:
Don't half-ass things. This is nowhere near functional.

PR #768, opened 2021-04-21
Update FORGE MAVEN URL to use new repository

brandonhoughton:


ravenbuilder934:
#767 already exists

brandonhoughton:
Sorry - didn't mean to PR, just needed the patch so I can build against FG_2.2 rather than FG_2.3

PR #771, opened 2021-04-24
add artifact only if downloads.artifact is not null

### LINKED ISSUE #769: ForgeGradle add library even if Downloads.artifact is null
Relates #627
This issue cause problem currently if targeting 1.12.2 but I think this is a general issue so I report this.

In 1.12.2.json, there's a library without downloads.artifact, only with downloads.classifiers. ("org.lwjgl.lwjgl:lwjgl-platform:2.9.2-nightly-20140822")
Because the library without a classifier does not exist at the server, I think it's only for libraries with a classifier.
So I think ForgeGradle should not add the libraries without downloads.artifact to a pom.

I fixed this issue in [my fork](https://github.com/anatawa12/ForgeGradle) and tested it with my project, it looks to work well with 1.12.2-14.23.5.2855.

anatawa12:
Cherry-pick #770, Fixes #769 for ForgeGradle 3.x


LexManos:
We really should sunset FG3. All setups are compatible with FG4 so there should be no functional reason why people don't update.

PR #774, opened 2021-04-26
Update maven endpoint + Jenkinsfile + upload task + add warning

SizableShrimp:
This is mainly for all the whiners out there. This PR is simple in what it does. I updated the maven endpoint to the new one, along with the Jenkinsfile and upload task used by Jenkins. I have no real way to test the Jenkinsfile, but I copied the one from FG4 and modified one line to use the correct gradle task for the older version. I also modified the actual upload task to use the modified variables. A warning is now spit out every time something relating to gradle is executed warning the user that the version is unsupported and they should upgrade or see the Forge discord for more info on helping. The exact message is:
```
WARNING: You are using an unsupported version of ForgeGradle.
Please consider upgrading to ForgeGradle 4 and helping in the efforts to get old versions working on the modern toolchain. See #mcpconfig in the Forge discord for more info.
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,12 +3,12 @@
 pipeline {
     agent {
         docker {
-            image 'gradlewrapper:latest'
+            image 'gradle:jdk8'
             args '-v gradlecache:/gradlecache'
         }
     }
     environment {
-        GRADLE_ARGS = '-Dorg.gradle.daemon.idletimeout=5000 --stacktrace --info'
+        GRADLE_ARGS = '-Dorg.gradle.daemon.idletimeout=5000'
         DISCORD_WEBHOOK = credentials('forge-discord-jenkins-webhook')
         DISCORD_PREFIX = "Job: ForgeGradle Branch: ${BRANCH_NAME} Build: #${BUILD_NUMBER}"
         JENKINS_HEAD = 'https://wiki.jenkins-ci.org/download/attachments/2916393/headshot.png'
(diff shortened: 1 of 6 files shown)


CDAGaming (approved):
I can see this likely being the last warning for <1.11.2 users (As 1.12.2 still works fine in FG4). LGTM

LexManos:
The entire point of the RetroGradle people, is to do that I did with 1.12.2, and make a MCPConfig/Update the Forge branches to build the userdevs. It is completely possible for FG4 to be used on ANY MC versions. It just takes someone putting in the effort. Which is why this is a call to people to help out. If they are so concerned about the old versions working still, they can spend some time helping upgrade the toolchain.

CDAGaming:
> The entire point of the RetroGradle people, is to do that I did with 1.12.2, and make a MCPConfig/Update the Forge branches to build the userdevs. It is completely possible for FG4 to be used on ANY MC versions. It just takes someone putting in the effort. Which is why this is a call to people to help out. If they are so concerned about the old versions working still, they can spend some time helping upgrade the toolchain.

For the most part, the majority of users either lack the experience to do so, or that the effort just seems so daunting to achieve (Not to mention, likely becoming more difficult the further you try to extend the version range). That's primarily the reason most people see it as easier to just bandaid FG2.3 to continue working, even though doing so harms the RetroGradle efforts.

If it were up to me, I'd like to see a FG2.5 put into place that is able to use parts of FG4 to be more future-proof whilst still maintaining it's reliability (And being a version people can use if some MC versions never do get the RetroGradle treatment, but work with FG2.3). That way, at least in this case, both coin sides are happy (The most played versions can be ported via RetroGradle to work in FG4 and beyond and receive continued support, whilst other versions have a compatibility layer that doesn't need much support).

Edit: Another point I have while writing this is that while other MC versions can come into FG4, the current standards within the Discord consolidate that to an n-1 format, which to me and likely others seems unfair. For example, if the effort has already gone through to allow 1.12.2 to be supported via FG4, it should still receive at least some form of support and maintainability (Otherwise, what's the point if it just falls back into disarray again?).

In either case, I'm rambling a bit but this is just my two cents on the matter of legacy version support.

LexManos:
That was not an invitation to try and justify their argument. I have heard it all before, and it is all invalid. The decision has been made on how to address this. I was simply adding context to the warning message.

PR #778, opened 2021-05-02
Update publishing + remove tests + automate uploading old version format

SizableShrimp:
* Excluded all tests as they cause more issues than are worth and this version is EOL anyways.
* Modify `Jenkinsfile` to upload as both the old format (`2.3-SNAPSHOT`) and newer format (`2.3.x`)

These modifications will serve as the basis for making it easier to update older FG versions to work with the new maven + Jenkins process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ buildscript {
 apply plugin: 'java'
 apply plugin: 'idea'
 apply plugin: 'eclipse'
-apply plugin: 'maven-publish'
+apply plugin: 'maven'
 apply plugin: "com.gradle.plugin-publish"
 apply plugin: 'license'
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -357,65 +364,26 @@ pluginBundle {
...
-            artifactId = project.archivesBaseName
-            project {
-                name project.archivesBaseName
-                packaging 'jar'
-                description 'Gradle plugin for Forge'
-                url 'https://github.com/MinecraftForge/ForgeGradle'
-
-                scm {
-                    url 'https://github.com/MinecraftForge/ForgeGradle'
-                    connection 'scm:git:git://github.com/MinecraftForge/ForgeGradle.git'
```
LexManos: Poms are important...

PR #783, opened 2021-05-16
[2.0] Update to use new maven endpoint

jriwanek:
This PR fixes 2.0 to work with the new maven, along with updating the Jenkins process to match that of FG4.

-    Update maven endpoint
-    Add warning stating the version is unsupported
-    Disabled tests as they are outdated and not needed
-    Add JenkinsFile
-    Add settings.gradle
-    Use jgit for version resolving
-    Removes old, obsolete version checker


SizableShrimp:
There is no point in merging this if we currently have no solution for TeamCity to build old gradle versions (as GradleUtils is required for setup stuff each CI invocation). And I don't believe 2.0 was ever used for the latest minor version of a major MC version, so this isn't super high priority anyways. Closing.

PR #786, opened 2021-06-02
Support the new MCPConfig format

SizableShrimp:
Currently, FG does not support the new MCPConfig format and will filter it out accidentally, causing SRG files to be generated that map srg <-> srg because all the mappings are filtered out. Tested and confirmed to work.


SizableShrimp:
Added a new commit which fixes the split and extra jar generation with new MCPConfig exports. Explained in discord and in a comment in the commit. Tested and working as expected.

LexManos:
probably should just move that to use srgutils

PR #791, opened 2021-06-11
Make JarExec respect the project java toolchain

SizableShrimp:
JarExec now respects the project's java toolchain version. Tested and confirmed to use the executable from the project.


LexManos:
This would probably be better to allow configuration based on the task itself.
But default to the project level.
So Say.. I could compile for J8, but use MergeTools on J16.

LexManos:
Looks right.. lets hope it doesn't break things.

PR #792, opened 2021-06-14
Fix issue with HackyJavaCompile and java 8

SizableShrimp:
Fixes an issue where HackyJavaCompile would throw an exception trying to parse "1.8" as a number.


LexManos:
That's weird, older gradle used to support '1.8' as the version....

PR #794, opened 2021-06-17
Major cleanup, use Gradle properties and providers

sciwhiz12:
This PR does a major cleanup of ForgeGradle tasks and plugins. The list of changes are too numerous to individually list out (you can read the commits to see all changes), so here are the ~~two~~ three key and most notable points:

- **Apply the concepts found in [the Lazy Configuration chapter of the Gradle User Guide][lazy_config_gradle].** This includes the use of `Property<T>`, `ConfigurableFileCollection`, and co., providing two distinct advantages: allowing buildscript configuration of task and extension properties without needing to make more adjustments in `afterEvaluate`, and being able to avoid task creation unless necessary for execution.
- **Refactors some classes to new packages, and splitting code.** Notably, a number of methods in `Util` have moved to specialized subclasses such as `DownloadUtil` and `EnvironmentChecks`.
- **Bumps minimum Gradle version to 7.1.**

Buildscripts for user development (using the `net.minecraftforge.gradle` plugin) may need adjustments if they are manually adjusting tasks. Testing with a fresh MDK for 1.16.5 reveals no changes necessary (other than the required Gradle update and changing the ForgeGradle version).

Buildscripts for the Forge development workspace (the `MinecraftForge` repository) require more extensive changes. For reference, [my fork's `fg5` branch][forge_fork_fg5] shows the changes needed for ForgeGradle 5 from this PR to run. To run the branch, clone this branch to an adjacent directory to the repository folder named `ForgeGradle`.

As of writing this description, this PR is marked as a draft due to insufficient testing (due to my insufficient hardware). ~~<sup>/cc @SizableShrimp pls test this PR as you promised me twice</sup>~~

[lazy_config_gradle]: https://docs.gradle.org/7.0.2/userguide/lazy_configuration.html
[forge_fork_fg5]: https://github.com/sciwhiz12/MinecraftForge/tree/fg5


LexManos:
Just a note, this is the major thing pending before a RB/No more breaking changes in FG5.
We wanted to make sure we updated the modern gradle concepts to make the lifecycle of FG5/G7 consistent.
So if people could test this and give feedback that'd be great.

build.gradle
```diff
@@ -65,22 +62,20 @@ license {
 }
 
 wrapper {
-    gradleVersion = '6.8'
+    gradleVersion = '7.0.2'
```
pupnewfster: It may potentially make sense to update this to 7.1.0 just in case there are any issues that come up due to Gradle 7.1 so that we don't end up being stuck with FG5 only supporting Gradle [7, 7.1). The one thing I am unsure about is if we can update to 7.1 yet or not due to there not being a released version of [this PR](https://github.com/CadixDev/licenser/pull/30) that got merged into the licenser plugin (as the plugin got broken by 7.1, which is part of what makes me wonder if any of the things FG does broke as well).

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -20,17 +20,9 @@
 
 package net.minecraftforge.gradle.common.util;
 
-import com.google.gson.Gson;
```
pupnewfster, line 23: Seems the following imports can also be removed from this class due to being unused:
- java.net.HttpURLConnection
- java.net.URL
- java.util.function.Consumer

src/common/java/net/minecraftforge/gradle/common/util/MojangLicenseHelper.java
```diff
@@ -20,6 +20,8 @@
 
 package net.minecraftforge.gradle.common.util;
 
+import org.gradle.api.Project;
```
pupnewfster, line 23: Not touched by this but it seems that the following two imports are unused in this class so may as well be removed as the class and its imports are being touched anyways:
- java.util.function.Predicate
- java.util.regex.Pattern

src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java
```diff
@@ -124,12 +127,11 @@ public void attach(Project project) {
...
                     SimpleRepository.of(ArtifactProviderBuilder.begin(ArtifactIdentifier.class).provide(
-                            new ArtifactProvider<ArtifactIdentifier>() {
-                                @Override
-                                public Artifact getArtifact(ArtifactIdentifier artifact) {
-                                    return repos.stream().map(repo -> repo.getArtifact(artifact)).filter(Artifact::isPresent).findFirst().orElse(Artifact.none());
-                                }
-                            }
+                            artifact -> repos.stream()
+                                    .map(repo -> repo.getArtifact(artifact))
+                                    .filter(s -> s.isPresent())
```
pupnewfster, line 134: This can be `.filter(Artifact::isPresent)`

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -300,463 +294,384 @@ public void apply(@Nonnull Project project) {
...
+                    extractConstructors.configure(task -> {
+                        task.getConfig().set(downloadConfig.flatMap(DownloadMCPConfig::getOutput));
+                        task.getKey().set("constructors");
+                        task.setAllowEmpty(true);
+                        task.getOutput().set(project.getLayout().getBuildDirectory()
+                                .dir(task.getName()).map(d -> d.file("output.txt")));
+                    });
+
+                    createExc.configure(task -> {
+                        task.getConfig().convention(downloadConfig.flatMap(s -> s.getOutput()));
```
pupnewfster: This can be `task.getConfig().convention(downloadConfig.flatMap(DownloadMCPConfig::getOutput));`

src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
```diff
@@ -1420,6 +1440,7 @@ public File getZip() {
         public File getUniversal() {
```
pupnewfster, line 1458: Given it seems you are adding nullable annotations to things this probably should get one as well (and the universal field that goes with it).

pupnewfster (commented):
I did a bit of miscellaneous skimming and let my IDE do some as well and made some comments based on what I saw. I haven't had a chance to actually test this yet, but I will try to do a few tests against this tomorrow and see if anything explodes.

build.gradle
sciwhiz12: From what the image of the conversation in the comment of that PR tells me, it seems that the issue lies with the plugin's use of dynamic Groovy. If I understand it correctly, when in a class with an `extension` field and some code from within a `project.with` closure tries to resolve `extension`, it resolves `project.extension` and not `extension`.

ForgeGradle 5 should be fine with Gradle 7.1 on that basis alone (barring some deprecation warnings because they deprecated `Project#getConvention` and we cannot move away due to `JavaPluginConvention` missing methods in its extension counterpart, `JavaPluginExtension`).

However, I do think that in any case, we should consider moving to Gradle 7.1 since we are currently within the breaking changes window. But that would need confirmation from Lex if we are to move to Gradle 7.1.

SizableShrimp: Jamierocks has not yet pushed the fixed version of Licenser to the gradle plugin portal, meaning we cannot move to 7.1 at this time.

src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java
sciwhiz12: The fun thing is, I did change it to `.filter(Artifact::isPresent)`, but then javac complained that it was invalid somehow when I was loading it in via `includeBuild` and Gradle was building it for use as a plugin (see [`5561101` (#794)](https://github.com/MinecraftForge/ForgeGradle/pull/794/commits/5561101f15a09bce70359feb72f57fa5e31c918a) which reverted it). I'll have to re-investigate that, and if I can't find a solution then I'll add a comment explaining why it can't be a method reference (along with a comment suppression)

src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java
sciwhiz12: Added nullable annotations, both suggested and a few more in the same class.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
sciwhiz12: Must've missed this due to copy-pasting and forgetting to do the quick fix (it's easier to copy-paste `.flatMap(s -> s.getOutput())` then using IntelliJ inspection to fix later than copying the method reference and having to correct the class).

src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java
sciwhiz12: Further checking showed that I had Java 16 as the JDK in my IDE, which is why rebuilding from the IDE showed no errors. Switching it to Java 8 and rebuilding now made it show the same error as compiling through Gradle ("formal and actual arguments list differ in length"). This makes me reasonably certain that this is either a Java 8 compiler bug, or a change made between Java 8 and 16.

I've added a comment explaining this, along with a comment inspection suppression to remove the warning from IntelliJ IDEA.

src/common/java/net/minecraftforge/gradle/common/util/MojangLicenseHelper.java
sciwhiz12: Removed the unused imports.

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
sciwhiz12: Removed the unused imports.

build.gradle
pupnewfster: Seems they just [did](https://github.com/CadixDev/licenser/commit/6e73b6b63a93c7a2402891c757502bf33f6748ad) so, so once it finishes building this can probably be done.

sciwhiz12: Bumped minimum Gradle version to 7.1, and updated the Licenser plugin to 6.0.1. Also went through and removed uses of deprecated code, notably `JavaPluginConvention` (used for getting `SourceSet`s) in favor of `JavaPluginExtension`.

This hasn't been tested yet, would appreciate testing on Gradle 7.1 if it works (and on Gradle 7.0.x to see if the Gradle version env. check works).

src/common/java/net/minecraftforge/gradle/common/util/EnvironmentChecks.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    /**
+     * Checks the current project environment, and throws an exception if not satisfied.
+     *
+     * Current environment checks:
+     * <ul>
+     *     <li>Java version is above <em>1.8.0_101</em> (first JDK version to include Let's Encrypt certificates)</li>
+     *     <li>Gradle version is above <em>6.8.1</em> (minimum version required by ForgeGradle)</li>
```
SizableShrimp: This line is outdated

src/common/java/net/minecraftforge/gradle/common/util/JavadocAdder.java
```diff
@@ -83,27 +83,27 @@ public static String buildJavadoc(String indent, String javadoc, boolean multili
...
         {
-            return new ArrayList<String>();
+            return new ArrayList<>();
         }
 
         // return text if len is zero or less
         if (len <= 0)
         {
-            return new ArrayList<String>(Arrays.asList(text));
+            return new ArrayList<>(Collections.singletonList(text));
```
SizableShrimp: Other places in the codebase are using google commons collections. Why do something different here?

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -139,156 +150,143 @@ public void apply(@Nonnull Project project) {
...
+        downloadAssets.configure(task -> task.getMeta().set(dlMCMetaConfig.flatMap(DownloadMCMeta::getOutput)));
+
         applyPatches.configure(task -> {
-            task.setOutput(project.file("build/" + task.getName() + "/output.zip"));
-            task.setRejects(project.file("build/" + task.getName() + "/rejects.zip"));
-            task.setPatches(extension.patches);
-            task.setPatchMode(PatchMode.ACCESS);
+            final Provider<Directory> workDir = project.getLayout().getBuildDirectory().dir(task.getName());
+            task.getOutput().set(workDir.map(s -> s.file("output.zip")));
+            task.getRejects().set(workDir.map(s -> s.file("rejects.zip").getAsFile()));
```
SizableShrimp, line 183: Why the getAsFile()?

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -300,463 +298,384 @@ public void apply(@Nonnull Project project) {
...
+                    .map(ver -> "net.minecraft:client:" + ver.getVersion() + ":extra"));
+            // Add mappings so that it can be used by reflection tools.
+            project.getDependencies().add(MC_DEP_CONFIG, extension.getMappingChannel()
+                    .zip(extension.getMappingVersion(), MCPRepo::getMappingDep));
+
+            dlMCMetaConfig.configure(task -> task.getMCVersion().convention(extension.getMcVersion()));
 
             if (!extension.getAccessTransformers().isEmpty()) {
-                SetupMCPTask setupMCP = (SetupMCPTask) mcp.getTasks().getByName("setupMCP");
+                SetupMCP setupMCP = (SetupMCP) mcpParent.getTasks().getByName("setupMCP");
```
SizableShrimp: This can use named() and remove the cast

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -300,463 +298,384 @@ public void apply(@Nonnull Project project) {
...
                 extension.getAccessTransformers().forEach(f -> {
                     userdevJar.get().from(f, e -> e.into("ats/"));
-                    userdevConfig.get().addAT(f);
+                    userdevConfig.get().getATs().from(f);
                 });
             }
 
             if (!extension.getSideAnnotationStrippers().isEmpty()) {
-                SetupMCPTask setupMCP = (SetupMCPTask) mcp.getTasks().getByName("setupMCP");
+                SetupMCP setupMCP = (SetupMCP) mcpParent.getTasks().getByName("setupMCP");
```
SizableShrimp: This can use named() and remove the cast

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -300,463 +298,384 @@ public void apply(@Nonnull Project project) {
...
-                for (TaskProvider<GenerateBinPatches> task : Lists.newArrayList(genJoinedBinPatches, genClientBinPatches, genServerBinPatches)) {
-                    task.get().dependsOn(fakePatches);
-                    task.get().addPatchSet(fakePatches.getOutput());
+                final TaskProvider<CreateFakeSASPatches> fakePatchesTask = fakePatches;
+                for (TaskProvider<GenerateBinPatches> binPatchesTask : Lists.newArrayList(genJoinedBinPatches, genClientBinPatches, genServerBinPatches)) {
+                    binPatchesTask.configure(task -> task.getPatchSets().from(fakePatchesTask.flatMap(CreateFakeSASPatches::getOutput)));
                 }
             }
 
-            applyRangeConfig.get().setExcFiles(extension.getExcs());
```
SizableShrimp, line 625: Why was this removed?

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -300,463 +298,384 @@ public void apply(@Nonnull Project project) {
...
-                userdevConfig.get().setPatchesOriginalPrefix(genPatches.get().getOriginalPrefix());
-            }
-            if (!"b/".contentEquals(genPatches.get().getModifiedPrefix())) {
-                userdevConfig.get().setPatchesModifiedPrefix(genPatches.get().getModifiedPrefix());
-            }
+            userdevConfig.configure(task -> {
+                task.getTool().convention(genJoinedBinPatches.map(JarExec::getResolvedVersion)
+                        .map(ver -> "net.minecraftforge:binarypatcher:" + ver + ":fatjar"));
+                task.getArguments().addAll("--clean", "{clean}", "--output", "{output}", "--apply", "{patch}");
+                task.getUniversal().convention(universalJar.flatMap(t ->
```
SizableShrimp, line 537: This seems like callback hell, surely it can be reduced?

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -300,463 +298,384 @@ public void apply(@Nonnull Project project) {
...
+
+                // Fixup the inputs.
+                applyPatches.configure(task -> task.getBase().set(toMCPClean.flatMap(s -> s.getOutput().getAsFile())));
+                genPatches.configure(task -> {
+                    task.getBase().set(toMCPClean.flatMap(ApplyMappings::getOutput));
+                    task.getModified().set(dirtyZip.flatMap(AbstractArchiveTask::getArchiveFile));
+                });
             }
 
             {
```
SizableShrimp, line 594: Why keep the random brackets?

src/patcher/java/net/minecraftforge/gradle/patcher/tasks/GeneratePatches.java
```diff
@@ -0,0 +1,149 @@
...
+        getModifiedPrefix().convention("b/");
+        getLineEnding().convention(System.lineSeparator());
+    }
+
+    @TaskAction
+    public void doTask() throws Exception {
+        Path base = getBase().get().getAsFile().toPath();
+        Path modified = getModified().get().getAsFile().toPath();
+        Path output = getOutput().get().getAsFile().toPath();
+        getProject().getLogger().info("Base: {}", base);
```
SizableShrimp, line 58: Why the new logging statements? Isn't that what verbose already does?

SizableShrimp (changes requested):
Looks good overall, just a few comments here and there.

src/common/java/net/minecraftforge/gradle/common/util/JavadocAdder.java
sciwhiz12: Some background: this change was done during a cleanup pass, and what prompted the change was the specificity of the generics for `ArrayList`. That also prompted me to change it to a `singletonList` instead of `Arrays.asList`, since that would be more 'performant' (in the sense that `Arrays.asList` is geared towards handling arrays, while we are just passing in one value so a singleton list would be better).

While it is possible to use `ImmutableList<T>.of(T)` here instead of `Collections.<T>singletonList(T)`, my own preferrence would be using the Java-provided `Collections` method. If you are instead asking about why we are returning an `ArrayList`, it's to preserve the behavior of the method (returning a mutatable collection, whereas the three methods aforementioned do not guarantee mutability/guarnatee immutability).

If you wish, I can change it to use `ImmutableList`, but that's a negligible change with zero impact.

SizableShrimp: I meant `Lists.newArrayList`, not necessarily immutable collections.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
sciwhiz12: If you look to within the `if` condition, you can see that the rejects 'output' is set to a directory, while here it is set to a file. Because of this duality of meaning, I could not use either `RegularFile` and `Directory`, so at that point I moved it to be a `Property<File>` instead. Though in hindsight it may be possible to move it to a `Property<FileSystemLocation<?>>`, but that would need more testing if Gradle is capable of handling that (it is the superinterface of `RegularFile` and `Directory`).

src/common/java/net/minecraftforge/gradle/common/util/JavadocAdder.java
sciwhiz12: I was not fully aware that the rest of the codebase used that method, as explained what I did was a sweep of the class to remove unnecessary warnings which were the uses of `<String>` instead of `<>`. You'd have to ask the creator of the class as to why they chose to create the array list directly instead of using `Lists.newArrayList`.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
sciwhiz12: Moved to the configuration of `applyRangeConfig` (and the same with `applyRangeBaseConfig`):
https://github.com/MinecraftForge/ForgeGradle/blob/1885c3a4070460a30a9ee61025a70080cb96019c/src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java#L226

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
sciwhiz12: It can be simplified at the cost of possibly removing property metadata. The reason why this apparent callback hell exists is to at the best of my ability preserve the attached metadata to the property, so in the case that the properties are derived from tasks or other sources, Gradle can use that metadata to figure out what must be run before the task. I welcome you to try improve it, as that was my second attempt at cleaning it up (and the formatting makes it at least a bit more bearable).

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
sciwhiz12: No particular reason, I didn't think to remove them so they are left in (As to why they are there, I can only guess as to clean up and separate the code a bit). If you wish, they can be removed.

src/patcher/java/net/minecraftforge/gradle/patcher/tasks/GeneratePatches.java
sciwhiz12: They are not new, those logging statements are from the original task:
https://github.com/MinecraftForge/ForgeGradle/blob/e608466178453c402f6a2f015c94d6a32551667f/src/patcher/java/net/minecraftforge/gradle/patcher/task/TaskGeneratePatches.java#L56-L57

And `verbose` is used by DiffPatch to control whether to print out each file being patched, and each hunk being processed within each patch.

SizableShrimp: Ah, they are at the `info` level instead of `lifecycle`, so they will not normally be seen in the Gradle output. I see now.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
SizableShrimp: Keeping them is fine.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
SizableShrimp: `FileSystemLocationProperty` exists, but it fails when trying to use it as `FileSystemLocationProperty<FileSystemLocation>` because `FileSystemLocation` is an interface, and it cannot make a decorated method for it. Seems fine as-is for now.

SizableShrimp (approved):

SizableShrimp:
> 
> 
> Just a note, this is the major thing pending before a RB/No more breaking changes in FG5.
> We wanted to make sure we updated the modern gradle concepts to make the lifecycle of FG5/G7 consistent.
> So if people could test this and give feedback that'd be great.

Some preliminary testing shows it works with Forgedev for 1.17 and Userdev for both 1.16 and 1.17. Should be ready for review & merge. Since this is mostly just gradle updates to use more modern features, it shouldn't break or change much in terms of functionality. However, the more testing, the better. If anyone would like to test this still before it is merged and doesn't want to go through the trouble of cloning this locally, you can use my maven `https://sizableshrimp.me/maven` with a Group ID of `me.sizableshrimp.gradlecleanup` to test (FG5 only).

PR #796, opened 2021-06-25
Add support for `parchment` mappings channel

SizableShrimp:
Parchment data is provided from the ParchmentMC team. This PR adds support to use Parchment mappings inside of ForgeGradle for modding development. Draft because it could probably be reworked to be better and simpler, but from testing it seems to work well in its current state.


src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -92,12 +92,13 @@
                                                           //manifest doesn't include sha1's so we use this for the per-version json as well.
     public static final String FORGE_MAVEN = "https://maven.minecraftforge.net/";
     public static final String MOJANG_MAVEN = "https://libraries.minecraft.net/";
+    public static final String PARCHMENT_MAVEN = "https://maven.parchmentmc.org/";
```
kashike: Do we trust this to not go down and to be able to handle the amount of traffic that could be sent to it, or should it be mirrored (or such)?

Lanse505: Linking from Dev Discord for transparancy.

We pinged @gigaherz  flagging for us being setup to be mirrored earlier this week. Since he's been our main point of contact for forge-related maven work.

We actually encourage our partners to mirror our maven to ensure we have backup copies in-case one goes down or something happens with the data on our maven. 

So I'd say forge can feel completely free to mirror our maven to feel as safe as possible.

gigaherz: I forgot 👀 
I was going to mention it on the internal channel but got sidetracked.

Mysterious-Dev (approved):

3TUSK:
There has been a related PR (#761: expanding the channel system to be more generic). I wonder SizableShrimp's stance on that PR. That is, is there any chance that two PRs can incorporate each other?

SizableShrimp:
Ater has plans to close that PR once this one is merged I believe.

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
LexManos: We will eventually mirror it, however, we need to do some development on reposilite to allow a better proxy system. For now I think we should Not explicitly auto-add these repos, but instead rely on the modder that is adding a custom channel, to add them to their repos. 

Overall the change looks like it should just be a channel -> maven coord map.

LexManos:
Anyways, just a update on this, due to some design decisions of parchment, this is still not in a state that can be pulled. We'll add support eventually but it still needs some work.

src/mcp/java/net/minecraftforge/gradle/mcp/ChannelProviders.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Adds a {@link ChannelProvider} to the provider map using all of its defined channels in {@link ChannelProvider#getChannels()}.
+     * These channels will overwrite any previous channels.
+     *
+     * @param provider The provider to add to the channel provider map
+     */
+    public static void addProvider(ChannelProvider provider) {
+        for (String channel : provider.getChannels()) {
+            PROVIDERS.put(channel, provider);
```
LexManos: Should probably throw if someone overrrides  an existing channel.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -151,6 +151,10 @@ public void apply(@Nonnull Project project) {
                 m.artifact();
             });
         });
+        project.getRepositories().maven(e -> { // Needed for parchment mappings
```
LexManos: Should we require the end user to add this repo? Or should we add it optionally when they select the parchment mappings channel?
OR should we expose the parchment stuff as a seperate channel provider plugin?
We discussed this in Discord, but the idea is to be generic and impartial to mapping providers.
I'm a little worried about people claiming favoritism if we ship the provider ourselves.

The rest of the API for defining channel providers seems fine, and extensible enough that other services can add onto it.

SizableShrimp: Having a separate plugin for Parchment support seems reasonable if it were to be under the ParchmentMC org so that it could be updated whenever is necessary by the team themselves. This could be converted into a generic PR just adding the support to register it. I am also fine with only conditionally adding the repo if the parchment channel is in use. The benefit of having this under something Parchment controls is that it can be updated whenever, however most likely means less people overall will use it as they would have to edit their buildscripts and add the repo and plugin. But that's the whole point of not having favoritism. ¯\\\_(ツ)\_/¯

LexManos: Yes, I was thinking that they would be the ones to provide and manage the plugin. 
It's a little extra work on the modder's end. But it's just adding a repo/plugin.
The API for adding channel providers is simple enough tho. So if you can finalize that then I think that would be the best approach.
Speak with the  parchment  people and see what they think. At the end of the day it'll be like 2 classes in their plugin.

LexManos:
Alright, so this looks fine as a generic channel extension API.
The main thing is I do not want to play favorites with data sources. 
Combined with not being a fan of Parchment's rolling archive scheme.
They should be able to add a addon for this in one simple class.
Hell it could be done via a include ('URL') style build script.

reserveword:
How do I add a custom channel? I tried
```groovy
channelProviders.addProvider(new ChannelProvider() {
    Set<String> getChannels() {
        println "channels"
        return ImmutableSet.of('yarn')
    }

    File getMappingsFile(MCPRepo mcpRepo, Project project, String channel, String version) {
        String desc = "net.fabricmc:yarn:" + version + ":v2@zip";
        println 'get' + desc
        return MavenArtifactDownloader.manual(project, desc, false);
    }
})
```
but it didn't work. It seems that `channelProviders` get reset after I changed it. error says:
```
Error getting artifact: net.minecraft:mappings_yarn:1.0.1:null@zip from  MCPRepo
java.lang.IllegalArgumentException: Unknown mapping provider: yarn_1, currently loaded: [stable_nodoc, snapshot_nodoc, stable, official, snapshot]
```

gigaherz:
Please do not post in random issues for help questions, we have a [forum](https://forums.minecraftforge.net/), and a [discord server ](https://discord.minecraftforge.net/) where you can ask questions.

PR #801, opened 2021-06-28
[FG5] Fix DownloadAssets race condition

### LINKED ISSUE #800: [FG5] downloadAssets sometimes fails due to a race condition
**Edit**: After looking at it again more closely, this is yet another race condition, similar to #755, but this time affecting Vanilla 
assets.

This happens because `minecraft/icons/minecraft.icns` and `icons/minecraft.icns` both use the same remote path.
The download task will download the same file twice, and if it's unlucky, two threads will be writing to the same file on disk, resulting in a corrupt download.

~~Sometimes asset just fail to download correctly. Since it involves remote access, this can be expected.
On CI/CD servers this can be very annoying since it leads to false-positive build failures.~~

~~It would be nice for one of the following:~~
- ~~Use JDK11 HttpClient in HTTP/2 mode to support pipelining and remove the need to do 8 parallel downloads~~
  - ~~After checking, resources.download.minecraft.net does not seem to support HTTP2 for some reason (S3?)~~
- ~~Make the parallel download count configurable, since this could be the issue in the CI/CD cloud environments (this is pure speculation)~~
- ~~Evaluate HTTP status codes correctly, this could explain the failed hash check (server error response written to file, instead of the actual asset)~~
  - ~~status codes 400 and up should already lead to IOExceptions, it's unclear if the server uses some other weird status code to indicate rate-limiting(?)~~
- ~~Configurable retry count~~

Example issue from a GitHub Actions build:

```
Downloading: https://resources.download.minecraft.net/99/991b421dfd401f115241601b2b373140a8d78572 Asset: icons/minecraft.icns
Downloading: https://resources.download.minecraft.net/99/991b421dfd401f115241601b2b373140a8d78572 Asset: minecraft/icons/minecraft.icns
icons/minecraft.icns Hash failed.
minecraft/icons/minecraft.icns Failed.

java.io.FileNotFoundException: /home/runner/.gradle/caches/forge_gradle/assets/objects/99/991b421dfd401f115241601b2b373140a8d78572 (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at net.minecraftforge.gradle.common.util.HashFunction.hash(HashFunction.java:63)
	at net.minecraftforge.gradle.common.task.DownloadAssets.lambda$run$0(DownloadAssets.java:70)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

shartte:
Fixes #800.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/tasks/DownloadAssets.java
+++ b/src/common/java/net/minecraftforge/gradle/common/tasks/DownloadAssets.java
@@ -36,8 +36,10 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -52,6 +54,8 @@ public abstract class DownloadAssets extends DefaultTask {
         AssetIndex index = Utils.loadJson(getIndex(), AssetIndex.class);
         List<String> keys = new ArrayList<>(index.objects.keySet());
         Collections.sort(keys);
+        removeDuplicateRemotePaths(keys, index);
+
         File assetsPath = new File(Utils.getMCDir(), "/assets/objects");
         ExecutorService executorService = Executors.newFixedThreadPool(8);
         CopyOnWriteArrayList<String> failedDownloads = new CopyOnWriteArrayList<>();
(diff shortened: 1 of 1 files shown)


LexManos:
This should be done in the first pass, filter out all the keys that are duplicates.

shartte:
Which first pass? Stream.distinct wouldn't help here since the keys are unique, the paths aren't.

LexManos:
yes, but you can  filter it out above the loop instead of inside it.
Or since the key is only used for the pretty debug message, you could not use the key at all.

shartte:
Hm yes, i wanted to do this with minimal changes.
Do you see any benefit to filtering it out before? We're iterating over it anyway, and I didn't judge modifying the key list as easier to grasp in my opinion.

p.s.: if you think filtering it out first is better, i'll just do it.

LexManos:
Filtering out first keeps the logical blocks small and simple to understand. 
As well as removing the long lasting set that is continuously built/kept for the entire lifecycle of the method.
Also, this method would spam the console every time you run the task because you do it outside of checking if the value actually needs to be downloaded.

You could just do something like:
```
Set<String> seen;
keys.stream().filter(k -> seen.add(object.get(k).hash))
seen.empty() // Tho not needed, the Jitter will see it go out of use and kill it.
```

shartte:
Alright I've changed it. Is this better?

PR #809, opened 2021-07-18
Add back API method used by Mixin-Gradle plugin

shartte:
The Sponge Mixin Gradle plugin uses the method removed by sciwhiz in their cleanup to set extra mappings produced by the Mixin AP.

Once that plugin moves to the newer property-based methods, this compatibility fix can be removed again.
I've created a PR on their end: https://github.com/SpongePowered/MixinGradle/pull/30

But adding back the API method here will at least allow for compatibility in the meantime.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/tasks/RenameJarInPlace.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/tasks/RenameJarInPlace.java
@@ -38,11 +38,14 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import java.io.File;
 import java.io.IOException;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 public abstract class RenameJarInPlace extends JarExec {
     private final Provider<Directory> workDir = getProject().getLayout().getBuildDirectory().dir(getName());
     private final Provider<RegularFile> temp = workDir.map(s -> s.file("output.jar"));
+    private File extraMapping;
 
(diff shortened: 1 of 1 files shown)


LexManos:
We are not going to accept a 'compatibility fix', updating MixinGradle is on my long list of TODOs.
If it's to be fixed, it should be done correctly.
I, or someone else who knows what they are doing, just needs to get time.

shartte:
Well, I did submit the PR to MixinGradle which does adapt them to the new API.
Your comment on that PR makes it sound like you want to release FG/Forge without those fixes in place though.

What I am afraid of is breaking the out-of-box experience for modders in such a way that several mods will just not build on FG5, which will just be a bad experience all around (and it's preventable!).

I don't believe there's going to be any changes to the affected task for 1.17, so should I just ask Mumfrey to evaluate the PR on his end, or do you really want to hold off on it?

LexManos:
There are quite a bit of changes necessary for Mixin to properly support J16, and the new Module system.
And there are most definitely gunna be changes necessary in Mods that mean they simply 'not build' in FG5/J16.
Let me be perfectly clear, It is well known that Forge's official stance on coremods is to not support them and actively discourage their use.
We WILL NOT be going out of our way to add a incorrect hack back in to allow them to function. If you want it, it must be done correctly.

It will be done correctly after the inital Forge release/Mixin update. It's a very low priority right now so those who do have the knowledge to do it correctly on our team are working on other higher priorates. Also, I can't speak for Mumfrey, as he's not part of our team. But I expect once things get out and are settled he will deal with what he needs to on his end.

PR #812, opened 2021-07-20
Fix Gradle 8 warnings

sciwhiz12:
## Lazily evaluate updated tokens map for run tasks

The updated tokens map is created by passing in the contents of the `runtimeClasspath` configuration. However, the run tasks are configured with the updated tokens map, which means the configuration is queried on configuration, which is deprecated and liable to be invalid in Gradle 8 according to [1] and the warnings when running FG.

To avoid resolving the configuration at configuration time, then, the updated tokens map is lazily created and used for configuring the run tasks.

The `WrappedArgument` class provides a way to lazily provide a (optionally post-processed) string, overriding the `toString()` method to return the string instead of the default object behavior. This allows us to pass the object directly into the various configuration methods of JavaExecSpec, which internally store `Object`s which are lazily `toString()`ed when the actual arguments/environment variables/system properties are needed for execution.

As far as I can tell, this is the only place we need to do this: while the configuration is also resolved as part of the individual IDE run configuration generation tasks, it is resolved within the task execution as opposed to the task configuration of the run tasks.

[1]: [Gradle 7.1 User Guide, "Viewing and debugging dependencies", subsection "Resolving unsafe configuration resolution errors"](https://docs.gradle.org/7.1/userguide/viewing_debugging_dependencies.html#sub:resolving-unsafe-configuration-resolution-errors)

##  Add explicit task ordering to prevent Gradle dep. warnings

Gradle 7 now outputs more warnings for different wrong behaviors. One of these behaviors is when a task has an input set indirectly to the output of another task, without any ordering or dependency defined between them (whether via `dependsOn`, `mustRunAfter`, or passing a `Provider` from the output to the input).

To fix this, we define some tasks to `mustRunAfter` other tasks which produce their inputs without being connected via `dependsOn` or `Provider`s. Through this, Gradle will ensure that the producing tasks run _before_ the consuming tasks, but will not force the producers to run (which is what `dependsOn` specifies).

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
@@ -37,6 +37,7 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 
 import com.google.common.base.Strings;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Streams;
@@ -54,6 +55,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


sciwhiz12:
Closing until a more comprehensive review can be done as to whether ForgeGrade 5 can be fitted to support Gradle 8. A future major version of FG will support Gradle 8.

PR #822, opened 2021-08-22
fix: DynamicJarExec using unsupported type for @InputFiles

AterAnimAvis:
Declaring a processor for Post Processing using the ForgeGradle Patcher Plugin will result in hard failing the build in `:postProcess`.
```
Caused by: org.gradle.internal.typeconversion.UnsupportedNotationException: Cannot convert the provided notation to a File or URI: {}.
The following types/formats are supported:
  - A String or CharSequence path, for example 'src/main/java' or '/usr/include'.
  - A String or CharSequence URI, for example 'file:/usr/include'.
  - A File instance.
  - A Path instance.
  - A Directory instance.
  - A RegularFile instance.
  - A URI or URL instance.
  - A TextResource instance.
```

This is due to `DynamicJarExec#getData` returning an unsupported type for the @InputFiles which Gradle then tries to convert.
See https://docs.gradle.org/current/userguide/more_about_tasks.html#table:incremental_build_annotations for reference.

Currently fixed in a non breaking way, this may want a second look in 5.2.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/tasks/DynamicJarExec.java
+++ b/src/common/java/net/minecraftforge/gradle/common/tasks/DynamicJarExec.java
@@ -22,14 +22,17 @@ package net.minecraftforge.gradle.common.tasks;
 
 import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.provider.MapProperty;
+import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.InputFiles;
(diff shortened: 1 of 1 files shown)


LexManos:
Gradle doesn't support a map of files.. as files... 
That'd dumb, might as  well design the 'proper' way now. As there is no point doing half measures. And it should be able to be done in a bin compat way.

SizableShrimp:
@AterAnimAvis Where are we at on this?

SizableShrimp:
Fixed by 014a07ed144888fe6c39adf1df6d51c05490b8b6.

PR #825, opened 2021-09-25
Properly implement the SourceSet selection when updating mappings

marchermans:
This PR fixes an issues that prevented custom sourcesets from being selected when performing mappings updates using the task.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -148,54 +148,7 @@ public class UserDevPlugin implements Plugin<Project> {
         });
         downloadAssets.configure(task -> task.getMeta().set(downloadMCMeta.flatMap(DownloadMCMeta::getOutput)));
 
-        final boolean doingUpdate = project.hasProperty("UPDATE_MAPPINGS");
-        final String updateVersion = doingUpdate ? (String) project.property("UPDATE_MAPPINGS") : null;
-        final String updateChannel = doingUpdate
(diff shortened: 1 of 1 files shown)


src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -293,6 +252,50 @@ public void apply(@Nonnull Project project) {
 
             extension.getRuns().forEach(runConfig -> runConfig.token("asset_index", finalAssetIndex));
             Utils.createRunConfigTasks(extension, extractNatives, downloadAssets, createSrgToMcp);
+
+            if (doingUpdate) {
+                logger.lifecycle("This process uses Srg2Source for java source file renaming. Please forward relevant bug reports to https://github.com/MinecraftForge/Srg2Source/issues.");
+
+                final String[] updateSourceSets = project.hasProperty("UPDATE_SOURCESETS") ? ((String) project.property("UPDATE_SOURCESETS")).split(";") : new String[] { SourceSet.MAIN_SOURCE_SET_NAME };
```
ChampionAsh5357, line 259: Main thing of note when it comes to this is that two things must be true:
- If the property exists, all source sets wanting to be updated must be specified. This may just want to include main by default, but that more QoL than anything.
- The SourceSets listed must have no whitespace whether leading or trailing.

These two are mainly considerations, but neither I believe are necessary for change as long as those nuances are specified somewhere.

LexManos: Honestly this whole thing needs to be moved to a separate reusable script, considering its copy pasta in multiple of the plugins we expose.

marchermans:
What is the state on this?
We seem to need this relatively often.

SizableShrimp:
Fixed in d4bf98051eda53e9fe0f323e5dad623c05e86fd4.

PR #828, opened 2021-10-09
Fix eclipse run configurations for multi-projects

sciwhiz12:
This PR fixes an issue where referencing the source set of another project from a mod's sources in a run configuration causes the output folder to be referenced for that mod in the generated Eclipse runs to be incorrect.

See the commit message for more details:

> Originally, the Eclipse runs generator uses the eclipse model of the current project to retrieve the classpath, in order to make a mapping of source set names to their corresponding output folders. This is important so the mod code loaded by FML is the classes compiled by Eclipse, not by Gradle.
> 
> However, there's a problem when a run config uses multiple source sets with the same name from different projects
> 
> First, because only the current project's eclipse model is queried, the run generator has no information on the output folders for source sets outside the current project. Second, there is no way to retrieve the project for a source set in order to get the eclipse model for those projects because source sets do not carry with them their owning project's information. Third, because the map keys are only source set names, there is no way to retrieve the proper output folder for source sets which share the same name but come from different projects.
> 
> To solve this, there are now two maps: one is the `allOutputs` map which is the original map of source set names to output folders, and the new one is `sourceSetsToOutputs`, which tracks `SourceSet` instances (using an `IdentityHashMap`) to a corresponding project-specific map of source set names to output folders. For each `SourceSet` that belongs to a project, they all share the same project-specific outputs map. This allows looking up first by source set for the project-specific outputs map (falling back to the original shared outputs map if not present), then by source set name.
> 
> To create the `sourceSetsToOutputs` map, all projects within the Gradle build are iterated over and retrieved for the `eclipse` plugin if it is present (which is usually the case when running from Eclipse), and then retrieving the eclipse model object and classpath from there.

---

**Post-merge note:** The `allOutputs` map was removed because of the revelation that it will never normally be queried.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
@@ -34,6 +34,7 @@ import org.w3c.dom.Element;
 import javax.annotation.Nonnull;
 import javax.xml.parsers.DocumentBuilder;
 import java.io.File;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.LinkedHashMap;
@@ -98,7 +99,6 @@ public class EclipseRunGenerator extends RunConfigGenerator.XMLConfigurationBuil
     }
 
     static Stream<String> mapModClassesToEclipse(@Nonnull final Project project, @Nonnull final RunConfig runConfig) {
-        final Map<String, String> allOutputs = new HashMap<>();
         final Map<SourceSet, Map<String, String>> sourceSetsToOutputs = new IdentityHashMap<>();
 
         project.getRootProject().getAllprojects().forEach(proj -> {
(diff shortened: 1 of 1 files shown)


LexManos:
Please provide a example zip/git of the setup so we can reproduce and test.

CreativeMD:
Here is an example setup for testing: [1.18_multi_mod.zip](https://github.com/MinecraftForge/ForgeGradle/files/8384064/1.18_multi_mod.zip)

**About the setup**
It uses McJty's [MultiWorkspace](https://github.com/McJtyMods/MultiWorkspace/tree/1.18_tech). The `workspace` folder should "connect" all mods together and generate a launch command to start all of them at the same time. This test setup only contains the mod `CreativeCore`. In the root directory there is a gradle project which should links all subprojects together.

Executing the `gradlew genEclipseRuns` command in the root directory generates launch files in `workspace`, which unfortunately contains a wrong reference to the bin folder of the CreativeCore project:
`CreativeCore%%...PATH...\1.18_multi_mod\workspace\bin\main;...`

Here is the same setup with the fixed ForgeGradle version (newest + this pr): [1.18_multi_mod_fixed.zip](https://github.com/MinecraftForge/ForgeGradle/files/8384175/1.18_multi_mod_fixed.zip)

Using the same command the generated launch files have now the correct path:
`CreativeCore%%...PATH...\1.18_multi_mod\CreativeCore\bin\main...`

PR #830, opened 2021-11-06
Fix IntelliJ downloading sources

SizableShrimp:
Newer IntelliJ versions prefix the `DownloadSources` task with a colon, although there may be other cases that make IntelliJ automatically prefix a subproject. The check in `addRepoFilters` has been changed to check all starting task names if they end with `DownloadSources` to future proof this. This fix allows clicking the `Download sources` button on dependency class files with ForgeGradle enabled to run without issues and attempt to download sources.


LexManos:
We'll see if this works, got to love stupid IDE specific things.

PR #845, opened 2022-03-23
[1.2] Fix forge 1.7.10 build 

gravit0:
Fixed version of #843
Forge 1.7.10 cannot be built from source due to old URLs in ForgeGradle

- Fixing Outdated URLs

The fix allows building Forge 1.7.10 from source

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/common/Constants.java
+++ b/src/main/java/net/minecraftforge/gradle/common/Constants.java
@@ -50,11 +50,11 @@ public class Constants
     public static final Closure<Boolean> CALL_FALSE = new Closure<Boolean>(null){ public Boolean call(Object o){ return false; }};
 
     // urls
-    public static final String MC_JSON_URL      = "http://s3.amazonaws.com/Minecraft.Download/versions/{MC_VERSION}/{MC_VERSION}.json";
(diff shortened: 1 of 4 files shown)


TheCurle:
Sorry, we are not accepting PRs to this branch.

However, this issue can be trivially worked around in your userdev build.gradle, so this change is not necessary.

gravit0:
@TheCurle How can I do that?
The `./gradlew setupForge`works successfully
But `./gradlew build` downloads corrupted `forge-installer` file which contains `Moved Permanently`
And the build process stops

TheCurle:
This is an issue tracker, not the Gradle documentation. Wink wink.

LexManos:
You are more then welcome to help in the RetroGradle project on our discord to try and update older versions to the new dynamic toolchain. Thus allowing older versions to be built with non-EOL Gradle versions. However, this does not change the fact that developing with those versions are not supported. This is purely an opportunity for people, like yourself, who claim that *have* to use those versions to stand up, and do it correctly. It's all about manpower. We are not going to allocate any to it because they are old unsupported versions. However we're more then willing to manage and give pointers on where the grunt work needs to be done.

(Curle, we should really standardize this message)

PR #848, opened 2022-04-14
convert MANIFEST_URL into a property in FG_5.0

InkerBot:


TheCurle:
No, *again*.

We are not going to let you change the Mojang launcher metadata. That can only be used to facilitate hacks and cheats.

PR #861, opened 2022-06-01
Fix project name in Eclipse and VSCode run generators

Maaxed:
This fixes an issue with Eclipse and VSCode run generators where runs are generated with the wrong project name.

Changing the name of the eclipse project in the build.gradle file of a mod, ie adding the following line 
```groovy
eclipse.project.name = "Test" // This name must be different from the gradle project name for the error to appear
```
would previously result in an error in the generated run files.

With this PR, the run generators use the name of the eclipse project as declared in the build.gradle to fill the PROJECT_ATTR attribute of the eclipse run xml and the projectName property of the VSCode run json.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
@@ -68,7 +68,7 @@ public class EclipseRunGenerator extends RunConfigGenerator.XMLConfigurationBuil
             {
                 rootElement.setAttribute("type", "org.eclipse.jdt.launching.localJavaApplication");
 
-                elementAttribute(javaDocument, rootElement, "string", "org.eclipse.jdt.launching.PROJECT_ATTR", project.getName());
+                elementAttribute(javaDocument, rootElement, "string", "org.eclipse.jdt.launching.PROJECT_ATTR", getEclipseProjectName(project));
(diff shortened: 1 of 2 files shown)


SizableShrimp:
Can you provide the exact error?

Maaxed:
In Eclipse I get 
image
and I cannot run the client/server unless I change "Projet" from "Example" to "Test" in the configuration

And in VSCode, I get
image
when running the runClient configuration

Maaxed:
To be more precise, I tested by unzipping the MDK in a folder called Example and just added the following line to the build.gradle
```groovy
eclipse.project.name = "Test"
```

When running the genEclipseRuns task, in the *.launch xmls I get
```xml
<stringAttribute key="org.eclipse.jdt.launching.MODULE_NAME" value="Example"/>
```
instead of
```xml
<stringAttribute key="org.eclipse.jdt.launching.MODULE_NAME" value="Test"/>
```
and in the VSCode launch.json when running the genVSCodeRuns, I get
```json
"projectName": "Example",
```
instead of
```json
"projectName": "Test",
```

sciwhiz12 (approved):
Confirmed using Eclipse that the issue exists using `eclipse.project.name = 'Boopity'`, preventing execution of the run configurations, and that this PR's changes fixes the issue.

SizableShrimp (approved):
Looks good, thank you.

PR #863, opened 2022-06-09
[Feature] Implement support for generating JarJar dependencies and cleaning of the generated POMs

marchermans:
#### JarJar Task
This new feature allows for the optional generation of JarJar compatible Jar in Jar shadow jars.
The feature is optional and needs to be enabled with a call to `jarJar.enable()` anywhere during project evaluation. Or better said, any call to any of the methods in the `jarJar` extension (more on it later) which would configure the task, will enable it (for example by pinning a version, also more on that later).
By default this adds a single `JarJar` task which gets enabled by the call to `jarJar.enable()`, afterwards the task then processes the registered configurations and generates the additional `from(...)` statements so the files are included into the resulting jar.

#### JarJar project extension
This PR adds a new extension known as `jarJar` which handles the configuration for all `JarJar` tasks (but a call to `jarJar.enable()` will only enable the default task named `jarJar`, as opposed to enabling all, however, all other methods in this class operate on all `JarJar` instances unless otherwise stated).

#### JarJar configuration
This PR adds a new configuration which is by default added to all `JarJar` tasks, named `jarJar`, and is the source of the dependencies which are embedded into the jar during build.

#### Filtering the dependencies
The `JarJar` task has a method: `dependencies(Action<DependencyFilter> configurator)` which can be used to filter the dependencies which get included into the jar, by setting included or excluded dependency specifications.

The configurations of the included dependencies can also be done on a project wide level by invoking the same method on the `jarJar` extension.

##### Example of excluding a specific dependency:
```groovy
tasks.register('greeting', net.minecraftforge.gradle.userdev.tasks.JarJar) {
    dependencies {
        exclude(dependency('com.google.gson.gson:gson:2.9.0'))
    }
}
```

##### Example of excluding a group of dependencies:
```groovy
tasks.register('greeting', net.minecraftforge.gradle.userdev.tasks.JarJar) {
    dependencies {
        exclude(dependency('com.google.gson.*')) //Will exclude all dependencies whoes group matches the regex.
    }
}
```

##### Exclusively including a dependency or a group of dependencies 
```groovy
tasks.register('greeting', net.minecraftforge.gradle.userdev.tasks.JarJar) {
    dependencies {
        include(dependency('com.google.gson.gson:gson:2.9.0'))
        include(dependency('com.google.gson.*')) //Will include all dependencies whoes group matches the regex.
    }
}
```

#### Dependency version pinning
Since JarJar is first and foremost a way for multiple mods to supply dependencies it is likely a common occurance that two mods provide the same dependency. To that effect JarJar does not support fixed version definitions for the dependencies a user tries to include. However, to allow for the user to force JarJar to include a particular version as a preference of the developer version pinning can be used. To pin a version of a dependency pass it and the requested version to `jarJar.pin(...)` as follows:
```groovy
dependencies {
    implementation(group: 'com.google.code.gson', name: 'gson', version: '[2.0,3.0)') {
        jarJar.pin(it, "2.8.0")
    }
}
```

#### Dependency range setting
Since in many scenarios it is preferable to compile against a fixed version, but JarJar requires a version range to be given the plugin supports setting a range on a dependency which is then included in the generated metadata: 
```groovy
dependencies {
    implementation(group: 'com.google.code.gson', name: 'gson', version: '2.8.0') {
        jarJar.ranged(it, "[2.0,3.0)")
    }
}
```
This will tell JarJar to generate a metadata entry for this dependency which supports the 2.x version range of the dependency while your project hard compiles against the 2.8.0 version of the library.

#### Dependency version changing
If you want to compile against a fixed version in your project, but want to include a different version in the jar for minimal support, this can be achieved by combining the above two techniques:
```groovy
dependencies {
    implementation(group: 'com.google.code.gson', name: 'gson', version: '2.9.0') {
        jarJar.ranged(it, "[2.0,3.0)")
        jarJar.pin(it, "2.8.0")
    }
}
```
In this configuration the project would compile against 2.9.0, while including a jar of version 2.8.0, and specifying that it supports the 2.x version range of the artifact.

#### Implementation details
- Tasks of the type JarJar allow for the registration of additional configuration (besides the default jarJar one) via the `configuration(Configuration configuration)` method.
- Tasks of the type JarJar can have their default configuration removed by invoking the `setConfiguration(List<Configuration> configurations)` method, which overrides all already registered configurations.
- Tasks of the type JarJar allow for the filtering of the included dependencies which are pulled from the registered configurations
- All tasks of type JarJar are automatically reobfuscated (since they are supposed to be used at runtime).
- In practice there is currently no support for multi-project workloads inside of FG, to that end, and eventhough things like the `DependencyFilter` support filtering on projects, JarJar does not support packaging other projects and **will silently ignore** them.

#### Publishing a JarJar artifact to maven.
To publish the artifact and its cleaned dependencies it suffices to setup the following maven publication configuration:
```groovy
    publications {
        mavenJava(MavenPublication) {
            from components.java
            jarJar.component(it)

            //Other statements related to configuring the POM go here
        }
    }
```

In case you have multiple JarJar tasks you can configure each of the publications to use a particular task as its source:
```groovy
    publications {
        mavenJava(MavenPublication) {
            from components.java
            jarJar.component(it, tasks.jarJar) //From the jarJar task (note jarJar here alone does not suffice since that is an extension object)

            //Other statements related to configuring the POM go here
        }
        mavenJavaOther(MavenPublication) {
            from components.java
            jarJar.component(it, tasks.jarJarOther) //From the jarJarOther tasks

            //Other statements related to configuring the POM go here
        }
    }
```

This will add the required artifact from all `JarJar` tasks (or just from the one given) as well as configure the generated POM file to include the required dependencies.

Invoking `jarJar.fromRuntimeConfiguration()` (Either on the extension object, or on a JarJar task) will add the runtime configuration to the selected configurations to draw from. Note this requires the user to properly configure the filtering of the dependencies to include to prevent a massive mess from being generated, but this should at least provide a shortcut so that as minimal of a duplication of code in the build.gradle file is generated.

### Processing of generated POMs.
Due to the need of handling additional dependencies during POM generation (which requires a cleaned up POM to properly work) FG gains with this PR the ability to clean the POMs generated from Maven artifacts.
This cleaning ability can now also be used when the project does not use JarJar.

By invoking `fg.component(it)` when configuring a publication like so:
```groovy
    publications {
        mavenJava(MavenPublication) {
            from components.java
            fg.component(it)

            //Other statements related to configuring the POM go here
        }
    }
```
The given publication will have the POM analyzed during POM generation and then have its dependency on Forge stripped as well as all obfuscated dependencies will be reverted to their original dependency notation (so without the `_mapped_` suffix).

### Example of minimal buildscript:
<details>
  <summary>build.gradle</summary>
 
  ```groovy
  buildscript {
    repositories {
        maven { url = 'https://maven.minecraftforge.net' }
        mavenCentral()
        mavenLocal()
    }
    dependencies {
        classpath group: 'net.minecraftforge.gradle', name: 'ForgeGradle', version: '5.1.+', changing: true
    }
}

plugins {
    id 'eclipse'
    id 'maven-publish'
}

version = '1.0'
group = 'com.yourname.modid' // http://maven.apache.org/guides/mini/guide-naming-conventions.html
archivesBaseName = 'modid'

// Mojang ships Java 17 to end users in 1.18+, so your mod should target Java 17.
java.toolchain.languageVersion = JavaLanguageVersion.of(17)

println "Java: ${System.getProperty 'java.version'}, JVM: ${System.getProperty 'java.vm.version'} (${System.getProperty 'java.vendor'}), Arch: ${System.getProperty 'os.arch'}"
minecraft {
    mappings channel: 'official', version: '1.18.2'
    runs {
        client {
            workingDirectory project.file('run')
            property 'forge.logging.markers', 'REGISTRIES'
            property 'forge.logging.console.level', 'debug'
            property 'forge.enabledGameTestNamespaces', 'examplemod'
            mods {
                examplemod {
                    source sourceSets.main
                }
            }
        }

        server {
            workingDirectory project.file('run')
            property 'forge.logging.markers', 'REGISTRIES'
            property 'forge.logging.console.level', 'debug'
            property 'forge.enabledGameTestNamespaces', 'examplemod'
            mods {
                examplemod {
                    source sourceSets.main
                }
            }
        }

        gameTestServer {
            workingDirectory project.file('run')
            property 'forge.logging.markers', 'REGISTRIES'
            property 'forge.logging.console.level', 'debug'
            property 'forge.enabledGameTestNamespaces', 'examplemod'
            mods {
                examplemod {
                    source sourceSets.main
                }
            }
        }

        data {
            workingDirectory project.file('run')
            property 'forge.logging.markers', 'REGISTRIES'
            property 'forge.logging.console.level', 'debug'
            args '--mod', 'examplemod', '--all', '--output', file('src/generated/resources/'), '--existing', file('src/main/resources/')
            mods {
                examplemod {
                    source sourceSets.main
                }
            }
        }
    }
}

sourceSets.main.resources { srcDir 'src/generated/resources' }

repositories {
    mavenCentral()
}

dependencies {
    minecraft 'net.minecraftforge:forge:1.18.2-40.1.48'

    implementation fg.deobf(group: 'com.google.code.gson', name: 'gson', version: '[2.0,3.0)')
    jarJar fg.deobf(group: 'com.google.code.gson', name: 'gson', version: '[2.0,3.0)') {
        jarJar.pin(it, "2.8.0")
    }
}

jar.finalizedBy('reobfJar')

publishing {
    publications {
        mavenJava(MavenPublication) {
            from components.java
            jarJar.component(it)
        }
        mavenJavaSpecific(MavenPublication) {
            from components.java
            jarJar.component(it, tasks.jarJar)
        }
        mavenJavaCleaned(MavenPublication) {
            from components.java
            fg.component(it)
        }
    }
    repositories {
        maven {
            url "file://${project.projectDir}/mcmodsrepo"
        }
    }
}

tasks.withType(JavaCompile).configureEach {
    options.encoding = 'UTF-8' // Use the UTF-8 charset for Java compilation
}

jarJar{
    enable()
    dependencies {
        include(dependency(group: 'com.google.code.gson', name: 'gson', version: '[2.0,3.0)'))
        exclude(dependency(group: 'com.google.code.gson', name: 'gson', version: '2.9.0'))
    }
}

  ```
  
</details>


Lanse505:
Feedback was requested.
Looked it over real quick and it looks good.
However I will leave you with a rather nasty stare for naming it "JarJar" >_>

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
```diff
@@ -41,10 +57,96 @@ public Dependency deobf(Object dependency) {
...
+                }
+                final NodeList dependencies = dependenciesNode.getAt(QName.valueOf(MAVEN_POM_NAMESPACE + "*")); //grab all dependency nodes in a neat list;
+
+                final List<Node> dependenciesNodeList = (List<Node>) dependencies.stream()
+                        .filter(Node.class::isInstance)
+                        .map(Node.class::cast)
+                        .collect(Collectors.toList());
+
+                dependenciesNodeList.stream()
+                        .filter(el -> hasChildWithText(el, MAVEN_POM_NAMESPACE + "artifactId", "forge") && hasChildWithText(el, MAVEN_POM_NAMESPACE + "groupId", "net.minecraftforge"))
```
SizableShrimp: Needs the same treatment for fmlonly.

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
```diff
@@ -41,10 +57,96 @@ public Dependency deobf(Object dependency) {
...
+        final List<Node> childList = (List<Node>) (children.stream()
+                .map(Node.class::cast)
+                .collect(Collectors.toList()));
+
+        childList.forEach(el -> el.setValue(expectedValue));
+    }
+
+    private String getVersionFrom(final String version) {
+        if (version.contains("_mapped_")) {
+            return version.split("_mapped_")[0];
```
SizableShrimp: Is this what other code spots use? To me it would make more sense to substring.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -72,15 +75,25 @@
...
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
 
 public class UserDevPlugin implements Plugin<Project> {
-    private static final String MINECRAFT = "minecraft";
+
+    public static final String JAR_JAR_TASK_NAME = "jarJar";
+    public static final String JAR_JAR_GROUP = "JarJar";
```
SizableShrimp: I think task groups should be lowercase

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -294,6 +310,15 @@ public void apply(@Nonnull Project project) {
             extension.getRuns().forEach(runConfig -> runConfig.token("asset_index", finalAssetIndex));
             Utils.createRunConfigTasks(extension, extractNatives, downloadAssets, createSrgToMcp);
         });
+
+        project.afterEvaluate(projectAfter -> {
+            final NamedDomainObjectContainer<RenameJarInPlace> reobfJars = (NamedDomainObjectContainer<RenameJarInPlace>) project.getExtensions().getByName("reobf");
```
SizableShrimp: This can just use `reobfExtension` from outside the lambda.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -316,4 +341,25 @@ private NamedDomainObjectContainer<RenameJarInPlace> createReobfExtension(Projec
...
+        project.getTasks().register(JAR_JAR_TASK_NAME, JarJar.class, jarJar -> {
+            jarJar.setGroup(JAR_JAR_GROUP);
+            jarJar.setDescription("Create a combined JAR of project and selected dependencies");
+            jarJar.getArchiveClassifier().convention("all");
+
+            jarJar.getManifest().inheritFrom(((Jar) project.getTasks().getByName("jar")).getManifest());
+            jarJar.from(extension.getSourceSets().getByName("main").getOutput());
+
+            jarJar.configuration(project.getConfigurations().getByName(JAR_JAR_DEFAULT_CONFIGURATION_NAME));
+            jarJar.exclude("META-INF/INDEX.LIST", "META-INF/*.SF", "META-INF/*.DSA", "META-INF/*.RSA", "module-info.class");
```
SizableShrimp: Should this be configured for users automatically?

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -316,4 +341,25 @@ private NamedDomainObjectContainer<RenameJarInPlace> createReobfExtension(Projec
...
+        JavaPluginExtension extension = project.getExtensions().getByType(JavaPluginExtension.class);
+        project.getTasks().register(JAR_JAR_TASK_NAME, JarJar.class, jarJar -> {
+            jarJar.setGroup(JAR_JAR_GROUP);
+            jarJar.setDescription("Create a combined JAR of project and selected dependencies");
+            jarJar.getArchiveClassifier().convention("all");
+
+            jarJar.getManifest().inheritFrom(((Jar) project.getTasks().getByName("jar")).getManifest());
+            jarJar.from(extension.getSourceSets().getByName("main").getOutput());
+
+            jarJar.configuration(project.getConfigurations().getByName(JAR_JAR_DEFAULT_CONFIGURATION_NAME));
```
SizableShrimp: I think this would be better if you store the configuration you created just a few lines earlier.

src/userdev/java/net/minecraftforge/gradle/userdev/dependency/DefaultDependencyFilter.java
```diff
@@ -0,0 +1,153 @@
...
+        this.project = project;
+        this.ownerTask = ownerTask;
+    }
+
+    /**
+     * Exclude dependencies that match the provided spec.
+     *
+     * @param spec
+     * @return
+     */
```
SizableShrimp: Duplicate javadocs should be deleted from overrides.

src/userdev/java/net/minecraftforge/gradle/userdev/dependency/DependencyFilter.java
```diff
@@ -0,0 +1,121 @@
...
+import org.gradle.api.artifacts.ExternalModuleDependency;
+import org.gradle.api.artifacts.ResolvedDependency;
+import org.gradle.api.file.FileCollection;
+import org.gradle.api.specs.Spec;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+
+public interface DependencyFilter {
```
SizableShrimp, line 40: Is this copied from somewhere?

src/userdev/java/net/minecraftforge/gradle/userdev/dependency/DependencyFilter.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Exclude dependencies that match the provided spec.
+     *
+     * @param spec
+     * @return
+     */
```
SizableShrimp: If you don't document these lines then it's best to delete (same for later methods).

src/userdev/java/net/minecraftforge/gradle/userdev/dependency/DependencyFilter.java
```diff
@@ -0,0 +1,121 @@
...
+    Spec<? super ArtifactIdentifier> dependency(Closure<Boolean> spec);
+
+    boolean isIncluded(ResolvedDependency dependency);
+
+    boolean isIncluded(ExternalModuleDependency dependency);
+
+    boolean isIncluded(ArtifactIdentifier dependency);
+
+    final class ArtifactIdentifier {
```
SizableShrimp, line 111: Don't we already have one of these?

src/userdev/java/net/minecraftforge/gradle/userdev/jarjar/JarJarProjectExtension.java
```diff
@@ -0,0 +1,165 @@
...
+import org.gradle.api.publish.maven.MavenPublication;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class JarJarProjectExtension extends GroovyObjectSupport {
+    public static final String EXTENSION_NAME = "jarJar";
+    private static final String MAVEN_POM_NAMESPACE = "{http://maven.apache.org/POM/4.0.0}";
```
SizableShrimp: Does this namespace need to be duplicated?

src/userdev/java/net/minecraftforge/gradle/userdev/jarjar/JarJarProjectExtension.java
```diff
@@ -0,0 +1,165 @@
...
+
+    private final Project project;
+
+    public JarJarProjectExtension(final Project project) {
+        this.project = project;
+    }
+
+    public void enable() {
+        if (project.getTasks().findByPath(UserDevPlugin.JAR_JAR_TASK_NAME) != null) {
+            Objects.requireNonNull(project.getTasks().findByPath(UserDevPlugin.JAR_JAR_TASK_NAME)).setEnabled(true);
```
SizableShrimp: I believe you should use getByPath, which will throw a nice exception saying what was missing. Also, what about other jar jar tasks in the configuration that aren't the default? Do people have to state enabled = true in those custom tasks? Seems a bit weird.

src/userdev/java/net/minecraftforge/gradle/userdev/jarjar/JarJarProjectExtension.java
```diff
@@ -0,0 +1,165 @@
...
+
+            mavenPublication.pom(pom -> {
+                pom.withXml(xml -> {
+                    final NodeList potentialDependenciesList = xml.asNode().getAt(QName.valueOf(MAVEN_POM_NAMESPACE + "dependencies"));
+                    Node dependenciesNode;
+                    if (potentialDependenciesList.isEmpty()) {
+                        dependenciesNode = xml.asNode().appendNode("dependencies");
+                    } else {
+                        dependenciesNode = (Node) potentialDependenciesList.get(0);
+                    }
```
SizableShrimp: Slight code duplication hmm.

SizableShrimp (changes requested):

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
marchermans: Yeah it uses the same code, I wondered as well. But I just copied it.

marchermans: Or more or less adapted it
The deobfuscating repo uses this technique to get the version and other information, which is the only place where it effectively does this, hence it being identical.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
marchermans: That is indeed not needed anymore and a remnant from me testing this with shadow code.
Removing since it is not needed.

src/userdev/java/net/minecraftforge/gradle/userdev/dependency/DependencyFilter.java
marchermans: Derived is the better word: The shadow plugin function as inspiration.

src/userdev/java/net/minecraftforge/gradle/userdev/dependency/DependencyFilter.java
marchermans: Yes but not as narrowly scoped, which is the point here.
Since I need to validate both ResolvedDependencies as well as Dependencies the one in artifactural is too widely scoped and requires me to supply information i don't have in every situation.

Hence a small reimplementation with just the fields that are needed / supported.

src/userdev/java/net/minecraftforge/gradle/userdev/jarjar/JarJarProjectExtension.java
marchermans: Not needed. The default task is always registered and the null check above takes core of it.

And no if you create a custom task task it will be enabled by default. The UserDevPlugin creates the default task and sets that to disabled.

SizableShrimp: I see. then just use a variable?

marchermans: That could work, hold on will do.

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
```diff
@@ -41,10 +50,40 @@ public Dependency deobf(Object dependency) {
...
+
+                final List<Node> dependenciesNodeList = (List<Node>) dependencies.stream()
+                        .filter(Node.class::isInstance)
+                        .map(Node.class::cast)
+                        .collect(Collectors.toList());
+
+                dependenciesNodeList.stream()
+                        .filter(el -> MavenPomUtils.hasChildWithText(el, MavenPomUtils.MAVEN_POM_NAMESPACE + "artifactId", "forge") && MavenPomUtils.hasChildWithText(el, MavenPomUtils.MAVEN_POM_NAMESPACE + "groupId", "net.minecraftforge"))
+                        .filter(el -> MavenPomUtils.hasChildWithText(el, MavenPomUtils.MAVEN_POM_NAMESPACE + "artifactId", "fmlonly") && MavenPomUtils.hasChildWithText(el, MavenPomUtils.MAVEN_POM_NAMESPACE + "groupId", "net.minecraftforge"))
+                        .forEach(el -> el.parent().remove(el));
```
SizableShrimp: This is not gonna work. Has to be one filter call to be functional. `(artifactId == "fmlonly" || artifactId == "forge") && groupId == "net.mcf")`

src/userdev/java/net/minecraftforge/gradle/userdev/dependency/DefaultDependencyFilter.java
```diff
@@ -0,0 +1,104 @@
...
+                if (it instanceof ArtifactIdentifier) {
+                    final ArtifactIdentifier identifier = (ArtifactIdentifier) it;
+                    return (dependency.getGroup() == null || Pattern.matches(dependency.getGroup(), identifier.getGroup())) &&
+                            (dependency.getName() == null || Pattern.matches(dependency.getName(), identifier.getName())) &&
+                            (dependency.getVersion() == null || Pattern.matches(dependency.getVersion(), identifier.getVersion()));
+                }
+
+                return false;
+            }
+        });
```
SizableShrimp, line 76: ~~What exactly is this doing?~~ Ah I see, there is another interface method for dependency(Closure<Boolean>)

SizableShrimp (approved):

SizableShrimp (approved):

Issue #609, opened 2019-08-05
Gradle 5.5 dependency artifacts resolution feature


Bluexin:
Gradle 5.5 was released, with a feature which looks like it could be very interesting to use instead of the hacks currently being used to deobfuscate mc, see [here](https://docs.gradle.org/5.5/release-notes.html#transforming-dependency-artifacts-on-resolution).
The nice part about it would be the use of an official API, which means more forward compatibility guarantees.
(just to make it clear, I'm not demanding a change now, just probing for opinions)

LexManos:
We know about this, it dropped sometime after we built our system.
Feel free to start working on a PR. But I'm not going to. Our system works for now and I am so tired of digging in gradle's shitty ecosystem that I am not going to touch this with a 100ft pole.

On top of that, last I looked 5.5 still had the issue where it'll randomly 'clean' your cache for.you. Just making random libraries and resources dissapear even if actively used by projects/ides that are open.

AbrarSyed:
I saw the new Gradle 5.5 feature, and am considering starting development on FG4 or something that will be built the whole new gradle modules and dependency attributes systems. My only conceptual issue is that it would be difficult to apply arbitrary access transformers (assuming those are still a thing). Also, this would work probably work well for userDev but idk how it would for forge development...

Then again, ive been out of the game for some time, and don't have a clear idea of how stuff works in the current FG3.

josephworks:
BTW, Gradle 5.6 was released.

mezz:
Please see this comment and the one below it for some prior discussion on this topic: https://github.com/MinecraftForge/ForgeGradle/pull/577#issuecomment-485280600

Based on that, I believe we still need to wait for this to be fixed: https://github.com/gradle/gradle/issues/9050

mezz:
cc @Cypher121 on this topic

Cypher121:
Yes, unless something major changed, this gradle bug makes it impossible to select artifacts to transform on anything except their artifact type, so it's either deobfuscate every jar dependency or none of them.

SizableShrimp:
Closing for now until gradle gets their act together. See discussion on #865 for more info.

PR #865, opened 2022-06-13
Add forgeDeobf configuration for dependencies deobfuscating with artifact transformer

sasanquaa:
This pull request proposes a configuration to be used for deobfuscating Minecraft codes related to #609 #747. 
- It makes use of Gradle's artifact transformer and lazy configuration to manually resolve dependencies. 
- Only the dependencies added to the configuration will be deobfuscated. 

Example usage:
```gradle
buildscript {
    repositories {
        maven { url = 'https://maven.minecraftforge.net/' }
    }
    dependencies {
        classpath 'net.minecraftforge.gradle:ForgeGradle:5.1.43'
    }
}

java.toolchain.languageVersion = JavaLanguageVersion.of(8)

minecraft {
    mappings channel: 'official', version: '1.16.5'
}

repositories {
    ivy {
        url "https://download.nodecdn.net/containers/reforged/server/release"
        patternLayout {
            artifact "[revision]/[artifact].[ext]"
        }
        metadataSources {
            artifact()
        }
    }
}

dependencies {
    minecraft 'net.minecraftforge:forge:1.16.5-36.2.20'
    forgeDeobf ':Pixelmon-1.16.5-9.0.2-server:9.0.2'
}
```

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -47,6 +47,7 @@ import net.minecraftforge.gradle.userdev.tasks.JarJar;
 import net.minecraftforge.gradle.userdev.tasks.RenameJarInPlace;
 import net.minecraftforge.gradle.userdev.util.DeobfuscatingRepo;
 import net.minecraftforge.gradle.userdev.util.Deobfuscator;
+import net.minecraftforge.gradle.userdev.util.DeobfuscatorTransformer;
 import net.minecraftforge.gradle.userdev.util.DependencyRemapper;
 import net.minecraftforge.srgutils.IMappingFile;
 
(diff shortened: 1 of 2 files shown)


SizableShrimp:
What about sources? (https://github.com/gradle/gradle/issues/19207)

sasanquaa:
Hmm, yeah this doesn't seem to work with sources though. Like in the issue, it is not triggering the transformer

marchermans:
This is a known problem, and currently Gradle is not able to support this.
There are at least two known issues with this:

1. Sources can not be transformed, IDEs in particular use `ArtifactResolutionQuery` instances which do not use configurations or attributes as such they will, at least as of now, never use transformers.
2. Dependencies which have dependencies marked as runtime only (like for example LWJGL with its natives) will not work when they are passed through transformers since as of now no IDE can handle a dependency with two different files in two different usecase configurations (IDEA for example will mark both as being a runtime dependency making compiling the mod fail).

As such I will be closing this PR for now, if you want to talk about this, you can come visit our Discord, or possibly even send me a DM and I will show you every possible way I have tried to make this work over the last two years. I personally have at least 3 issues open (one on IDEAs youtrack and two on Gradles repo) which handle this.

Greets,

Orion

LexManos:
Just a clarification as it's somewhat related. A ArtifactTransformer would be useful for someething  like AccessTransformers, as yes it would slightly desync the source files, it woulld be worth it for the cleanup of decompile/cache files. Soo  if you want to look into  something llike that. It may be worth it.

PR #866, opened 2022-06-17
Add option to generate Eclipse launch groups

Matyrobbrt:
This PR adds an option in the MinecraftExtension (`generateLaunchGroups()`), which will make the `runX.launch` configurations be a launch group, running `prepareX` and the game sequentially, in this order, and as such, having the same behaviour as IntelliJ. 
I'm not sure if the option should be on by default, as it affects launch times (eclipse compiling on file save), needing to run a gradle task.

Draft until I manage to test all contexts in which the configuration can be run.


Matyrobbrt:
Merged with #868

PR #867, opened 2022-06-17
Improve Gradle configuration cache support

PaintNinja:
This PR improves support for [Gradle's configuration cache feature](https://docs.gradle.org/7.4.2/userguide/configuration_cache.html) by reducing `getProject()` usages and marking incompatible tasks accordingly so that the configuration cache can be used for compatible tasks.

Note that this is my first time working on a Gradle plugin, so merge with caution - please perform your own tests (I might've overlooked something) and let me know if I've missed anything or done something wrong so that I can improve. :)


marchermans (commented):
This looks okey....ish.

That said how much do we need of this @SizableShrimp with regards to gradle 8?

PR #868, opened 2022-06-17
Overhaul IDE run configurations

Matyrobbrt:
This PR overhauls run configs by:
- making running `prepareRunX` configurable (previously this task was never even run for eclipse)
- adding an option that will copy gradle resource output to the IDE-specific out folders (bin / out)
- allowing the generation of run folders, which for IDEA will group all the configurations for a project, and for Eclipse will append the folder name to the `.launch` files

A possible idea was also having a FG-specific configuration file (that documents all the available options) that is not tracked by VCS, which allows the end user to independently configure IDE-specific FG settings, though that is still undecided.


Matyrobbrt:
At the request of @SizableShrimp, this PR has been merged with #866.

TheCurle:
Second

src/common/java/net/minecraftforge/gradle/common/util/MinecraftExtension.java
```diff
@@ -127,4 +136,34 @@ public void accessTransformer(Object file) {
...
+
+    public void copyIDEResources() {
+        copyIDEResources.set(true);
+        // Prepare runs need to be forced in this scenario
+        enableEclipsePrepareRuns();
+        enableIdeaPrepareRuns();
+    }
+    public boolean copiesIDEResources() {
+        return copyIDEResources.get();
+    }
```
SizableShrimp: These are not good. These should use abstract property getters.

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -491,4 +497,27 @@ public static String replaceTokens(Map<String, ?> tokens, String value) {
...
+                task.dependsOn("eclipse");
+                project.getTasks().named("eclipse").configure($ -> $.doLast($$ -> {
+                    final EclipseModel eclipse = project.getExtensions().findByType(EclipseModel.class);
+                    task.configure(eclipse, project);
+                }));
+            });
+        } else {
+            project.getLogger().warn("'eclipse' plugin is not found! Eclipse resources will not be copied.");
+            project.getTasks().create("copyEclipseResources", Task.class); // Create dummy class
+        }
```
SizableShrimp: Perhaps we should enable both plugins automatically if this option is enabled?

src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
```diff
@@ -93,7 +97,68 @@ protected Map<String, Document> createRunConfiguration(@Nonnull final Project pr
...
+
+                    if (mc.copiesIDEResources()) {
+                        final Element copyIde = gradleDocument.createElement("listEntry");
+                        copyIde.setAttribute("value", project.getTasks().getByName("copyEclipseResources").getPath());
+                        tasks.appendChild(copyIde);
+                    }
+                }
+                rootElement.appendChild(tasks);
+                gradleDocument.appendChild(rootElement);
+            }
```
SizableShrimp: I really dislike the use of empty blocks for organization.

SizableShrimp (changes requested):
Can't really comment on the functionality of the eclipse stuff as it's not my IDE of choice. Additionally, this contains some breaking changes for some protected methods in RunConfigGenerator. The question is; do we consider this API "public" enough to warrant requiring backwards-compatibility?

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
Matyrobbrt: Enabling the eclipse plugin will make all the eclipse files be generated, creating unnecessary files.

SizableShrimp: How so?

Matyrobbrt: The presence of the eclipse plugin will make the `copyEclipseResources` task depend on the `eclipse` task for proper configuration. The `eclipse` task will generate the eclipse files.

SizableShrimp: And only people in eclipse will be running the eclipse task... Also this seems like bad design, running `eclipse` is no longer necessary to import into eclipse, so this should be designed to be more versatile.

src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
Matyrobbrt: I just followed the style of the existing run config generator codebase.

src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
```diff
@@ -68,6 +69,7 @@ public class RunConfig extends GroovyObjectSupport implements Serializable {
     private Boolean client; // so we can have it null
     private Boolean inheritArgs;
     private Boolean inheritJvmArgs;
+    private final Property<String> folderName;
```
SizableShrimp: This is not gonna work. RunConfigs are serializable objects. using properties here is a bit useless and should be avoided.

src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
```diff
@@ -77,6 +79,7 @@ public RunConfig(final Project project, final String name) {
         this.name = name;
 
         this.mods = project.container(ModConfig.class, modName -> new ModConfig(project, modName));
+        this.folderName = project.getObjects().property(String.class).convention(project.getName());
```
SizableShrimp: Why this convention?

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
SizableShrimp: I don't really like warning logs just cuz a plugin is a missing. I am still thinking that it would be nice if we applied these plugins automatically for the user if they enable the property. However, since the property is read pretty late, this is probably not possible, and it would be best to just update the MDK to include the `idea` plugin. So I guess this is fine...

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -491,4 +497,27 @@ public static String replaceTokens(Map<String, ?> tokens, String value) {
...
+
+        if (project.getPlugins().hasPlugin(EclipsePlugin.class)) {
+            project.getTasks().create("copyEclipseResources", CopyEclipseResources.class, task -> {
+                task.dependsOn("eclipse");
+                project.getTasks().named("eclipse").configure($ -> $.doLast($$ -> {
+                    final EclipseModel eclipse = project.getExtensions().findByType(EclipseModel.class);
+                    task.configure(eclipse, project);
+                }));
+            });
```
SizableShrimp: Why does this task need to depend on `eclipse`?

src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
```diff
@@ -93,7 +97,69 @@ protected Map<String, Document> createRunConfiguration(@Nonnull final Project pr
...
             javaDocument.appendChild(rootElement);
         }
-        documents.put(runConfig.getTaskName() + ".launch", javaDocument);
+
+        final String configName = (mc.getGenerateRunFolders().getOrElse(false) ? runConfig.getFolderName() + " - " : "") + runConfig.getTaskName() + ".launch";
+        final boolean copyResources = mc.getCopyIDEResources().getOrElse(false);
+
+        if (mc.getEnableEclipsePrepareRuns().getOrElse(false) || copyResources) {
+            final String launchConfigName = project.getName() + " - " + runConfig.getTaskName() + "Slim";
+            documents.put(".eclipse/configurations/" + launchConfigName + ".launch", javaDocument);
```
SizableShrimp, line 93: This prefixes with a folder while the others do not. Is this proper?
My assumption would be you do it this way as otherwise the launch files go into the root folder of the project (from memory).
Does this mean they will automatically be imported into Eclipse without any user interaction?

src/common/java/net/minecraftforge/gradle/common/util/runs/IntellijRunGenerator.java
```diff
@@ -198,7 +203,13 @@ protected Map<String, Document> createRunConfiguration(@Nonnull final Project pr
                         {
                             gradleTask.setAttribute("name", "Gradle.BeforeRunTask");
                             gradleTask.setAttribute("enabled", "true");
-                            gradleTask.setAttribute("tasks", project.getTasks().getByName("prepare" + Utils.capitalize(runConfig.getTaskName())).getPath());
+                            final List<String> tasks = new ArrayList<>();
+                            final boolean copyResources = mc.getCopyIDEResources().getOrElse(false);
+                            if (mc.getEnableIdeaPrepareRuns().getOrElse(true) || copyResources)
```
SizableShrimp: I feel like I would prefer a `.convention(true)` on the original property (probably constructor of MinecraftExtension) rather than a getOrElse(true) everywhere it is referenced.

src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
```diff
@@ -356,11 +352,14 @@ public final void createRunConfiguration(@Nonnull final MinecraftExtension minec
...
                 minecraft.getRuns().forEach(runConfig -> {
-                    final Map<String, Document> documents = createRunConfiguration(project, runConfig, docBuilder, additionalClientArgs);
+                    final Map<String, Document> documents = createRunConfiguration(minecraft, project, runConfig, docBuilder, additionalClientArgs);
 
                     documents.forEach((fileName, document) -> {
                         final DOMSource source = new DOMSource(document);
-                        final StreamResult result = new StreamResult(new File(runConfigurationsDir, fileName));
+                        final File location = new File(runConfigurationsDir, fileName);
+                        if (!location.getParentFile().exists())
+                            location.getParentFile().mkdirs();
```
SizableShrimp, line 351: Should we throw an exception if `mkdirs()` returns false?

SizableShrimp (changes requested):

src/common/java/net/minecraftforge/gradle/common/util/runs/EclipseRunGenerator.java
Matyrobbrt: Eclipse run configurations are global. The project name prefix makes the launch group know exactly what configurations to launch.
The slim and gradle configurations are in the `.eclipse/configurations` folder because they shouldn't be manually launched and to reduce the files in the root dir. 
Also eclipse scans the entire project for run configurations, so manual interaction is not needed no.

SizableShrimp: Is this folder used by eclipse or something you made up?

Matyrobbrt: No it is not, by default eclipse dumps configurations in the root project dir, but it scans the entire project for configurations.

src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
Matyrobbrt: Because I think that the default folder name should be the project's name.

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
Matyrobbrt: Because the model and / or the information I need from it is `null` when the eclipse task hasn't been run.

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -491,4 +497,27 @@ public static String replaceTokens(Map<String, ?> tokens, String value) {
...
+            project.getTasks().create("copyIdeaResources", CopyIDEAResources.class, task -> task.configure(idea.getModel()));
+        } else {
+            project.getLogger().warn("'idea' plugin is not found! IDEA resources will not be copied.");
+            project.getTasks().create("copyIdeaResources", Task.class); // Create dummy class
+        }
+
+        if (project.getPlugins().hasPlugin(EclipsePlugin.class)) {
+            project.getTasks().create("copyEclipseResources", CopyEclipseResources.class, task -> {
+                task.dependsOn("eclipse");
+                project.getTasks().named("eclipse").configure($ -> $.doLast($$ -> {
```
marchermans: Use proper naming instead of `$`

src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
```diff
@@ -644,6 +648,14 @@ public List<SourceSet> getAllSources() {
         return sources;
     }
 
+    public void folderName(String folderName) {
```
marchermans, line 635: This does not follow the proper naming schema.
Either make this a proper setter, or a property.

Note: this is not something GSON can handle as you have it right now!

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -491,4 +497,27 @@ public static String replaceTokens(Map<String, ?> tokens, String value) {
...
+                    final EclipseModel eclipse = project.getExtensions().findByType(EclipseModel.class);
+                    task.configure(eclipse, project);
+                }));
+            });
+        } else {
+            project.getLogger().warn("'eclipse' plugin is not found! Eclipse resources will not be copied.");
+            project.getTasks().create("copyEclipseResources", Task.class); // Create dummy class
```
marchermans: This should not be needed.
You can easily detect if the plugin or task is loaded where you reference it.
Don't polute the task tree with not needed dummy tasks.

marchermans (changes requested):
See previous comments.

Matyrobbrt:
Suggestions applied.

marchermans (approved):

src/common/java/net/minecraftforge/gradle/common/tasks/ide/CopyIDEAResources.java
```diff
@@ -0,0 +1,51 @@
...
+
+public abstract class CopyIDEAResources extends Copy {
+    public static final String NAME = "copyIdeaResources";
+
+    public void configure(IdeaModel model) {
+        final Path outDir;
+        if (model.getModule().getOutputDir() == null)
+            outDir = model.getProject().getProject().file("out").toPath();
+        else
+            outDir = model.getModule().getOutputDir().toPath();
```
sciwhiz12: Use braces, pls. And add a blank line between this and the `for`.

Additionally, it may be worth reusing this code for `IntellijRunGenerator#getIdeaPathsForSourceset`.

https://github.com/MinecraftForge/ForgeGradle/blob/1fce46ef54e42adb69176edb0e8c3401a707b39c/src/common/java/net/minecraftforge/gradle/common/util/runs/IntellijRunGenerator.java#L250-L255

src/common/java/net/minecraftforge/gradle/common/tasks/ide/CopyIDEAResources.java
```diff
@@ -0,0 +1,51 @@
...
+        for (final SourceSet sourceSet : model.getProject().getProject().getExtensions().getByType(JavaPluginExtension.class).getSourceSets()) {
+            dependsOn(sourceSet.getProcessResourcesTaskName());
+            model.getProject().getProject().getTasks().named(sourceSet.getProcessResourcesTaskName(), ProcessResources.class)
+                    .configure(processResources -> {
+                        final String outName = sourceSet.getName().equals(SourceSet.MAIN_SOURCE_SET_NAME) ? "production" : sourceSet.getName();
```
sciwhiz12: I think we also have a special-casing for this in the IDEA run generator, so we should deduplicate the two into a method in `Utils`.

src/common/java/net/minecraftforge/gradle/common/util/MinecraftExtension.java
```diff
@@ -127,4 +132,9 @@ public void accessTransformer(Object file) {
     }
 
     public abstract ConfigurableFileCollection getSideAnnotationStrippers();
+
+    public abstract Property<Boolean> getEnableEclipsePrepareRuns();
+    public abstract Property<Boolean> getEnableIdeaPrepareRuns();
```
sciwhiz12: Might be better to document these briefly, since the name doesn't imply much.

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -491,4 +498,26 @@ public static String replaceTokens(Map<String, ?> tokens, String value) {
 
         return buf.toString();
     }
+
+    public static void setupIDEResourceCopy(@Nonnull final Project project) {
+        if (project.getPlugins().hasPlugin(IdeaPlugin.class)) {
+            final IdeaPlugin idea = project.getPlugins().getPlugin(IdeaPlugin.class);
+            project.getTasks().create(CopyIDEAResources.NAME, CopyIDEAResources.class, task -> task.configure(idea.getModel()));
+        } else {
+            project.getLogger().warn("'idea' plugin is not found! IDEA resources will not be copied.");
```
sciwhiz12: I'm not fond of logging warning messages when either or both of the plugins are not present. Could we instead have it log a warning when neither of the plugins are found?

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -491,4 +498,26 @@ public static String replaceTokens(Map<String, ?> tokens, String value) {
 
         return buf.toString();
     }
+
+    public static void setupIDEResourceCopy(@Nonnull final Project project) {
+        if (project.getPlugins().hasPlugin(IdeaPlugin.class)) {
+            final IdeaPlugin idea = project.getPlugins().getPlugin(IdeaPlugin.class);
+            project.getTasks().create(CopyIDEAResources.NAME, CopyIDEAResources.class, task -> task.configure(idea.getModel()));
```
sciwhiz12: Is it deliberate that this and the eclipse task eagerly creates\ the task rather than lazily registering them? If yes, add a comment why.

sciwhiz12 (changes requested):

src/common/java/net/minecraftforge/gradle/common/util/MinecraftExtension.java
Matyrobbrt: I will do so on the [wiki](https://github.com/MinecraftForge/ForgeGradle/wiki/Configuration#properties) if / after the PR is merged.

SizableShrimp: Both places would be good.

src/common/java/net/minecraftforge/gradle/common/util/Utils.java
```diff
@@ -500,24 +501,27 @@ public static String replaceTokens(Map<String, ?> tokens, String value) {
...
-                    // The `eclipse` task has been run, the only way the model would be null is if something has gone seriously wrong
-                    task.configure(Objects.requireNonNull(eclipse), project);
-                }));
-            });
-        } else {
-            project.getLogger().warn("'eclipse' plugin is not found! Eclipse resources will not be copied.");
+            final TaskProvider<CopyEclipseResources> taskProvider = project.getTasks().register(CopyEclipseResources.NAME, CopyEclipseResources.class, task -> task.dependsOn("eclipse"));
+            project.getTasks().named("eclipse").configure(eclipseTask -> eclipseTask.doLast(eclTask -> {
+                final EclipseModel eclipse = project.getExtensions().findByType(EclipseModel.class);
+                // The `eclipse` task has been run, the only way the model would be null is if something has gone seriously wrong
```
sciwhiz12, line 511: I wonder if we should guard against this, just in case.

sciwhiz12 (approved):
Code looks fine to me at first glance, though I don't have the capacity right now to test this. :shipit:

SizableShrimp:
Completed in [ForgeGradle 6](https://github.com/MinecraftForge/ForgeGradle/tree/FG_6.0).

PR #872, opened 2022-07-07
Support MC Only workflows

marchermans:
This fixes a case where FG is run in MC Only mode (Without Forge) while it expects its inject target to exists.

Current the forge variant creates the inject target automatically, but the MC only variant does not.
This causes the MCP injection to fail with a file not found exception since neither the target nor its directories exists.

This fixes this.


SizableShrimp:
Fixed by fbaee8c6d4a864d0e84a85708d732639be6cea7f.

PR #876, opened 2022-07-21
Rudimentary support for including subprojects with Jar-in-Jar

EdwinMindcraft:
This adds rudimentary support for including sub projects into a jar-in-jar build by replacing calls referring to `ExternalModuleDependency` with a call to `Dependency` where possible.


marchermans (changes requested):
No this is not the way to solve this.
You need proper handling of resolution time to make this work, i have this already setup internally but some other things got in the way of making this public.

If you want to help make this work please contact me on discord.

PR #878, opened 2022-08-01
Allow run configurations to build all projects

amadornes:
The current default pre-run behavior in IntelliJ is `Make` (or "Build" inside the IDE), which only builds the source set targetted by the run configuration and none of its dependencies. This is incredibly inconvenient when working with multiple source sets or projects, as changes won't always auto-build when starting a run configuration. It can be changed manually, but the moment run configurations are re-generated it will be overridden again. As far as I can tell, this is not relevant in Eclipse.

This PR adds an extra option to run configurations that allows the pre-run behavior to be set to `MakeProject` (or "Build Project" inside the IDE), which builds all the modules and source sets.


marchermans:
:shipit: Was requested by me.

PR #881, opened 2022-08-10
Implement a fix for a crash that can occur when JiJ is enabled.

marchermans:
When #879 was merged a corner case opened up with the way FG registers the reobfuscation tasks for JiJ.
This has been fixed with this PR by using a immediate lazy resolve when the JiJ tasks get added.

Additionally cleans up the maven publication from any vanilla artifacts.
Adds a logging line to keep track of the automatically added Reobf tasks for JiJ to make future testing and diagnosing easier.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -77,6 +77,8 @@ import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 
 public class UserDevPlugin implements Plugin<Project> {
+    private static final org.slf4j.Logger LOGGER = org.slf4j.LoggerFactory.getLogger(UserDevPlugin.class);
+
     public static final String JAR_JAR_TASK_NAME = "jarJar";
     public static final String JAR_JAR_GROUP = "jarjar";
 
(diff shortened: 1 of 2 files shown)


src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -77,6 +77,8 @@
 import javax.annotation.Nonnull;
 
 public class UserDevPlugin implements Plugin<Project> {
+    private static final org.slf4j.Logger LOGGER = org.slf4j.LoggerFactory.getLogger(UserDevPlugin.class);
```
SizableShrimp: Imports please.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -329,11 +331,12 @@ public void apply(@Nonnull Project project) {
             Utils.createRunConfigTasks(extension, extractNatives, downloadAssets, createSrgToMcp);
         });
 
-        project.afterEvaluate(projectAfter -> projectAfter.getTasks().withType(JarJar.class).configureEach(jarJar -> {
+        project.getTasks().withType(JarJar.class).configureEach(jarJar -> {
             if (jarJar.isEnabled()) {
+                LOGGER.info("Creating reobfuscation task for JarJar task: " + jarJar.getName());
```
SizableShrimp: Does this log print every time? Or is INFO hidden by default. Also this should use {} pattern.

SizableShrimp (changes requested):

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
marchermans: Fixed used the wrong logger.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
marchermans: Info is hidden by default.
You need to enable it either with `--info` or by setting the property: `org.gradle.logging.level` to `info` or lower.

SizableShrimp (approved):

PR #884, opened 2022-08-18
Properly queue creation of reobf tasks for JarJar tasks

### LINKED ISSUE #883: Reobf tasks for JarJar tasks are not created
Related to #881. Seems to be an issue with using `configureEach` at plugin application rather than at `project.afterEvaluate`.

Original Discord Message

https://discord.com/channels/313125603924639766/801175194298744902/1009829595803041813

Jonathing:
Fixes #883.

Additionally, I have moved the call to `whenTaskAdded` above the call to `configureJarJarTask` so that the primary JarJar task that is created by ForgeGradle can be effected by this as well.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
@@ -219,6 +219,13 @@ public class UserDevPlugin implements Plugin<Project> {
             updateMappings.configure(task -> task.dependsOn(extractMappedNew));
         }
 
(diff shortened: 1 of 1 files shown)


Jonathing:
Closing this as an invalid solution to #883 (and by extension #881). `configureEach` does in fact work but is negated by how JarJar tasks on creation are disabled by default until the end user in their own `build.gradle` file enables it manually or includes a JarJar dependency. I shall leave it up to our friends @OrionDevelopment and @SizableShrimp 🦐. Here's a screenshot basically summarizing what we talked about in #gradle today.

Related Discord Message

https://discord.com/channels/313125603924639766/801175194298744902/1009874822177751101

Farewell, chat.

PR #889, opened 2022-10-18
Support customization of the java launcher for run tasks

TheMrMilchmann:
Since the run tasks are created after the project has been evaluated, "setting" the javaLauncher makes it extremely difficult to customize the launcher. This becomes apparent when using a more recent Java version for the build process (which can be useful to evade rare compiler bugs) while still targeting old bytecode. To enable users to configure the launcher more easily, the task is updated to set the *default* value for the launcher. Setting the value via any of the Provider#set methods from user code now will work as expected and the current default behavior is preserved.


marchermans (approved):

SizableShrimp (approved):
Confirmed to work with this test case:
```groovy
tasks.configureEach {
    if (it.name == 'runClient') {
        javaLauncher = javaToolchains.launcherFor {
            languageVersion = JavaLanguageVersion.of(19)
        }
    }
}
```

Without this PR, there is no error and it silently fails to use the declared version, instead falling back to the project declared one. (Accessing the generated run config task is annoying but still possible as shown in the example.)

PR #896, opened 2023-01-07
Add ability to mark client config as authenticated using DevLogin

ApexModder:
This PR brings the ability to mark client run configs as being authenticated. These configs use Covers1624 DevLogin tool to login & authenticate against Mojangs account servers on launch.

See [here](https://github.com/covers1624/DevLogin#devlogin) for more details on DevLogin.

---

I have added Covers maven repo to be able to pull in the DevLogin jar, and have needed to change the main class for the client run config generated to `net.covers1624.devlogin.DevLogin` aswell as append a new program arg `--launch_target <original main>`.
[_Following the [wiki](https://github.com/covers1624/DevLogin/wiki/ForgeGradle-Setup) as a setup guide_]

---

To mark a config as being authenticated the user must specify `authenticated` in their gradle file like so
```groovy
minecraft {
	runs {
		client {
			// ...
			authenticated()
			// ...
		}
	}
}
```

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
@@ -66,6 +66,7 @@ public class RunConfig extends GroovyObjectSupport implements Serializable {
     private List<String> args, jvmArgs;
     private boolean forceExit = true;
     private Boolean client; // so we can have it null
+    private Boolean authenticated;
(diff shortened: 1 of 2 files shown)


marchermans:
I am sorry to say, but this won't be accepted.
We have a policy to not integrate external uncontrolled projects into ours.

Besides this, we hope to introduce extension support on runs in FG6, but that is still a bit out.

PR #898, opened 2023-01-10
Update artifactural and make FG's manual maven respect content filtering

pupnewfster:
Opening as a draft PR for now as it requires https://github.com/MinecraftForge/Artifactural/pull/7, but this PR in combination with that Artifactural PR will make the manual maven downloading FG does respect declared content filtering of the different repositories.

When doing some initial testing against Mekanism that declares a lot of repositories along with content filters for them, I was seeing rough timing changes on reimporting the project of about 3 minutes without this change to about 20 seconds with this change.


Matyrobbrt (approved):
Tested by explicitly excluding the group of a deobfuscated mod dependency and seeing the build fail.
On a similar note, we should probably make sure to filter the contents of the FG-added repositories but that's a task for another PR and another Artifactural update.

PR #907, opened 2023-02-15
Introduce `ideaFolder` run config property

andriihorpenko:
## Introduction

This PR introduces `ideaFolder` run config property, which adds a `folderName` run attribute to IntelliJ-based runs. This allows grouping runs into folders, which helps organizing project-related runs into their own groups.

## Examples

Suppose you have a multi-project build with three subprojects: `mod-a`, `mod-b` and `mod-c`. All of them are mods, which generate their own runs.

Currently, all runs are placed on the same level, which makes runs list bloated. With this PR, adding `ideaFolder(project.name)` inside a run config will place runs under a folder named after subproject's name.

Before             |  After
:-------------------------:|:-------------------------:
image  |  image

## Implementation notice
1. As IntelliJ permits folders with empty name, `folderName` attribute should be present only when a folder is used.
2. `ideaFolder` follows the same code style as `ideaModule`.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/RunConfig.java
@@ -59,7 +59,7 @@ public class RunConfig extends GroovyObjectSupport implements Serializable {
 
     private Boolean singleInstance = null;
 
-    private String taskName, main, ideaModule, workDir;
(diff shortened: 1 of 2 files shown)


SizableShrimp:
Completed in [ForgeGradle 6](https://github.com/MinecraftForge/ForgeGradle/tree/FG_6.0) with the IDE run configuration overhaul (use the new RunConfig property `folderName` instead).

PR #911, opened 2023-03-21
Add content filtering to FG-added repositories

Matyrobbrt:
This PR adds content filtering to the FG added repositories:
- `MinecraftRepo`
- `MCPRepo`
- `MinecraftUserRepo`
- `DeobfuscatingRepo`

The addition of filtering will slightly speed up small buildscripts and may provide a noticeable difference for bigger ones. Additionally, Gradle will only attempt to request dependencies which match the filter, meaning that logging of dependency errors will be more accurate.

Proper testing should be done against a mod with access transformers and mod dependencies (which use `fg.deobf`) in a clean workspace.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java
@@ -112,9 +112,13 @@ public abstract class BaseRepo implements ArtifactProvider<ArtifactIdentifier> {
             return this;
         }
 
-        public void attach(Project project) {
+        public GradleRepositoryAdapter attach(Project project) {
+            return attach(project, "bundeled_repo");
+        }
+
+        public GradleRepositoryAdapter attach(Project project, String cacheName) {
             int random = new Random().nextInt();
-            File cache = Utils.getCache(project, "bundeled_repo");
+            File cache = Utils.getCache(project, cacheName);
             // Java 8's compiler doesn't allow the lambda to be a method reference, but Java 16 allows it
             // noinspection Convert2MethodRef to prevent IDEA from warning us about it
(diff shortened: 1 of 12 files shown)


src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java
```diff
@@ -15,13 +15,16 @@
 import net.minecraftforge.artifactural.gradle.GradleRepositoryAdapter;
 
 import org.gradle.api.Project;
+import org.gradle.api.artifacts.repositories.MavenArtifactRepository;
```
pupnewfster: The `MavenArtifactRepository` import being added seems to be unused

src/common/java/net/minecraftforge/gradle/common/util/BaseRepo.java
```diff
@@ -15,13 +15,16 @@
...
+import org.gradle.api.artifacts.repositories.RepositoryContentDescriptor;
 import org.gradle.api.logging.Logger;
 
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
+import java.util.function.Consumer;
```
pupnewfster: The `Consumer` import being added seems to be unused

pupnewfster (approved):
Seems to work nicely. I tested it against Mekanism and it sped up the reimport by a few more seconds and also allows for usage of Gradle's [exclusiveContent](https://docs.gradle.org/current/userguide/declaring_repositories.html#declaring_content_exclusively_found_in_one_repository) now by providing access to the deobf repo. Which allows [this mess](https://github.com/mekanism/Mekanism/blob/f46383f2a017d9abe617e3b5479a056de843307f/build.gradle#L343-L411) to be cleaned up to [something much more readable](https://gist.github.com/pupnewfster/6c21401789ca6d74f9892be8c1c505c9).

marchermans (approved):

PR #913, opened 2023-03-23
Implement a Legacy Extension

TheCurle:
## What?

This is for the Retrogradle Initiative. See https://gist.github.com/TheCurle/fe7ad3ede188cbdd15c235cc75d52d4a for more details on what that means.

In short, we're making old versions of Minecraft work under newer toolchains, by porting them to MCPConfig and implementing some changes into an opt-in extension in ForgeGradle.

This Legacy Extension is added for both UserdevPlugin and PatcherPlugin, so that it applies to both userdev and forgedev.  
It hooks into the generation of run configs to apply some changes:

It only activates (expresses its features) if it is manually activated; this is done by having an 
```
ext {
    MC_VERSION = '1.11.2'
}
```
or
```
legacy {
    <fix> = true
    ie.
    fixClasspath = true
}
```
entry in the buildscript at the root level. Any version that is registered as being lower than '1.13' by ForgeGradle internally is deemed a trigger for the LegacyExtension.

The fixes implemented are:

* Redirecting the separate classes,resources to a single jar file
* Extracting mappings and putting the path under a system property.

## Why?

These changes are done to account for the fact that FML1 (1.12 and below) was designed to work with the functionality of ForgeGradle 2.3 and below.

Namely, ForgeGradle 2.3 merges classes and resources into a single folder, thus FML only consumes a single folder when looking for classes and resources.

ForgeGradle 3 cleaned this up to separate them, but FML only consuming a single path means that it only finds the classes, leaving all mod resources out of the runtime.

When activated, the LegacyExtension hijacks the run configs to change the classes output to the `jar` task *input*, essentially merging all the compiled classes+resources pre-obfuscation and providing them all to FML.

Additionally, FML1 does runtime remapping of certain things; it is fed the mapping data by ForgeGradle 2, which was removed in FG3 as a cleanup. This is required to make dependencies with coremods work, so the LegacyExtension puts the mappings file into a SystemProperty that the LegacyDev system can make use of at runtime.

## Maintenance?

All of the features added by RetroGradle are implemented in a single location; `net.minecraftforge.gradle.common.legacy.LegacyExtension`. The extension is added to the UserDev and Patcher plugins, and is called once from Utils.createRunConfigTasks.

This makes a maintenance surface of 3 lines for the LegacyExtension.

Additionally, the rationale and functionality of the LegacyExtension is documented thoroughly in code, so that someone trying to fix or reimplement these changes in the future is able to understand what and why is going on.


src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,141 @@
...
+                // mappings are extracted to <root>/build/<taskName> by default
+                t.getOutput().convention(project.getLayout().getBuildDirectory().dir(t.getName()));
+            });
+            final File csvDir = extractMappingsTask.getOutput().get().getAsFile();
+
+            // attach the csvDir location property to each run configuration
+            minecraft.getRuns().configureEach(run -> run.property("net.minecraftforge.gradle.GradleStart.csvDir", csvDir));
+
+            // execute extractMappings before each run task
+            project.getTasks().getByName("prepareRuns").dependsOn(extractMappingsTask);
```
Matyrobbrt: Any reason this uses eager configuration method instead of `named` for example?

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    public LegacyExtension(Project project) {
+        Provider<Boolean> isLegacy = project.provider(() -> {
+            final MinecraftVersion version = MinecraftVersion.from((String) (project.getParent() != null ? project.getParent() : project).getExtensions().getExtraProperties().get("MC_VERSION"));
```
Matyrobbrt: This is really annoying for projects with subprojects. Please check if the current project has the property and if it doesn't only then query it from the root project.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
```diff
@@ -5,26 +5,14 @@
...
-import net.minecraftforge.gradle.common.tasks.ExtractMCPData;
-import net.minecraftforge.gradle.common.tasks.ExtractNatives;
-import net.minecraftforge.gradle.common.tasks.ExtractRangeMap;
-import net.minecraftforge.gradle.common.util.BaseRepo;
-import net.minecraftforge.gradle.common.util.EnvironmentChecks;
-import net.minecraftforge.gradle.common.util.MinecraftRepo;
-import net.minecraftforge.gradle.common.util.MojangLicenseHelper;
-import net.minecraftforge.gradle.common.util.Utils;
-import net.minecraftforge.gradle.common.util.VersionJson;
+import net.minecraftforge.gradle.common.tasks.*;
```
Matyrobbrt: No star imports

Matyrobbrt (changes requested):

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,141 @@
...
+     *  - Dependency AccessTransformers / Coremods;
+     *      FG2 GradleStart exposes a <code>net.minecraftforge.gradle.GradleStart.csvDir</code> property
+     *      that points to a directory containing CSV mappings files. This is used by LegacyDev to remap dependencies' AT modifiers.
+     *      We replicate the FG2 behavior by extracting the mappings to a folder in the build directory
+     *      and setting the property to point to it.
+     *
+     * This is called from Utils.
+     *
+     * In other words, it's a containment zone for version-specific hacks.
+     * For issues you think are caused by this function, contact Curle or any other Retrogradle maintainer.
```
marchermans: Update documentation to reflect that htis is not a RG project but now part of the official FG

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * extractMappings;
+     * FG2 GradleStart exposes a <code>net.minecraftforge.gradle.GradleStart.csvDir</code> property
+     * that points to a directory containing CSV mappings files. This is used by LegacyDev to remap dependencies' AT modifiers.
+     * We replicate the FG2 behavior by extracting the mappings to a folder in the build directory
+     * and setting the property to point to it.
+     * <p>
+     * Takes a boolean - true for apply fix, false for no fix.
+     */
+    public abstract Property<Boolean> getExtractMappings();
```
marchermans, line 146: Remove the public properties.
The PR for this states that this object is not supposed to be a part of the public API.

If this is part of the public API, then we should discuss that.

marchermans (changes requested):
At the moment the documentation of this still references RetroGradle.
Additionally the API surface currently extends into the legacy artifact which I am opposed to. Based on the specification discussed I see no reason for those properties to be public, or exist at all.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
TheCurle: This only runs in afterEvaluate; at that point, all tasks are already present, right? This shouldn't have a performance impact if i understand lazy evaluation correctly.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
TheCurle: The PR states that the maintenance surface is the 2 lines registering this extension + the 1 line calling runRetrogradleFixes once everything is set up.

The fixes need to be opt-in, which is why the ```legacy {``` block exists.

If there's some other way we can evaluate the `isLegacy` provider in the constructor using files that are only created after project evaluation is finalised, then i'm all ears. To my knowledge, this is the cleanest way.

src/userdev/java/net/minecraftforge/gradle/userdev/UserDevPlugin.java
TheCurle: Whoops.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
TheCurle: On the other hand, the things implemented here are done purely to aid the RG project, someone maintaining this for the benefit of FG only runs the risk of breaking features and fixes required in RG versions.

I'm named there as a maintainer of both.

We should figure out whether this should *stay* as a RG feature in the FG project.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
Matyrobbrt: That is wrong. A task may be registered but it does not exist until it is needed by either a task invocation, task dependency or when it is queried.
To quote the [docs](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html):
> To avoid creating and configuring tasks, we say that a task is "registered" but not **created**. When a task is in this state, it is known to the build, it can be configured, and references to it can be passed around, but the task object itself has not actually been created, and none of its configuration actions have been executed.
It will remain in this state until something in the build needs the instantiated task object (for instance if the task is executed on the command line or the task is a dependency of a task executed on the command line). If the task object is never needed, then the task will remain in the registered state, and the cost of creating and configuring the task will be avoided.

TheCurle: Okay, I see.

I'll fix that, then.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,141 @@
...
+ * }</code>
+ *
+ * It allows for configuring RetroGradle patches and fixes, which are otherwise set by version.
+ * Each fix is documented in this class and in the application function.
+ * {@link LegacyExtension#runRetrogradleFixes}
+ *
+ * @author Curle
+ */
+public abstract class LegacyExtension extends GroovyObjectSupport {
+    private static final Logger LOGGER = LogManager.getLogger(LegacyExtension.class);
```
Matyrobbrt: You should use the project's logger.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.File;
+
+
+/**
+ * Provides an extension block named "legacy".
+ *
+ * <code>
+ * legacy {
+ *     flag = set;
+ * }</code>
```
Matyrobbrt: Abstract documentation for the purpose of having documentation... Why exactly does this exist?

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * It allows for configuring RetroGradle patches and fixes, which are otherwise set by version.
+ * Each fix is documented in this class and in the application function.
+ * {@link LegacyExtension#runRetrogradleFixes}
```
Matyrobbrt: Should be a `@see` instead.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,151 @@
...
+     *      FG2 Userdev puts all classes and resources into a single jar file for FML to consume.
+     *      FG3+ puts classes and resources into separate folders, which breaks on older versions.
+     *      We replicate the FG2 behavior by replacing these folders by the jar artifact on the runtime classpath.
+     *  - Dependency AccessTransformers / Coremods;
+     *      FG2 GradleStart exposes a <code>net.minecraftforge.gradle.GradleStart.csvDir</code> property
+     *      that points to a directory containing CSV mappings files. This is used by LegacyDev to remap dependencies' AT modifiers.
+     *      We replicate the FG2 behavior by extracting the mappings to a folder in the build directory
+     *      and setting the property to point to it.
+     *
+     * This is called from Utils.
```
Matyrobbrt: Use a link reference.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,151 @@
...
+     * For issues you think are caused by this function, contact Curle or any other Retrogradle maintainer.
+     */
+    public static void runRetrogradleFixes(Project project) {
+        final LegacyExtension config = (LegacyExtension) project.getExtensions().getByName(LegacyExtension.EXTENSION_NAME);
+        // Get the Userdev extension for the run configs
+        final MinecraftExtension minecraft = project.getExtensions().getByType(MinecraftExtension.class);
+        final boolean shouldFixClasspath = config.getFixClasspath().get();
+        final boolean shouldExtractMappings = config.getExtractMappings().get();
+
+        if(shouldFixClasspath) {
```
Matyrobbrt:
```suggestion
        if (shouldFixClasspath) {
```

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
```diff
@@ -0,0 +1,151 @@
...
+            minecraft.getRuns().configureEach(run -> run.property("net.minecraftforge.gradle.GradleStart.csvDir", csvDir));
+
+            // execute extractMappings before each run task
+            project.getTasks().named("prepareRuns").configure(t -> t.dependsOn(extractMappingsTask));
+        }
+    }
+
+    public LegacyExtension(Project project) {
+        Provider<Boolean> isLegacy = project.provider(() -> {
+            String ver = "";
```
Matyrobbrt, line 104: My understanding is that this detection should only happen in forgedev so please put it behind a flag in the constructor.

src/patcher/java/net/minecraftforge/gradle/patcher/PatcherPlugin.java
```diff
@@ -85,6 +86,7 @@ public void apply(@Nonnull Project project) {
         EnvironmentChecks.checkEnvironment(project);
 
         final PatcherExtension extension = project.getExtensions().create(PatcherExtension.class, PatcherExtension.EXTENSION_NAME, PatcherExtension.class, project);
+        project.getExtensions().create(LegacyExtension.EXTENSION_NAME, LegacyExtension.class);
```
Matyrobbrt: You should probably apply this after the channel providers just for consistency with userdev.

Matyrobbrt (changes requested):

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
TheCurle: The extension property is merely a shortcut to enabling all fixes (as it is already set in forgedev).

Explicitly enabling the property via a legacy { block is also valid, as is using ext { MC_VERSION in userdev.

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
TheCurle: Some history: this was written before everything else in the LegacyExtension, as a guide and reminder so that i knew what to search for when building this.

It is not needed any more, so it has been removed.

Matyrobbrt (approved):

src/common/java/net/minecraftforge/gradle/common/legacy/LegacyExtension.java
marchermans: Do you expect these properti 's tok be set by a user?

TheCurle: Yes. It is opt-in.

marchermans: Let me state this differently, is there any **legacy** version of Minecraft for which this extension is designed which has either one set to false?

TheCurle: Both fixes may cause issues with some setups (such as those mods that have explicitly fixed the resources issue by changing the build.gradle with:

```
runs {
        client {
            workingDirectory project.file('run')
            property 'forge.logging.markers', ''
            property 'forge.logging.console.level', 'debug'
            property 'forge.logging.noansi', "false"

            ideaModule "${project.name}.main"

            environment 'MOD_CLASSES', "${project.file("out/production//resources").canonicalPath};${project.file("out/production//classes").canonicalPath}"
        }
    }
```
or 
```
sourceSets {
    main {
        output.resourcesDir = file('build/combined')
        java.outputDir = file('build/combined')
    }
}
```

or such.
    
Additionally, the coremods fix may crash on setups that do not expect things to be externally remapped at setup time like this.

Therefore, both must be ***opt-in by the modder***, and not set automatically.

marchermans (approved):
We will add this under the stipulation that none of it will become part of the public api. If the RetroGradle team is not available to fix issues with it, the n it can and will be removed at any point in time.

PR #918, opened 2023-05-14
Fix MavenArtifactDownloader deadlock.

### LINKED ISSUE #886: MavenArtifactDownloader deadlocks when used from mapping channel provider.
When `MavenArtifactDownloader#_download` is called with an artifact that is currently downloading, it [waits on the other thread](https://github.com/MinecraftForge/ForgeGradle/blob/FG_5.0/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java#L128) to complete the download. The problem is, that this is done *inside* the `synchronized` block.

This causes the waiting thread to hold the monitor of `ACTIVE_DOWNLOADS` indefinitely and blocks every other request for `MavenArtifactDownloader`. The problem arises if the artifact that currently is being downloaded is a generated artifact that relies on `MavenArtifactDownloader` to generate.

### When does the issue occur in practise?

If a project has `fg.deobf(...)` dependencies, they'll be resolved together with the `minecraft` dependency. This causes both [Deobfuscator#findMapping](https://github.com/MinecraftForge/ForgeGradle/blob/FG_5.0/src/userdev/java/net/minecraftforge/gradle/userdev/util/Deobfuscator.java#L198-L207) and [MinecraftUserRepo#findMapping](https://github.com/MinecraftForge/ForgeGradle/blob/FG_5.0/src/userdev/java/net/minecraftforge/gradle/userdev/MinecraftUserRepo.java#L473-L493) to be called in parallel, where both of them download the same mappings artifact through `MavenArtifactDownloader`.

The first call to `MavenArtifactDownloader.download` will start a mapping lookup through [MCPRepo](https://github.com/MinecraftForge/ForgeGradle/blob/FG_5.0/src/mcp/java/net/minecraftforge/gradle/mcp/MCPRepo.java) which calls the matching channel provider. The second call to `MavenArtifactDownloader.download` then waits for the download to complete which completely locks `MavenArtifactDownloader` for **any** artifact.

If the channel provider then uses `MavenArtifactDownloader` in its [getMappingsFile](https://github.com/MinecraftForge/ForgeGradle/blob/FG_5.0/src/mcp/java/net/minecraftforge/gradle/mcp/ChannelProvider.java#L55) method (which both builtin channel providers do), this can cause a deadlock.

As `MavenArtifactDownloader#download` calls happen very early in the builtin channel providers, it is unlikely, yet still possible for a deadlock to occur. Other 3rd party channel providers however, might call `MavenArtifactDownloader#download` later and do other stuff before, so they can become more prone for deadlocks.

noeppi-noeppi:
This fixes #886 by releasing the monitor of `ACTIVE_DOWNLOADS` (through `Object#wait`) when waiting on a concurrent download. Whenever a download finishes, all waiting threads will be notified and check whether the download they are currently waiting on has completed. If that is the case, the result is returned. If not, the thread invokes `Object#wait` again and waits for the next download to complete.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
@@ -112,8 +112,8 @@ public class MavenArtifactDownloader {
                 try {
                     project.getLogger().info("Waiting for download of {} on other thread", artifact);
                     while (!activeDownload.isDone()) {
-                        // Release the monitor of ACTIVE_DOWNLOADS while waiting on the download
-                        // When a new download finishes, we'll get notified and check again, whether the download is complete.
+                        // Release the monitor of ACTIVE_DOWNLOADS while waiting on the download;
+                        // when a new download finishes, we'll get notified and we can check whether the download is complete again. 
                         ACTIVE_DOWNLOADS.wait();
                     }
                     return activeDownload.get();
--- a/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
@@ -111,6 +111,11 @@ public class MavenArtifactDownloader {
                 // Some other thread is already working downloading this exact artifact, wait for it to finish
                 try {
                     project.getLogger().info("Waiting for download of {} on other thread", artifact);
+                    while (!activeDownload.isDone()) {
+                        // Release the monitor of ACTIVE_DOWNLOADS while waiting on the download
+                        // When a new download finishes, we'll get notified and check again, whether the download is complete.
+                        ACTIVE_DOWNLOADS.wait();
+                    }
                     return activeDownload.get();
                 } catch (InterruptedException e) {
                     throw new RuntimeException(e);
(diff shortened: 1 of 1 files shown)


src/common/java/net/minecraftforge/gradle/common/util/MavenArtifactDownloader.java
```diff
@@ -111,6 +111,11 @@ private static File _download(Project project, String artifact, boolean changing
                 // Some other thread is already working downloading this exact artifact, wait for it to finish
                 try {
                     project.getLogger().info("Waiting for download of {} on other thread", artifact);
+                    while (!activeDownload.isDone()) {
+                        // Release the monitor of ACTIVE_DOWNLOADS while waiting on the download
+                        // When a new download finishes, we'll get notified and check again, whether the download is complete.
```
Matyrobbrt:
```suggestion
                        // Release the monitor of ACTIVE_DOWNLOADS while waiting on the download;
                        // when a new download finishes, we'll get notified and we can check whether the download is complete again. 
```

Matyrobbrt (changes requested):

Matyrobbrt (approved):
Confirmed both the bug and that this fix is working through the example mod provided by @noeppi-noeppi in [Discord](https://discord.com/channels/313125603924639766/801175194298744902/1107596598344687727) (https://github.com/noeppi-noeppi/forgegradle886example). Thank you for your contribution!

PR #923, opened 2023-06-20
Support Gradle caching for some parts of FG

PaintNinja:
This PR adds support for Gradle caching in some parts of FG, mainly focused towards userdev AT applying.

To test this PR:
1. Publish this to local maven and use it in the MDK
2. Enable caching in the MDK's gradle.properties with `org.gradle.caching=true`
3. Add an AT the MDK and build the mod
4. Change the AT and build it again
5. Change the AT back to exactly what it was in step 3 and build again

To compare performance with and without this PR:
- For the PR: use the steps above
- Without the PR: use a fresh copy of the MDK for step 1, skip step 2 and then do steps 3, 4 and 5 above.

Known limitations:
- The AccessTransformerFunction creates a temporary AT file which prevents caching in some cases
- Due to the relative path sensitivity of inputs including an input for the user-supplied AT, it can't fully cache across different mod projects using the same mappings, AT and Forge version, *however* the main recompilation stage should still use the cache

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/tasks/JarExec.java
+++ b/src/common/java/net/minecraftforge/gradle/common/tasks/JarExec.java
@@ -20,6 +20,7 @@ import org.gradle.api.plugins.JavaPluginExtension;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.CacheableTask;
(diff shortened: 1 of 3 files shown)


PaintNinja:
Baseline (without this PR, caching not enabled): https://scans.gradle.com/s/oakl6gsps6iti (1m 7s)
With this PR (when getting a cache hit): https://scans.gradle.com/s/p2jeystmnwjlq (33s)

PR #927, opened 2023-06-29
Update Constants.java

jmdane:
Change minecraft assets URL to use https instead of http.
The url http://resources.download.minecraft.net no longer accepts non-https requests and the getAssets tasks fail.


LexManos:
https://github.com/MinecraftForge/ForgeGradle/commit/2b067d7be75903db7da8785e5486fcece91b2d5e

PR #928, opened 2023-06-30
Make lazy tokens lazier

SquidDev:
- Defer reading the user-supplied classpath until needed.
 - Pass around `FileCollection`s instead of `Set<File>`s, so we don't need to resolve the contents until actually needed.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
@@ -21,7 +21,6 @@ import org.gradle.work.DisableCachingByDefault;
 
 import java.io.File;
 import java.util.List;
-import java.util.Set;
 
 @DisableCachingByDefault(because = "IDE runs should always be regenerated")
 abstract class GenIDERunsTask extends DefaultTask {
@@ -41,11 +40,9 @@ abstract class GenIDERunsTask extends DefaultTask {
         MinecraftExtension minecraft = this.getMinecraftExtension().get();
         Project project = this.getProject();
         List<String> additionalClientArgs = this.getAdditionalClientArgs().get();
-        Set<File> minecraftArtifacts = this.getMinecraftArtifacts().getFiles();
-        Set<File> runtimeClasspathArtifacts = this.getRuntimeClasspathArtifacts().getFiles();
 
         runConfigGenerator.createRunConfiguration(minecraft, runConfigurationsDir, project,
-                additionalClientArgs, minecraftArtifacts, runtimeClasspathArtifacts);
+                additionalClientArgs, getMinecraftArtifacts(), getMinecraftArtifacts());
     }
 
     @Internal
--- a/src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
@@ -42,7 +42,7 @@ abstract class GenIDERunsTask extends DefaultTask {
         List<String> additionalClientArgs = this.getAdditionalClientArgs().get();
 
         runConfigGenerator.createRunConfiguration(minecraft, runConfigurationsDir, project,
-                additionalClientArgs, getMinecraftArtifacts(), getRuntimeClasspathArtifacts());
+                additionalClientArgs, this.getMinecraftArtifacts(), this.getRuntimeClasspathArtifacts());
     }
 
     @Internal
--- a/src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
@@ -42,7 +42,7 @@ abstract class GenIDERunsTask extends DefaultTask {
         List<String> additionalClientArgs = this.getAdditionalClientArgs().get();
 
         runConfigGenerator.createRunConfiguration(minecraft, runConfigurationsDir, project,
-                additionalClientArgs, getMinecraftArtifacts(), getMinecraftArtifacts());
+                additionalClientArgs, getMinecraftArtifacts(), getRuntimeClasspathArtifacts());
     }
 
     @Internal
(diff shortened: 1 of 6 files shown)


marchermans:
I support this fully. But for documentation reasons: why is this change needed?; Other then to comply with gradles examples that is....

SquidDev:
I've been fiddling with a couple of alternative approaches to #924, as from conversations on Discord, people weren't very happy with that PR.

I ended up having a few issues where we end up trying to evaluate tokens when configuring a task, which causes deobf dependencies to be fetched, which in turn causes all sorts of other problems, and so it felt worth fixing that first before going any further.

src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
```diff
@@ -41,11 +40,9 @@ public void run() {
         MinecraftExtension minecraft = this.getMinecraftExtension().get();
         Project project = this.getProject();
         List<String> additionalClientArgs = this.getAdditionalClientArgs().get();
-        Set<File> minecraftArtifacts = this.getMinecraftArtifacts().getFiles();
-        Set<File> runtimeClasspathArtifacts = this.getRuntimeClasspathArtifacts().getFiles();
 
         runConfigGenerator.createRunConfiguration(minecraft, runConfigurationsDir, project,
-                additionalClientArgs, minecraftArtifacts, runtimeClasspathArtifacts);
+                additionalClientArgs, getMinecraftArtifacts(), getMinecraftArtifacts());
```
XFactHD: This doesn't look right. Shouldn't it be `this.getMinecraftArtifacts(), this.getRuntimeClasspathArtifacts()` instead of `this.getMinecraftArtifacts(), this.getMinecraftArtifacts()`?

SquidDev: Good spot, thanks!

src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
```diff
@@ -181,6 +174,15 @@ protected static Map<String, Supplier<String>> configureTokensLazy(final Project
         return tokens;
     }
 
+    private static BiFunction<String, Supplier<String>, Supplier<String>> makeClasspathToken(FileCollection classpath) {
+        return (key, supplier) -> Suppliers.memoize(() -> {
+            String resolvedClasspath = getResolvedClasspath(classpath.getFiles());
+            if (supplier == null) return resolvedClasspath;
+            String oldCp = supplier.get();
+            return oldCp == null || oldCp.isEmpty() ? resolvedClasspath : String.join(File.pathSeparator, oldCp, resolvedClasspath);
```
SizableShrimp: This should probably be improved to be an optional merge if either classpath string is empty or null to return the one that is non-null. I don't think our file collection will ever be empty, but it is best to safeguard anyways. And if both are null or empty perhaps return an empty string or null string depending on compute()'s contract?

src/common/java/net/minecraftforge/gradle/common/util/runs/GenIDERunsTask.java
```diff
@@ -41,11 +40,9 @@ public void run() {
         MinecraftExtension minecraft = this.getMinecraftExtension().get();
         Project project = this.getProject();
         List<String> additionalClientArgs = this.getAdditionalClientArgs().get();
-        Set<File> minecraftArtifacts = this.getMinecraftArtifacts().getFiles();
-        Set<File> runtimeClasspathArtifacts = this.getRuntimeClasspathArtifacts().getFiles();
 
         runConfigGenerator.createRunConfiguration(minecraft, runConfigurationsDir, project,
-                additionalClientArgs, minecraftArtifacts, runtimeClasspathArtifacts);
+                additionalClientArgs, getMinecraftArtifacts(), getRuntimeClasspathArtifacts());
```
SizableShrimp: Consistency: The other class has a `this.` prefix you kept, while you removed it here. Probably add it.

SizableShrimp (changes requested):
This looks fine. My comments are minor. I will likely test it in a few days when I am able to.

src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
SquidDev: Yep, that's a good point. I think we want to avoid returning `null` here, as `classpathFileWriter` assumes the returned string is non-null.

src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java
```diff
@@ -179,7 +179,9 @@ private static BiFunction<String, Supplier<String>, Supplier<String>> makeClassp
             String resolvedClasspath = getResolvedClasspath(classpath.getFiles());
             if (supplier == null) return resolvedClasspath;
             String oldCp = supplier.get();
-            return oldCp == null || oldCp.isEmpty() ? resolvedClasspath : String.join(File.pathSeparator, oldCp, resolvedClasspath);
+            if (Strings.isNullOrEmpty(oldCp)) return resolvedClasspath;
+            if (Strings.isNullOrEmpty(resolvedClasspath)) return oldCp;
+            return String.join(File.pathSeparator, oldCp, resolvedClasspath);
```
SizableShrimp, line 185: How does this guarantee that an empty string is returned if both are null-or-empty? Doesn't look like it does.

SquidDev:
```java
// resolvedClasspath != null by definition of getResolvedClasspath.
if (Strings.isNullOrEmpty(oldCp)) return resolvedClasspath;
// At this point, oldCp != null, so this returns a non-null value.
if (Strings.isNullOrEmpty(resolvedClasspath)) return oldCp;
// Finally, oldCp != null && resolvedClasspath != null
return String.join(File.pathSeparator, oldCp, resolvedClasspath);
```

I can add a comment to clarify if that would be helpful?

SizableShrimp: Ah, that makes sense. Perhaps just add a `@NotNull` annotation to getResolvedClasspath, that should be good enough for the future.

SquidDev: :+1: I've gone for `@Nonnull` as the rest of FG seems to use javax annotations right now, but can change if preferred.

SizableShrimp (approved):
If no one else on the team gets to this, I will probably test it in a few days when I'm at a PC. Cc @marchermans if he wants to test it

marchermans (approved):

SquidDev:
Thanks!

PR #934, opened 2023-08-22
Replace getBuildDir() (Deprecated) with getLayout().getBuildDirectory()

JT122406:
image


PaintNinja (approved):
LGTM - FG already does this in the DownloadMCMeta task, so might as well accept this PR to do it elsewhere too and fix some more deprecations.

PR #939, opened 2023-12-22
Force preprocessor to run on VS Code run configurations

natrow:
Changes:
 - Use Intellij/Gradle output build paths on VS Code run configurations rather than Eclipse ones
 - Add `preLaunchTask` to each run configuration to trigger compilation
 - Create `tasks.json` with a `prepareXXXCompile` task for each run configuration


LexManos:
So should of fixed the gradleutils stuff.
And looking at it basically it looks like you're just forcing the prepare task to be run each time the run task is.
Why do we not just make the run task depend on the prepare task and skip the whole runconfig cluster?

Besides that I dont really care about VSCode so as long as it works im fine.

natrow:
I'm not sure if that's possible for the VSCode launch configurations. AFAIK that is the only way to get them to 'depend' on something already being completed. Anyways it doesn't effect launch times too much.

PR #944, opened 2024-05-02
Update Constants.java

Sekupas:
Fix the problem with downloading error from http protocol


LexManos:
In my tests it send the protocol redirect, but this doesn't hurt.
However this PR isn't possible to be pulled right now as the 2.3 branch needs to be migrated to a modern CI.
the 6.x branch uses TeamCity, but ideally we should migrate to github actions.
If you wanna try that feel free, but i'm gunna poke @PaintNinja about it.

LexManos:
https://github.com/MinecraftForge/ForgeGradle/commit/3951589139c8796861a92f4cd88117b2eed16f12

PR #945, opened 2024-05-03
Fix fg.deobf trying source downloads on every project refresh

embeddedt:
The current `fg.deobf` logic attempts to find the sources jar from the original Maven repository using manual download logic. The problem is that querying this for every `fg.deobf` dependency on each project refresh is quite slow, especially if the sources don't actually exist on the Maven to begin with. (An example of a repository that tends to be like this is CurseMaven.)

This concern was raised [when the code was originally added](https://github.com/MinecraftForge/ForgeGradle/pull/746#discussion_r578238592), but was unfortunately never addressed.

This PR is a simple workaround for the problem. If sources fail to be downloaded for a given artifact, the download failure is flagged in FG's cache and downloads will not be attempted again. This shouldn't break any builds, as sources are only used for better presentation in an IDE, not for compilation.

When I originally added this to NeoGradle 6, I tested with https://github.com/embeddedt/OpenComputers/tree/dev-MC1.16, and it brought the time taken to refresh the project in IntelliJ from 40 seconds down to 6 seconds. Other modders reported similar improvements in their FG6-based projects when using lots of `fg.deobf` statements.

I've since been using this patch for several months without issues.

### DIFF
--- a/src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
+++ b/src/userdev/java/net/minecraftforge/gradle/userdev/util/DeobfuscatingRepo.java
@@ -130,8 +130,22 @@ public class DeobfuscatingRepo extends BaseRepo {
 
     @Nullable
     private File findSource(Artifact artifact, String mapping) throws IOException {
+        // Check if we have previously failed to retrieve sources for the artifact.
+        // If so, don't attempt the download again.
+        File noSourceFlag = cache(getArtifactPath(artifact, mapping) + ".nosources");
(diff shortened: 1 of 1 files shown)


Jonathing:
I trust this patch works well. After I am able to test it myself and no issues arise, I'll give this the thumbs up. Will mark Paint as a reviewer in case he can get to it faster than I can, since I'm currently finishing up school.

LexManos:
Interesting, would be better if the nosources flag had some form of caching system. But honestly, doubt that i'd be an issue in the MC community.
Simple enough change.

PR #949, opened 2024-09-10
Update support.yml

RealMangorage:
Fix discord url to the correct discord server

PR #950, opened 2024-09-10
Update support.yml

RealMangorage:
Fix Discord URL to correct URL


Jonathing:
No need. Since #949 targets the default branch, this is covered by the bot.

PR #951, opened 2024-09-11
Add overload for fg.deobf to support version catalogs & variantOf declarations

KitsuneAlex:
This PR introduces a new overload for the deobf() function to allow for smoother Gradle KTS integration; this allows to do the following in Gradle KTS:

```kts
dependencies {
    implementation(fg.deobf(libs.someLib))
    implementation(fg.deobf(variantOf(libs.someLib) { classifier("api") }))
}
```

This is way cleaner than converting to strings in-between.


gradle/libs.versions.toml
```diff
@@ -0,0 +1,42 @@
+[versions]
+forgeGradle = "6.0"
```
LexManos: I am not a fan of using toml for stuff like this, IF we were to do this then putting it in the settings.gradle would be the way to go as thats how I do it for all other projects in the org
https://github.com/MinecraftForge/Bootstrap/blob/master/settings.gradle#L12

However I've not bothered to do that here because there is no sub-project so there is no need to sync versions between them. So it's not really worth doing. Only real benefit would be the cases where we use the same version for multiple dependencies.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10.1-all.zip
```
LexManos: Bumping this isn't a bad idea, but I tend to keep it on the least supported version so that we don't accidently include gradle API that would break compatibility with 8.4.

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
```diff
@@ -63,6 +65,11 @@ public Dependency deobf(Object dependency, Closure<?> configure) {
         return remapper.remap(baseDependency);
     }
 
+    @SuppressWarnings("unused")
```
LexManos: No need for the suppress unused on public methods...

build.gradle
```diff
@@ -63,35 +64,36 @@ license {
 }
 
 tasks.named('wrapper', Wrapper).configure {
-    gradleVersion = '8.4'
+    gradleVersion = '8.10.1'
```
LexManos: you also didnt run the gradle wrapper task as there are updates to the shell scripts

LexManos:
Best bet is to not do these grouped PRs.
As the deobf change is fine, but its blocked by all the other changes.

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
KitsuneAlex: Then why did you/someone else do it on the extension API function? Just trying to stick to what was already there..

KitsuneAlex:
> Best bet is to not do these grouped PRs. As the deobf change is fine, but its blocked by all the other changes.

Didn't know there would be so many issues with this ^^

gradle/wrapper/gradle-wrapper.properties
KitsuneAlex: But it is a minor version, not a major one. Best way to go would be to just bump FG to 6.1 maybe?

gradle/libs.versions.toml
KitsuneAlex: Fair enough, thought you might wanna stick to the recommended ways.

build.gradle
KitsuneAlex: That is true, forgot about that!

gradle/wrapper/gradle-wrapper.properties
PaintNinja: Gradle sometimes makes breaking changes in minor versions

gradle/libs.versions.toml
LexManos: Gradle notoriously 'recommends' many things, and then changes their mind every 3 months.
The settings.gradle is the same thing as the toml. They both use the same system just declared in a different location. 
Both are 'recommended' by gradle.

gradle/wrapper/gradle-wrapper.properties
LexManos: If we drop support then yes it would require a version bump on our end. Which honestly I dont think is worth it. There is no compelling reason to bump gradle.

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
LexManos: Honestly, no idea IDEA users like to throw suppress warnings all over the place that themselves end up just causing warnings in other IDEs. There is no reason to have a suppress here.

gradle/libs.versions.toml
KitsuneAlex: They do not in fact use the same system, version catalogs are type safe. But since this project uses a Groovy buildscript that doesnt really matter anyways. I removed the catalog in my latest commit, consider it fixed ^^

gradle/wrapper/gradle-wrapper.properties
KitsuneAlex: Rolled back Gradle in latest commit

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
KitsuneAlex: Removed redundant supressions in latest commit

build.gradle
KitsuneAlex: Rolled back Gradle, consider fixed

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
```diff
@@ -51,7 +53,6 @@ public ArtifactRepository getRepository() {
         return repository;
     }
 
-    @SuppressWarnings("unused")
```
Jonathing, line 54: No need for this change. Please remove it.

Jonathing (changes requested):
Please update your PR description to reflect the shift in direction it has received.

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
KitsuneAlex: I am confused now, Lex told me that he didn't want these supressions in there.

KitsuneAlex:
> Please update your PR description to reflect the shift in direction it has received.

Sure :)

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
Jonathing: Ah sorry I read the diff wrong, the moment you sent that I was gonna remove my message because it was my bad haha

Jonathing: Ah sorry I read the diff wrong, the moment you sent that I was gonna remove my message because it was my bad haha

Jonathing (approved):
Cool. Code looks sane.

KitsuneAlex:
This seems to cause some issues with certain dependency configurations because MinimalExternalDeps are immutable. I'll fix the issue really quick! ^^

KitsuneAlex:
That seems to work nicely, just internally converting into the required four-segment dependency notation when a `DependencyVariant` is passed into `deobf()`.

src/userdev/java/net/minecraftforge/gradle/userdev/DependencyManagementExtension.java
```diff
@@ -68,7 +70,12 @@ public Dependency deobf(Provider<MinimalExternalModuleDependency> provider) {
         MinimalExternalModuleDependency dependency = provider.get();
         project.getConfigurations().getByName(UserDevPlugin.OBF).getDependencies().add(dependency);
 
-        return remapper.remap(dependency);
+        ModuleIdentifier module = dependency.getModule();
+        if(dependency instanceof DependencyVariant) {
+            DependencyVariant variant = (DependencyVariant) dependency;
+            return deobf(String.format("%s:%s:%s:%s", module.getGroup(), module.getName(), dependency.getVersion(), variant.getClassifier()));
+        }
+        return deobf(String.format("%s:%s:%s", module.getGroup(), module.getName(), dependency.getVersion()));
```
LexManos, line 78: This doesn't seem right, it causes the dependency to be added twice, once using the object passed in, and the second as a string.

KitsuneAlex: How and where? I tested it, it works fine and there's no duplicates, at least not for me..

KitsuneAlex: A dependency passed _into_ deobf as a provider is not added to the dep graph in any way yet. `MinimalExternalModuleDependency` is used as a means to define type safe version catalog entries.

LexManos: Yes, but you add the dependency to the graph on line 71, then pass in a string variant to deobf which in turn adds he string variant on line 64.
The issue is that the version will not be updated to include the `_mapped_` identifier that the remapper adds to let the deobf repo know to provide the dep.

You're also depending on internal gradle API using DependencyVaraint. Is there a public version of that API?
I couldn't find one, perhaps it would be better to add a secondary deobf function that took in a `Action<? super ExternalModuleDependencyVariantSpec> variantSpec` and passed it to the `variantOf` function ourselves. this making the syntax potentially `fg.deobf(libs.dep) { classifier('api') }`

Simplest solution I think would be to just remove line 71.

Is there a way to get a guaranteed mutable version of MinimalExternalModuleDependency? Specifically we need to modify the version constraints to include the mapped name.

Perhaps this is time for someone to figure out how to properly use the Variant/Attribute system gradle introduced after all of this was written to encode the required mapping version as an attribute. And then have the Deobf repo provide an artifact with that attribute. I have not looked into the system in years so I am unsure if this is possible.

But if it is, it would allow us to skip hacks we do to remove the `_mapped_`  entries from the generated pom, and the force disabling of the module metadata files. Thus simplifying the process.

KitsuneAlex: > Perhaps this is time for someone to figure out how to properly use the Variant/Attribute system gradle introduced after all of this was written to encode the required mapping version as an attribute. And then have the Deobf repo provide an artifact with that attribute. I have not looked into the system in years so I am unsure if this is possible.

I'd like to consider myself fairly well versed with Gradle from a plugin perspective, but i'd like to see a better way too and maybe improve my skills. I'll look into it.

> Is there a way to get a guaranteed mutable version of MinimalExternalModuleDependency?

Not any that i'm aware of right now sadly, that's why i've attempted to do it the way i did.

> You're also depending on internal gradle API using DependencyVaraint. Is there a public version of that API?
I couldn't find one, perhaps it would be better to add a secondary deobf function that took in a Action<? super ExternalModuleDependencyVariantSpec> variantSpec and passed it to the variantOf function ourselves. this making the syntax potentially fg.deobf(libs.dep) { classifier('api') }

This sounds really good! ^^ I will implement these changes in a moment.

LexManos: Feel free to join the discord and we can shat about some ideas there. But ya, I personally hate Gradle with a burning passion because they don't know how to make things simple, and don't know how to keep things stable. But if you wanna poke at things, I have a few ideas and it wouldn't hurt to try.
https://discord.minecraftforge.net

KitsuneAlex:
I'm going to reopen this from a different source branch in a bit, cleaning up my local repository.

PR #953, opened 2024-09-14
Draft: Use dependency attributes instead of version markers for remapped dependencies

KitsuneAlex:
This adds two new overloads to the `deobf` function in `DependencyManagementExtension` which allows clean integration with Gradle version catalogs, especially useful when working with the Kotlin DSL.


Jonathing:
I noticed that this PR was still open, so I'll quickly give my two cents on this.

If we wanted to merge this, it would probably be for a ForgeGradle 6.2, since changes to how dependencies are mapped will likely break plugins that hook into ForgeGradle. While that isn't inherently bad, ForgeGradle is historically one of those plugins that have always been used with other plugins that hook into it. So, I don't see myself merging this.

I'm taking a completely different approach to deobfuscated dependencies in ForgeGradle 7.0 that doesn't use this old system at all. If I think a ForgeGradle 6.2 is warranted, then I'll consider merging this in and working with it as such. For now though, consider this as stale. I'll close when ForgeGradle 7.0 releases which will have better, proper support for deobfuscating external module dependencies in a non-intrusive manner.

Jonathing:
Closing as this will be superseded by a separate plugin (`net.minecraftforge.obfuscation`).

Also, if your goal wasn't to respect the laziness of providers, a lot of this setup seems a little unnecessary. In ForgeGradle 6, I added support for version catalog entries by simply unpacking the dependency providers. This is safe because realizing `MinimalExternalModuleDependency` instances is rather inexpensive. However, in the new plugin, I will be writing deobf dependencies in a way that *attempt to* will defer creation of dependency objects, at least until after project configuration. For now, it's out of scope for ForgeGradle 6.

PR #955, opened 2024-10-22
Fg 1.2

Midaco-YT:


Jonathing:
ForgeGradle 1.2 is out-of-date and unsupported. If you need to make modifications to it, you should fork it yourself.

LexManos:
If he can get it building again and it's just a matter of accepting the pr then I'm fine with it. However this doesn't seem to do that.

PR #963, opened 2025-03-01
Move to GitHub Actions

Jonathing:
Unfortunately I have no real way to test this, but I made sure to set all the correct values, including the funky group name FG uses.


.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
...
+  contents: read
+
+jobs:
+  build:
+    uses: MinecraftForge/SharedActions/.github/workflows/gradle.yml@main
+    with:
+      java: 11
+      gradle_tasks: "publish"
+      artifact_group: "net.minecraftforge.gradle"
+      artifact_groups: "net.minecraftforge.gradle"
```
LexManos: Its one or the other

LexManos:
Nuke Jenkinsfile while you're at it. Its from when we used Jenkins before Team City

PR #967, opened 2025-05-23
Fixes 1.10.2 setupDecomWorkspace

SkyBlade1978:
This fixes the setupDecompWorkspace for MC 1.10, due to http protocol no longer accepted

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/common/Constants.java
+++ b/src/main/java/net/minecraftforge/gradle/common/Constants.java
@@ -116,7 +116,7 @@ public class Constants
 
     // urls
     public static final String URL_MC_MANIFEST     = "https://launchermeta.mojang.com/mc/game/version_manifest.json";
-    public static final String URL_ASSETS          = "http://resources.download.minecraft.net";
+    public static final String URL_ASSETS          = "https://resources.download.minecraft.net/";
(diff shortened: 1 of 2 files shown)


PaintNinja:
Duplicate of #927 and #944. Yes the fix is simple, no we can’t merge it because the build and publishing are broken.

We need the build and publishing fixed first.

SkyBlade1978:
Thats a shame, for now I have built it and included it as a local lib which seems to work out just fine, if a bit of an awkward solution

SkyBlade1978:
In case it helps anyone my build with the fix is here:

https://github.com/SkyBlade1978/IronAgeFurniture/tree/master-1.10/libs

LexManos:
That's not what he means, he means rhe build.gradle needs updated and a github actions workflow needs to be added to publish FG itself. The new branches build fine but the old doesnt. so if someone wants to do that feel free.

Once that's done pulling in these prs will be fine.

LexManos:
Alright, so got all old branches on to github workflows and publishing now.
However this PR is wrong, you add a / at the end and then the downloader adds another /, so you need to remove that.

LexManos:
https://github.com/MinecraftForge/ForgeGradle/commit/9b2e4d13c32ba11d2940fdab19b234178f4e5606

PR #971, opened 2025-06-17
[FG7] Add a way to configure Tools (Mavenizer/Slimelauncher)

RealMangorage:
This is mainly a PR to discuss the most ideal way of handling the configurability of these tools that FG7 uses.

The build script would look like this, to override a tool.

```groovy
dependencies {
    mavenizer "net.minecraftforge:minecraft-mavenizer:0.3.3"
    slimelauncher "net.minecraftforge:slime-launcher:0.1.0"

    implementation(minecraft.dep("net.minecraftforge:forge:1.21.5-55.0.23"))
}
```

### DIFF
--- a/src/main/groovy/net/minecraftforge/gradle/ForgeGradlePlugin.groovy
+++ b/src/main/groovy/net/minecraftforge/gradle/ForgeGradlePlugin.groovy
@@ -82,6 +82,17 @@ class ForgeGradlePlugin implements Plugin<ExtensionAware> {
             this.&getFileSystemOperations,
             this.&getArchiveOperations
         )
+
+        if (target instanceof Project) {
+            Project project = (Project) target;
+
+            for (final def tool in Tools.values()) {
+                project.getConfigurations().register(tool.getConfiguration()) {
+                    it.setTransitive(false) // Cant be transitive, maybe allow it to be later?
+                    it.setVisible(true)
+                }
+            }
+        }
     }
 
     @PackageScope DirectoryProperty getGlobalCaches() {
@@ -96,6 +107,12 @@ class ForgeGradlePlugin implements Plugin<ExtensionAware> {
     @PackageScope Provider<File> getTool(Tools tool) {
         tool.get(this.globalCaches, this.providers)
     }
+
+    @SuppressWarnings('GrDeprecatedAPIUsage') // Intentional deprecation, please use this method
+    @PackageScope Provider<File> getTool(Tools tool, Project project) {
+        tool.get(project, this.globalCaches, this.providers)
+    }
+
 
     @CompileDynamic
     private File getGradleUserHomeDir(ExtensionAware target) {
--- a/src/main/groovy/net/minecraftforge/gradle/ForgeGradlePlugin.groovy
+++ b/src/main/groovy/net/minecraftforge/gradle/ForgeGradlePlugin.groovy
@@ -12,6 +12,7 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.ArchiveOperations
 import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.file.FileCollection
 import org.gradle.api.file.FileSystemOperations
 import org.gradle.api.file.ProjectLayout
 import org.gradle.api.flow.FlowProviders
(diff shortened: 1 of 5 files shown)


Jonathing:
There are a couple of things about your implementation I don't like. But with regards to what we want to add, I do like this. 

Originally, I was planning something like this:
```groovy
minecraft {
    mavenizer = 'me.jonathing:mavenizer:0.3.0'
}
```

But this works better. I do think I need to clean up a bit of my code with the Tools enum, it's not entirely finished.

RealMangorage:
I did it this way, as to not *change* any method signatures, just add new ones, as I was not entirely sure how you wanted to deal with it.

I can go thru everything (my impl) and clean it up for you, if you would like to.

src/main/groovy/net/minecraftforge/gradle/ForgeGradlePlugin.groovy
```diff
@@ -82,6 +82,17 @@ class ForgeGradlePlugin implements Plugin<ExtensionAware> {
...
         )
+
+        if (target instanceof Project) {
+            Project project = (Project) target;
+
+            for (final def tool in Tools.values()) {
+                project.getConfigurations().register(tool.getConfiguration()) {
+                    it.setTransitive(false) // Cant be transitive, maybe allow it to be later?
+                    it.setVisible(true)
+                }
```
Jonathing: For the record, this absolutely can be transitive. Instead of having Tools return a `Provider<File>`, it would instead be better to use the configuration you created in `ForgeGradlePlugin`, add the default tool to that configuration, and then use that as the classpath in the task.

src/main/groovy/net/minecraftforge/gradle/Tools.java
```diff
@@ -25,15 +27,17 @@
...
-    SLIME_LAUNCHER("slime-launcher-" + Constants.SL_VERSION + ".jar", Constants.SL_DL_URL);
+    MINECRAFT_MAVEN("minecraft-maven-" + Constants.MCMAVEN_VERSION + ".jar", Constants.MCMAVEN_DL_URL, "mavenizer"),
+    SLIME_LAUNCHER("slime-launcher-" + Constants.SL_VERSION + ".jar", Constants.SL_DL_URL, "slimelauncher");
 
     private final String fileName;
     private final String downloadUrl;
+    private final String configuration;
 
-    Tools(String fileName, String downloadUrl) {
+    Tools(String fileName, String downloadUrl, String configuration) {
```
Jonathing: This might be better off using the Enum's name with `String#toLowerCase(Locale.ENGLISH)`.

Jonathing (commented):
I think this whole bit might be better in an extension. I'm still thinking of potentially an alternate implementation? One that looks like this:
```groovy
dependencies {
    fgTools.mavenizer 'me.jonathing:mavenizer:0.3.0'
}
```
To do that, I'd need to make a new extension for `DependencyHandler` that returns configurations. If you'd like to present your own take on this, feel free.

src/main/groovy/net/minecraftforge/gradle/Tools.java
RealMangorage: I went with this, due to the underscore, but I could remove that, and do the needed operations within the function still, to get rid of the underscore.

src/main/groovy/net/minecraftforge/gradle/ForgeGradlePlugin.groovy
RealMangorage: Ah great! I will fix that.

src/main/groovy/net/minecraftforge/gradle/Tools.java
Jonathing: ah ok

Jonathing:
Closing because I added an experimental API that allows for this. The implementation details have yet to be finalized, but it works like this:
```groovy
fgtools {
    configure('mavenizer') {
        classpath = configurations.detachedConfiguration(
            dependencies.create('net.minecraftforge:minecraft-mavenizer:0.3.22')
        )
    }
}
```
`classpath` in this case is a `ConfigurableFileCollection`. I will not add helper methods for things like String artifacts because this is meant for advanced users only.

PR #974, opened 2025-07-20
[3.0] Workaround potential missing repositories within MavenArtifactDownloader

Jonathing:
There's a re-occurring and inconsistent bug with ForgeGradle 3 where the MavenArtifactDownloader will use 0 repositories to download a dependency for the MCP toolchain. In my testing, older projects would continuously fail to download `net.minecraftforge:accesstransformers:1.0.+`, despite this artifact existing on the Forge maven just fine.

My workaround for this is simply to add the requested downloading repositories on top of the project's existing repositories before downloading the artifact in `MavenArtifactDownloader#_gradle`.

For full transparency, this PR was done at the request of my friends at @Modding-Legacy to help them preserve older repositories without needing to update the buildscript on each one.

PR #977, opened 2025-07-21
[3.0] Do not fail download if content length is unknown

Jonathing:
- Supersedes #974.
- Partial backport of #905 to ForgeGradle 3.

ForgeGradle 3 has its own Maven artifact downloader, which in turn used a combination of Java's (Http)UrlConnection and Apache's Commons IO library. An oversight with the spec of `UrlConnection#getContentLength` has led to an issue where a downloaded file can be erroneously discarded of a source does not declare the content length in its header. This was the case for `https://maven.minecraftforge.net/net/minecraftforge/accesstransformers/maven-metadata.xml`, which caused the downloader to refuse to download AccessTransformers and thus stifle the build process.

This PR fixes this issue by simply skipping the content length check if it is unknown. `-1` is a valid return value if the content length is known, and should not be treated as an error nor overlooked.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/Utils.java
@@ -426,7 +426,7 @@ public class Utils {
                 read = IOUtils.copy(stream, out);
             }
 
-            if (len >= 0 && read != len) {
+            if (len != -1 && read != len) {
                 output.delete();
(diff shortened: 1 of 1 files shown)


Jonathing:
Ok, digging into it a bit more, this issue was already fixed in ForgeGradle 5 by #905. It was just never backported. So consider this the backport. I'm going to go ahead and merge this since this is handled in FG5 and FG6 exactly like so, and is a harmless one-line change.

PR #979, opened 2025-07-25
[6.0] Add support for version catalog versions in `minecraft.mappings`, `jarJar.pin` and `jarJar.ranged`

Jonathing:
Similar to #976, allows for using version catalog versions (i.e. `Provider<String>` or `ProviderConvertible<String>`) in mappings declarations and Jar-in-Jar dependency version declarations.

settings.gradle:

```groovy
dependencyResolutionManagement.versionCatalogs.register('libs') {
    version 'mappings-channel', 'official'
    version 'mappings-version', '1.21.8'
}
```

build.gradle:

```groovy
minecraft {
    mappings channel: libs.versions.mappings.channel, version: libs.versions.mappings.version
}
```

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/MinecraftExtension.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MinecraftExtension.java
@@ -16,6 +16,10 @@ import org.gradle.api.provider.Provider;
 import groovy.lang.Closure;
 import groovy.lang.GroovyObjectSupport;
 import groovy.lang.MissingPropertyException;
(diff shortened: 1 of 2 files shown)


LexManos:
Seems weird to declare it that way, but I don't question gradles weirdness

PaintNinja (approved):
Agreed, looks strange for mappings as it’s not being used in the dependencies block like you’d expect, but from a version catalogue perspective it’s fine.

Support for this in JiJ however is nice, approved.

PR #980, opened 2025-07-26
[6.0] Version catalog version support (re-try)

Jonathing:
- Reapplies and fixes #979 (edf7bf3).

My oversight with this PR was that I was enforcing a type requirement of `String`, which `GString` and other subclasses of `CharSequence` are not. Instead of attempting to cast the value to a string, I am using `String#valueOf` before passing it in to the property. This is basically the same as what is already done, the main difference now being that this removes the enforcement on the input being a CharSequence.

### DIFF
--- a/src/common/java/net/minecraftforge/gradle/common/util/MinecraftExtension.java
+++ b/src/common/java/net/minecraftforge/gradle/common/util/MinecraftExtension.java
@@ -17,8 +17,6 @@ import groovy.lang.Closure;
 import groovy.lang.GroovyObjectSupport;
 import groovy.lang.MissingPropertyException;
 import org.gradle.api.provider.ProviderConvertible;
-import org.jetbrains.annotations.Contract;
-import org.jetbrains.annotations.Nullable;
 
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


Jonathing:
Let me know if I should replace `String::valueOf` with `Object::toString`. They both do the same thing except `String::valueOf` handles null values by returning `"null"`.

Jonathing:
And obviously I will do better tests this time with non-Strings.

LexManos:
My only real question is does this eagerly resolve the GStrings/variables used and could that cause issues with existing setups?

Jonathing:
If it's not used in a provider, yes. That was the case before this as well. Could easily throw it in a provider tho.

LexManos:
I'd say part of the point of using GStrings is the delayed resolution. So thats needed

Jonathing:
Done. Using `Provider#map` affects the value on resolution and not immediately.

Jonathing:
Tested and this works for real now, used this cursed thing as reference:
```groovy
mappings channel: "${Boolean.valueOf(true) ? 'sugarcane' : 'official'}", version: libs.versions.mappings.version
```
Boolean.valueOf just incase the Groovy interpreter wanted to abstract away a simple string interpolation.

Apologies again @KnightMiner for causing you a headache but I wanna give you massive props for informing me to the bug so we could get it reverted as soon as we got word of it. Cheers. :beers:

PR #981, opened 2025-07-27
[6.0] Fix oversight with variants using fg.deobf with version catalog entries

Jonathing:
- Follows up on #976.

See the comment in my code for details. The short of it is that if we don't immediately create the dependency from the provider, additional info such as user-defined variants could be lost. This does not apply to bundles as the dependencies within them cannot be mutated with methods such as `DependencyHandler#variantOf`.

A quick note for future reference: `Property#finalizeValue` immediately calculates the value while `Property#finalizeValueOnRead` does not, but both put the property into a read-only state.


LexManos:
I hate dirty hacks, but if it works go for it.

Jonathing:
I've been doing my best to keep the dirty hacks as minimal as possible. Unfortunately there is only so much I can do with Gradle.

This problem will solve itself in FG7. The plan is to define the deobf status of dependencies using variants enforced by dependency constraints, and those variants will be output by the Mavenizer. So, instead of needing to copy the dependency to an arbitrary remapper, I can request the mappings attributes and tell the Mavenizer that we need this output.

Jonathing:
Tested this on my end and it works. But in case I wrote something shitty again (despite even more testing than last time), this affects a new method that most codebases do not even use.

Issue #991, opened 2025-10-08
[7.0] Intentionally support Configuration on Demand


Jonathing:
[Configuration on Demand](https://docs.gradle.org/current/userguide/configuration_on_demand.html) is an optimization feature in Gradle which makes it attempt to only configure the relevant projects for the requested tasks. This does not affect project import.

It is imperative to support this for multi-loader projects, as ForgeGradle is usually applied to a subproject that does not always need to be configured (i.e. running the task `:fabric:jar`). Configuration on demand also means that the Mavenizer invocation (discussed in #990) can be skipped if the project does not need to be configured.

So far, ForgeGradle 7 supports configuration on demand as the plugin for the applied project does not need to interact with any other projects. However, I'm creating this issue to ensure that I don't break configuration on demand by accident for any updates I make to ForgeGradle 7 until stabilization.

This issue will be closed as complete once ForgeGradle 7 is stabilized.

cc @PaintNinja @mezz

Jonathing:
I will be closing this issue since, as of ForgeGradle 7.0.0-rc.0, compatibility with Configuration on Demand is alive, well, and going on strong. If you notice any bugs with configuration on demand, please open a new issue so I can take care of it. Thank you!

Issue #986, opened 2025-10-06
[7.0] Sample MDK fails to compile or run when ATs are applied


PaintNinja:
Using the recent sample MDK provided on the Discord ([msg link](<https://discord.com/channels/1129059589325852724/1129095235889270844/1424385102569279530>)), setup completes however IntelliJ fails to find Forge and MC:
<img width="485" height="588" alt="Image" src="https://github.com/user-attachments/assets/b7e8d2b0-8091-4c87-a539-33529aa0e81d" />

Attempting to use the `runClient` task gives the following error:
```
Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Failed to transform forge-1.21.9-59.0.3.jar (net.minecraftforge:forge:1.21.9-59.0.3) to match attributes {artifactType=jar, net.minecraftforge.gradle.accesstransformers.automatic.0=true, net.minecraftforge.mappings.channel=official, net.minecraftforge.mappings.version=1.21.9, net.minecraftforge.native.operatingSystem=windows, org.gradle.category=library, org.gradle.dependency.bundling=external, org.gradle.jvm.environment=standard-jvm, org.gradle.jvm.version=21, org.gradle.libraryelements=jar, org.gradle.status=integration, org.gradle.usage=java-runtime}.
      > Execution failed for ArtifactAccessTransformer: C:\Users\oscar\Documents\Projects\FG7-MDK-59.0.0\build\minecraftforge\forgegradle\mavenizer\output\net\minecraftforge\forge\1.21.9-59.0.3\forge-1.21.9-59.0.3.jar.
         > Cannot query the value of this property because it has no value available.
```

Build scan: https://scans.gradle.com/s/pwmjeg2ih43eu

Removing the AT plugin, the `accessTransformers = true` line in the minecraft block and the use of the private logger field fixes the above issues regarding finding Forge, compiling and running, so this appears to be an issue with ATs specifically.

Jonathing:
ForgeGradle 7 now directly registers ArtifactAccessTransformer itself without using AT Gradle's API, so this is a ForgeGradle 7 bug. I'll look into what could be happening when I have the time. I might need to change the strategy of how the dependency is resolved, because I think right now it's timing-sensitive, while AT's resolution strategy is not. Thanks for the report, brother.

Jonathing:
[...]

PR #994, opened 2025-10-08
[7.0] Invoke Mavenizer, asynchronously, at configuration time

### LINKED ISSUE #989: [7.0] Slime Launcher config creation has questionable registration
The configurations for the Slime Launcher runs are generated based off the contents of the `metadata.zip` file that is generated as an artifact from Mavenizer. Currently, ForgeGradle 7 handles this by querying this file at configuration and setting each run task's convention values to what is in the run configs stored in that ZIP file. However, this can cause problems as configuration is not always guaranteed to run.

Instead of reading the `metadata.zip` at configuration, it should instead be passed in as an input of `SlimeLauncherExec`. When the task is executed, it should process the metadata and set values for any properties that have none set, so that any defaults may be used where values are not declared in the buildscript.

This will have one changed with project import. Run configurations will now have to be validated when that run task is executed, and not when the project finished importing (in after evaluate). But this is way better than trying to locate and use a volatile file which may or may not exist, especially for new projects.

### LINKED ISSUE #990: [7.0] Move Maven generation to configuration time
I originally made ForgeGradle 7 with the idea of having the maven generation be a task that is run on project sync. I am now having doubts about this implementation, especially considering the limitation brought by #989.

My idea now is to move the invocation of Mavenizer to `project.afterEvaluate`. I don't know if I'm able to parallelize this, however, as I've had trouble using [`WorkerExecutor`](https://docs.gradle.org/current/userguide/service_injection.html#workerexecutor) outside of tasks. Maybe the traditional JDK's `CompletableFuture` could work? I can await the completion of Mavenizer in `gradle.projectsEvaluated` so it doesn't seep into execution time.

Honestly a little torn on what exactly I should do here. Would love feedback from others (hopefully who know more than I do) on how I could tackle this in a way that doesn't destroy caching. My initial idea is to run the Mavenizer in a custom `ValueSource` so that Gradle is aware of the execution process.

Jonathing:
- Implements #990
- Fixes #989
- Partially addresses #991

This PR changes the invocation strategy of Mavenizer from an execution-time to a configuration-time completable future call.

For my reviewers: what I'm asking for is less of a code review (please do be thorough with `MavenizerAction`, though) but more of a sanity check.

The problem with invoking Mavenizer in a task is that it might not always be run if a change to a dependency is made. Adding tasks to `gradle.startParameter.taskRequests` is questionable at best and may not always produce the results we want. If a dependency change is made and the task is not invoked, the classpath configuration containing the Minecraft dependency will fail to resolve since it is looking for dependencies and/or variants that do not exist.

By running Mavenizer at configuration time, we are guaranteed to have the variants generated as needed at execution time. The Mavenizer also outputs the `metadata.zip` file that is used by the Slime Launcher run tasks to set default values. Keep in mind that Slime Launcher itself cannot use these values as JVM arguments cannot be modified mid-process.

I've additionally opted to have Mavenizer start in the background by using a completable future. Since Mavenizer is completely removed from Gradle land, this does not have any ramifications to the Gradle build itself. When *all* projects have finished evaluating (`gradle.projectsEvaluated`), the configuration time is paused until all requested Mavenizer invocations have finished. This means that if Mavenizer is up-to-date according to its own cache, this results in a near-zero drop in performance on project configuration.

There are a few important breaking changes and caveats this introduces:
- Minecraft mappings must be set either in the `minecraft {}` block BEFORE the dependency is declared with `minecraft.dependency`, or it must be declared inside of the `minecraft.dependency {}` block. Not after the dependency is declared.
- I am changing `minecraft.dep` to `minecraft.dependency` so it is clear what the method does at first glance. I will include a `methodMissing` in the implementation so `minecraft.dep` can continue to work.
  - Since ForgeGradle 7 is planned to be the *only* project with a dependency wrapper, this shouldn't be an issue.
- Mavenizer itself can run in parallel, which it currently isn't written for. I will add a warning message if there is more than one Minecraft dependency requested for a project. But, it should still be a goal in Mavenizer to be aware that there could be multiple instances of itself running.


Jonathing:
This is ready for review. There are other changes I want to make to FG7 soon, but will wait on this getting approved. Let me know your thoughts soon, please.

Jonathing:
Approval was effectively granted over Discord.
https://discord.com/channels/1129059589325852724/1129095235889270844/1425589050198462574

We're still in beta so if anything breaks it's not that big of a deal and I can fix it quickly (or revert this in the worst case).

Jonathing:
Looks like this is going to need to be reverted. Unfortunately I completely forgot that Gradle does not like it if you run any processes outside of a `ValueSource` at configuration time. And I am just not given enough control over the execution of the process in that context.

I'm currently researching ways I am able to invoke Mavenizer as a task, like before, but in a way to ensure that it is always run instead of sometimes occasionally. A bug with this approach last time that was noted in #986, is that changes to the Minecraft dependency absolutely must trigger this task to be executed, or else weird build failures could occur. It's not ideal, but I'm thinking of using some combination of `gradle.startParameter.taskRequests` again along with a few other things I've been looking into.

Issue #992, opened 2025-10-08
[7.0] Intentionally support the Configuration Cache


Jonathing:
The [Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html) is an optimization feature in Gradle which makes it attempt to skip the configuration phase of a Gradle invocation in order to save build time when running tasks repeatedly.

Support for this is highest priority, as the configuration cache has proven to save time in builds across a variety of projects, and will also be the default execution mode in Gradle 10.

So far, ForgeGradle 7 supports the configuration cache as all plugins, extensions, and tasks have been created with it in mind, using the proper properties and (Gradle-managed) types. However, I'm creating this issue to ensure that I don't break configuration caching by accident for any updates I make to ForgeGradle 7 until stabilization.

This issue will be closed as complete once ForgeGradle 7 is stabilized.

cc @PaintNinja @mezz

Jonathing:
Unfortunately hit a bit of a regression on this. Here's the relevant yappage on Discord in case it interests you:
https://discord.com/channels/1129059589325852724/1129095235889270844/1427830987672260730

In short, I just need to re-implement the Mavenizer invocation again to be a task. Need to make it in a way where it always runs if there's some sort of change to the Minecraft dependency.

Jonathing:
Configuration cache compatibility has been restored by #996.

Jonathing:
All lingering issues with the configuration cache should now be fixed by 9844177. I will close this issue upon the full release of ForgeGradle 7.

Jonathing:
I will be closing this issue since, as of ForgeGradle 7.0.0-rc.0, compatibility with the Gradle Configuration Cache is alive, well, and going on strong. If you notice any bugs with configuration caching, please open a new issue so I can take care of it. Thank you!

PR #998, opened 2025-10-20
[7.0] Lots of fixes

Jonathing:
- Addresses #992.
- Fixes Slime Launcher run tasks not working.
- Fixes variant mismatch on macOS.

Depends on:
- MinecraftForge/Utils#14
- MinecraftForge/Utils#15
- MinecraftForge/Utils#16
- MinecraftForge/Utils#17
- MinecraftForge/JavaProvisioner#9
- MinecraftForge/MinecraftMavenizer#6

---

Please don't squash and merge through GitHub.com. I'll merge this myself with the proper tag(s) once approval has been granted.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,9 +15,10 @@ plugins {
 
 gradleutils.displayName = 'ForgeGradle'
 description = "Forge's Gradle plugin for Mod development"
-base.archivesName = 'forgegradle'
 group = 'net.minecraftforge'
 version = gitversion.tagOffset
+
+println "Version: $version"
 
 java {
     toolchain.languageVersion = JavaLanguageVersion.of(17)
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,8 +8,5 @@ org.gradle.configuration-cache.problems=warn
 
 systemProp.org.gradle.unsafe.suppress-gradle-api=true
 
-net.minecraftforge.gradleutils.publishing.use-base-archives-name=true
 net.minecraftforge.gradleutils.ide.automatic.sources=true
 net.minecraftforge.gradleutils.compilation.defaults=true
-
-net.minecraftforge.gitversion.log.version=true
--- a/settings.gradle
+++ b/settings.gradle
@@ -20,12 +20,12 @@ gradle.beforeProject { Project project ->
 
 //@formatter:off
 dependencyResolutionManagement.versionCatalogs.register('libs') {
-    version 'gradleutils', '3.3.20'
+    version 'gradleutils', '3.3.21'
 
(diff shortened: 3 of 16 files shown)


Jonathing:
This is next, and will take a bit. Once complete, I will request reviews. If there are no blocking reviews within an hour, I will merge immediately afterwards.

Jonathing:
This is ready for review. I've tested my own example MDK with this build and all works well. If there are no blocking reviews within the next 2 hours, I will squash and merge.

Jonathing:
I should clarify that there are a good number of things I will be doing after this is merged, but will not be included in this PR. These include:
- Bumping minimum required Gradle to 9.2.0.
- Moving all implementations to the `internal` package.
  - This does remove the `sealed` attribute from the API interfaces, but is not really an issue.
  - The internal implementations will still be inaccessible, as the plugin entry-point can be package-private.
- Freezing API.
  - This means that once API is frozen, no more breaking changes may occur, even after 7.0.0 is stabilized. I'll be making a dedicated PR for this.

PaintNinja:
You should be able to keep sealed if you have a module-info. Sealed across packages only requires it at compile-time iirc.

The internal classes would need to be public and have the side-effect of being accessible outside of module mode, however people still wouldn't be able to access much if the methods and fields inside are package-private. Or you could say screw anyone that hacks into classes in the internal package in legacy mode... I'm not sure if IDEs would suggest them when not exported by the module(?)

Jonathing:
> You should be able to keep sealed if you have a module-info. Sealed across packages only requires it at compile-time iirc.

Unfortunately, Gradle trashes modules in their runtime. I've already tried this. Module details are not respected by Gradle. This is especially the case because they use custom class loaders and injections for everything.

> The internal classes would need to be public

They do not. The plugin entry-point class can be package-private (since the constructor is public and annotated with `@javax.inject.Inject`), so the implementations can remain package-private as well.

> I'm not sure if IDEs would suggest them when not exported by the module(?)

IntelliJ's linter for Gradle plugin classpath is not great. Even when using Gradle 9, it will often use the Java 8 linter which fails to account for sealed types in the first place. Best to not take our chances.

The point of me making everything package-private is to limit access for other plugin authors. ForgeGradle 6 was victim of hacky implementations because of this, and we have very little control over its applications because of it.

PaintNinja:
I meant public solely so that it can be referenced by the api package for the permits clause of the sealed interface. The contents of the classes can stay package-private.

But if the only place that it would actually be enforced is other Gradle plugins with a module-info and even the IDEs ignore it in buildscript suggestions, then yeah I can see why you wouldn't want so.

Ugh I wish Oracle would bite the bullet and deprecate the legacy classpath or something...

Jonathing:
> Sealed across packages only requires it at compile-time iirc.

This part totally flew over my head, whoops lmao. Doesn't matter though because I don't want to expose those classes.

PR #996, opened 2025-10-15
[7.0] Re-implement Mavenizer invocation as a task

Jonathing:
- Reverts #994.
- Fixes #995.
- Fixes #986.
- Addresses #992.
- Closes #990.
- Closes #989.
- Closes #987.

This PR re-implements the Mavenizer invocation to once again be a Gradle task that is run. However, this time I've taken the proper measures to ensure that the dependency resolves consistently with its assigned resolvable configuration (i.e. a classpath) and that the proper attributes are accounted for in time.

Additionally, ForgeGradle will check if the output directory for the Minecraft dependency exists before the `compileJava` task is run. This does not fix any issues, but instead gives a useful error message telling the consumer to re-sync the project rather than leaving them stuck with cryptic "class not found" compiler errors.

If no feedback or changes are requested by the end of the day, I will squash and merge this PR and close any related issues (except for #992).

Issue #988, opened 2025-10-06
[7.0] Improve Parchment 404 error messages


PaintNinja:
The Parchment website suggests the following format on [their getting started page](<https://parchmentmc.org/docs/getting-started>) for FG5+: `2025.10.05-1.21.9`. Their site needs updating eventually for FG7 (the v prefix in the version badges is an added source of confusion), but a nice QoL feature would be to detect a legacy version format and show a more helpful error.

For example, when setting the mapping version to `2025.10.05-1.21.9`, it currently dumps this stacktrace:
```
  download[2025.10.05-1.21.9][parchment]
1 actionable task: 1 executed
Exception in thread "main" java.lang.RuntimeException: Failed to execute task `download[2025.10.05-1.21.9][parchment]` which is required by task `srg2names[parchment-2025.10.05-1.21.9]`
	at net.minecraftforge.mcmaven.impl.util.Task$Simple.execute(Task.java:102)
	at net.minecraftforge.mcmaven.impl.repo.Repo$PendingArtifact.get(Repo.java:206)
	at net.minecraftforge.mcmaven.impl.MinecraftMaven.finalize(MinecraftMaven.java:138)
	at net.minecraftforge.mcmaven.impl.MinecraftMaven.run(MinecraftMaven.java:80)
	at net.minecraftforge.mcmaven.cli.MavenTask.run(MavenTask.java:146)
	at net.minecraftforge.mcmaven.cli.Main.run(Main.java:57)
	at net.minecraftforge.mcmaven.cli.Main.main(Main.java:21)
Caused by: java.lang.RuntimeException: Failed to execute task `download[2025.10.05-1.21.9][parchment]`
	at net.minecraftforge.mcmaven.impl.util.Task$Simple.execute(Task.java:116)
	at net.minecraftforge.mcmaven.impl.util.Task$Simple.execute(Task.java:100)
	... 6 more
Caused by: net.minecraftforge.util.download.DownloadUtils$DownloadFailedException: Failed to download https://maven.parchmentmc.org/org/parchmentmc/data/parchment-2025.10.05/1.21.9/parchment-2025.10.05-1.21.9-checked.zip
	at net.minecraftforge.util.download.DownloadUtils.downloadFile(DownloadUtils.java:161)
	at net.minecraftforge.util.download.DownloadUtils.downloadFile(DownloadUtils.java:142)
[...]

PR #1001, opened 2025-10-30
[7.0] Begin ForgeGradle 7.0's Release Candidate lifecycle

Jonathing:
The merging of this PR will mark the beginning of ForgeGradle 7's Release Candidate lifecycle. This crucially means a few things:

- API breaking changes will no longer be accepted.
  - All changes to API must strictly be additive **until ForgeGradle 8.0.**
- There might still be some bugs that need to be ironed out before the full release.
- As we get closer to full release, I will begin slowly phasing out ForgeGradle 6's development (though all versions of ForgeGradle are open to bug fixes and PRs).

As an additional note, I'd like to begin stabilizing the other plugins as well. This includes JarJar Gradle, Multi-Release Java, and JVM Aggregate Testing.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/internal/Constants.java
+++ b/src/main/java/net/minecraftforge/gradle/internal/Constants.java
@@ -28,9 +28,7 @@ final class Constants {
     /// Use these with [java.text.MessageFormat#format(String, Object...)].
     static final class Messages {
         static final String WELCOME = """
-            Welcome to ForgeGradle 7.0 BETA!
-            
-            This is an unstable release of ForgeGradle 7 for testing purposes.
+            Welcome to ForgeGradle 7.0 Release Candidate!
             
             Here are some release highlights:
             - Complete rewrite of the plugin and underlying code.
@@ -41,9 +39,6 @@ final class Constants {
               work-in-progress).
             
             A couple of important things to note:
-            - ForgeGradle 6 and older will no longer be supported, except for critical bug
-              fixes. We will (gently) encourage developers to move away from them and use
-              ForgeGradle 7 instead.
             - Many plugins that worked with ForgeGradle 6, such as Parchment's Librarian,
               do not work with ForgeGradle 7. For most cases (such as parchment), we have
               implemented native support. If ForgeGradle 7 is lacking in some aspect,
@@ -54,13 +49,20 @@ final class Constants {
               If you are on an older version (1.20.4 and older), you will need the
               'net.minecraftforge.renamer' plugin. Many of these come with our provided
               MDK, so this should not be an issue for you.
+            - If you come across any lingering bugs, please report them to our GitHub
+              issue tracker (https://github.com/MinecraftForge/ForgeGradle/issues).
+              Please do not use Discord as a means to report issues, as they will get
+              lost in the discussion very easily.
             
             For more details on this release, see https://docs.minecraftforge.net/en/fg-7.0/""";
 
         static final String WELCOME_CONDITION = """
-            This message will not display again until the next major beta update, ForgeGradle
-            7.1, or the below file is deleted:
-            {}""";
+            This message will not display again until ForgeGradle 7.0 or the below file is
+            deleted:
+            {}
+            
+            Documentation will be coming at a later date. Thank you for testing
+            ForgeGradle 7!""";
 
         static final String MAGIC = """
             This build is using ForgeGradle Magic. ForgeGradle Magic employs automatic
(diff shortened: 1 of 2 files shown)


PaintNinja:
I'm unclear on the current state of FG7, so take this statement more generally:

I'd rather the focus be on getting all the core functionality working reliably before locking in the API. Some past builds of the betas had breaking changes as it was later discovered that it couldn't be fixed without doing so.

This helps avoid unintentional inflation of the major version for the sake of fixing something that requires minor breakages.

Jonathing:
Fixes can always be made without needing to break the API. But if we make a release candidate and start telling people to use it to help us test/migrate, then there's an API breakage that requires them to change something in their buildscript instead of just upgrading ForgeGradle, that can put a bad taste in people's mouths. That's why I am adament on an API freeze for the release candidate cycle.

Please give me a clear list of the core functionalities I need to test over the next few days before starting release candidates. I'll start by making #988 a blocker for release candidates.

PaintNinja:
> Fixes can always be made without needing to break the API.

Not always... that's my point. I'd prefer to delay the RC in favour of more fixes during beta first, just in case you get unlucky and end up needing to break API, as that way it wouldn't be unexpected.

People can use fixed build numbers in their scripts during the beta period as usual. Yes that doesn't solve the MDK/buildscript update issue but isn't any worse than what we have now and would be a huge upgrade for those currently stuck on FG6 and wanting to help with real-world testing.

Once the core functionality is reliably stable and tested, then start RCs.

> Please give me a clear list of the core functionalities I need to test over the next few days before starting release candidates.

- [x] Able to build an existing non-reobf Forge mod
- [x] runClient task works with resources present (i.e. game loads with panorama background, no grey bg. Mod block texture present ingame)
- [x] runServer task works
- [x] Able to see decompiled MC sources in IDE
- [x] Able to apply an AT and compile against the now-public thing from mod code
- [x] Changing Forge build should work
- [x] Changing MC version should work (assuming mod code worked across the two versions)
- [ ] Able to build an existing non-reobf multiloader Forge mod in a multiloader build setup

JiJ becomes much less of a big need once MixinExtras is bundled in Forge and enabled automatically.

Parchment is a nice to have but not a blocker for starting RCs imo. There's currently uncertainty if it'll be continued after [Mojang distributes non-obf binaries](https://www.minecraft.net/en-us/article/removing-obfuscation-in-java-edition)

Jonathing:
Please see #996 for limitations on changing the Forge and Minecraft versions without re-syncing the project or explicitly running `syncMavenizer`

Jonathing:
I've checked everything that is confirmed working so far.

Jonathing:
Gonna push back my preferred due date for the first release candidate by a week so I can give myself enough time to confirm compatibility in multi-loader setups.

PaintNinja:
Needing to resync after changing versions is fine, that's already the norm for FG6. Presumably there's a check or error message if someone forgets?

Jonathing:
Yup! It is attached to compileJava. Though that reminds me to also attach it to the run tasks as well.

Jonathing:
> Able to build an existing non-reobf multiloader Forge mod in a multiloader build setup

jaredlll08/MultiLoader-Template#107

Link more multi-loader templates if you got them, please.

Jonathing:
Good evening. With the exception of Eclipse run configurations, I believe that ForgeGradle 7 is now in a state where it can be used in production. I've tested dropping it in as a replacement for ForgeGradle 6 in the template I linked earlier, along with bumping the Gradle wrapper to 9.2.0, and everything went smoothly.

As for Eclipse, unfortunately, JavaExec tasks still cannot be hooked in with the debugger. This issue can be tracked in eclipse-buildship/buildship#1130. For now, what I can do is make a successor to the `genEclipseRuns` task in ForgeGradle 7 that invokes Slime Launcher to make the configuration for us. It needs to do this since Slime Launcher is responsible for any classpath changes, downloading assets, and other pre-bootstrap tasks.

To clarify, I will *not* be recreating the `genIntellijRuns` and `genVscodeRuns` tasks, as those IDEs are fully capable of attaching the debugger to the Slime Launcher run tasks.

I will leave it up to Lex as to whether the inclusion of `genEclipseRuns` should be a blocker for the Release Candidate lifecycle or the full release.

For now, I will begin working on the full documentation, as well as the upgrade primer from ForgeGradle 6.

PaintNinja:
> Link more multi-loader templates if you got them, please.
- https://github.com/Tslat/Mod-Templates
- https://github.com/tr7zw/ProcessedModTemplate
- There's also the rabbithole of [Stonecutter](https://stonecutter.kikugie.dev/) and related projects

PR #1002, opened 2025-10-31
[6.0] Update buildscript to use external Gradle API

Jonathing:


Jonathing:
Build failed because I forgot to update the GitHub workflow to use Java 17. I'll deal with it later. The project builds fine.

PR #1003, opened 2025-10-31
[6.0] Add independently-scoped dependencies to userdev config

Jonathing:
Skipping context for now.

This PR adds magic to ForgeGradle 6.0 which makes it automatically add [mixinextras-common](https://central.sonatype.com/artifact/io.github.llamalad7/mixinextras-common) to the compileOnly and annotationProcessor configurations if it is not already included in the resultant classpaths *and* the source set's jar task contains an attribute of either "MixinConfigs" or "MixinConnector".

Current concern right now is how this will affect older Forge version development environments. Open to suggestions.


Jonathing:
**This is ready for final review.** It has now be come an effective requirement for the MixinExtras JiJ option for Forge.

I decided to move forward with the solution of declaring the MixinExtras artifact in the userdev config. This can be generalized in a future version of patcher/userdev where dependencies are declared as compile, runtime, annotation processor, or any of the three (as a reminder, `implementation` goes to both compile and runtime classpaths).

Jonathing:
I've decided that instead of checking the resultant jar's manifest attributes to see if it is using Mixin, MixinExtras will just attempt to be added to the compile classpath regardless. Again, I must re-iterate, that it is **not** being added to the runtime classpath. Forge, in dev, will not load MixinExtras is if it is not requested via Paint's FML magic selector (that is, if the developer does not use the `mixin.config` property or has a local `MANIFEST.MF` file to be loaded with that attribute).

Jonathing:
Since I was asked this in Discord, I'll mention it here. This PR is the effective alternative to Bootstrap filtering out MixinExtras-Common from the boot layer. Instead of hard-coding in a module blacklist, this instead expands on the UserDev config to allow for additional compile-only and annotation processor dependencies to be declared and used.

I still need to write in the logic to generalize the solution, but that's what this will be. This is something that can be expanded upon in patcher v4 or UserDev config v3, but is otherwise a good solution for now. Adding fields to the spec doesn't break any of our tools since foreign fields are simply ignored by GSON (and if they aren't I can account for that).

As for Mavenizer, it can easily be adapted by making a new variant with the `java-api` usage attribute.

Gocti:
Hello, will java 25 support be added in FG 6 or will it be done in FG 7?

Jonathing:
ForgeGradle 7, because it will support Gradle 9. Java 25 support was added in Gradle 9.1.0.

LlamaLad7 (approved):

PaintNinja (approved):

Jonathing:
Updated to generalize the concept of extra scoped dependencies.

Jonathing:
This PR is ready for re-review.

Jonathing:
Annotation processor inheritance can be disabled with `fg.inheritAnnotationProcessor = false` in the buildscript.

PR #1005, opened 2025-11-06
[7.0] Create Eclipse launchers based off of Slime Launcher run configs

Jonathing:
Name entails. This task is largely a copy of `SlimeLauncherExec` but with the sole purpose to create the launch file that Eclipse will use.

Still a work in progress, including managing the classpath.


Jonathing:
I am marking this PR as ready for review. The run configurations generate and work well.

**However,** I must state one big caveat: this run configuration is using the runtime classpath as stated by Gradle. Even though I am defining it all in the run task, that does mean that `bin` is not the classpath for the compiled mod classes, but rather `build/sourceSets/main`. I'm looking into how I can hook the Eclipse compiler to output to that directory (hopefully Buildship does that part for me already).

LexManos:
Is there any particular reason you're specify the classpath at all?
Letting gradle/eclipse manage that seems to be the best option.
The issue arrives with older versions when we need to specify custom module arguments but thats something we can address when the time comes.

Jonathing:
Slime Launcher is not a part of the runtime classpath, it is added to it as part of its run task.

Aside from that, I don't trust what Buildship thinks the output directories are. Don't forget, we also talked about how Buildship has the tendency to combine the compile and runtime classpaths. I just don't think it's a risk worth taking.

The run configuration simply has all the information the run task would have used.

LexManos:
As long as it works, then im fine with it, Just a bit worried about it using the incorrect directory for the project's path.
Does hot swapping work correctly?

Jonathing:
Don't know yet. I need to check what the ECJ's output folder is, which I hope Buildship takes care of. If not, I can probably change it through Gradle's Eclipse plugin.

Once I confirm hotswapping works, I will merge this.

Jonathing:
Alright, I've come up with a workaround to the Eclipse output directory issue for now.

If ForgeGradle is configured to merge source sets (via the property in `gradle.properties`), it will automatically set the default output directory to the unified source sets directory (`build/sourceSets`). When ForgeGradle creates the Eclipse run configurations, it will remove the destination directories for the Gradle source set's java and resources and add the Eclipse default destination directory instead. This isn't a perfect solution but I don't expect anyone to manually change the output directory for Eclipse's compiler.

Do note one limitation of this workaround: if you build *using Gradle* then launch the game, but save a file in Eclipse, the hotswap won't work. So you need to save/build the project in Eclipse *before* running the game. I think this is a fair drawback, but let me know your thoughts.

Jonathing:
Whenever I get time to breathe, I'll be merging this. While not perfect, Eclipse runs work and do not break in basic setups.

Insert Joker ahh monologue about how I'll infiltrate Buildship someday to fix eclipse-buildship/buildship#1130

Jonathing:
Yes. I can lessen the requirement and use some safe defaults, but I am using EclipseModel to get important details about the Eclipse project, such as the compiler output directory.

Jonathing:
@jebibot As of 7.0.0-rc.1, `genEclipseRuns` can now be run without needing the Eclipse plugin to be present. It will emit warnings if it is not, however. If you plan on actively using Eclipse, I would recommend either applying the Eclipse plugin yourself or running `genEclipseRuns` directly from Eclipse. `genEclipseRuns` is also run as a synchronization task as part of project import.

PR #1017, opened 2025-12-13
[7.0] Check minimum Gradle version in middle-man plugin entry point

### LINKED ISSUE #1016: [7.0] Failed to notify project evaluation listener
**Forge Gradle:** 7.0-beta.52 und 7.0-beta.53

I wanted to test the [Mixin MDK Example](https://github.com/MinecraftForge/MDKExamples/tree/master/Mixins/fg7) without any modification and getting an the `Failed to notify project evaluation listener` error, until beta.51 everything works fine.

Full Log: https://mclo.gs/JSEkfTn

Jonathing:
- Closes #1016

This PR adds a Java 8 compatible runtime checker as the plugin entry-point for ForgeGradle 7. All it does is check that the Gradle version is not less than 9.3.0-rc-1. If the version requirement is not met, an exception is immediately thrown.

The setup for this is a little scuffed, but it's good enough for now considering recent events with people needing to use ForgeGradle 7 since other plugins have bumped their minimum Gradle requirements.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,6 +26,10 @@ java {
     withJavadocJar()
 }
 
+shadow {
+    addTargetJvmVersionAttribute = false
+}
+
 gradleutils.pluginDevDefaults(configurations, libs.versions.gradle)
 
 configurations {
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,25 @@ java {
 
 gradleutils.pluginDevDefaults(configurations, libs.versions.gradle)
 
+configurations {
+    dependencyScope('runtimeCheckerOnly') {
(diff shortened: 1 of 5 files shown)


Jonathing:
I've disabled declaring the target JVM in Gradle module metadata. This is because the main project uses Java 17, but this Java 8 compatible entry point would not allow the plugin to be loaded. Since Gradle 9 requires Java 17, it is impossible to use ForgeGradle 7 using Java 8.

Jonathing:
This PR is fully finished and functional. I'll leave it to Lex to do the squash and merge once his internet is back online (don't forget to remove the [7.0] prefix).

Jonathing:
:shipit:

PR #1039, opened 2026-02-03
[7.0] Fix genEclipseRuns not using correct eclipse output directories

### LINKED ISSUE #1038: [7.0] Don't see code in debug with Eclipse and not getting code changes
This is a conversion of my fg6 project from report #1037 to fg7, by using the mixin-only fg7 example as reference (though I've used that in 1.21.11 with same results).

If I set a breakpoint and run then it stops but shows no source code.
If I also change the code to add a logger message and run then I don't see that message. Only after doing gradlew geneclipseruns,
do you see the changes when running.
Both of these problems were not seen in FG6.

Here is my repo https://github.com/Lupicus/TestMixin/tree/1.21.10-fg7

LexManos:
This is to address https://github.com/MinecraftForge/ForgeGradle/issues/1038
The eclipse runs were not using the correct directories.

Basically, because I didn't want to figure out how to pass in the eclipseModel to a task. I do all the work at configuration time.
But also this is called during afterEvalute because thats when MinecraftDependency.handle is called.
That needs to be fixed, as only validation should be happening in afterEvaluate, but thats a bigger thing for later.

I deleted getEclipseOutputDir because it is part of the internal api, and thus not what I consider a breaking change.
In its place, i grab the EclipseModel as needed.
From the EclipseModel I use:
 - sourceSets
 - baseSourceOutputDir
 - defaultOutputDir
 
From SourceSet I use:
  - name
  - output.resourcesDir
  - output.classesDirs

This works for now, if this looks sane enough to you @Jonathing pull.

Copy pasting some relevant details form discord:
Okay so The EclipseClasspath task uses the EclipseModel. Which includes [SourceFolder](https://github.com/gradle/gradle/blob/master/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/eclipse/model/SourceFolder.java) entries.
These entries have a 'output' value which is set via [SourceFoldersCreator](https://github.com/gradle/gradle/blob/master/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/eclipse/model/internal/SourceFoldersCreator.java#L123)
Who [collectSourceSetoutputPaths](https://github.com/gradle/gradle/blob/master/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/eclipse/model/internal/SourceFoldersCreator.java#L220) function basically just does: 
```java
path = base + '/' + sourceSet.name
while (claimed.contains(path)) path += '_'
return path
```
Easy enough to mimic. Except for the base part. That is gotten from [EclipseClasspath.getBaseSourceOutputDir](https://github.com/gradle/gradle/blob/master/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/eclipse/model/EclipseClasspath.java#L245), relativzed to the [project directory](https://github.com/gradle/gradle/blob/540b81b4fb2d312a8009fe3b7e6c5c014236b724/platforms/ide/ide/src/main/java/org/gradle/plugins/ide/eclipse/model/internal/SourceFoldersCreator.java#L60). 
Which for completeness sake, defaults to [project.dir("bin")](https://github.com/gradle/gradle/blob/master/platforms/ide/ide-plugins/src/main/java/org/gradle/plugins/ide/eclipse/EclipsePlugin.java#L219)

This is all simple enough if we can figure out how to pass in the EclispeModel and sourceSets into the task.

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/internal/MinecraftDependencyImpl.java
+++ b/src/main/java/net/minecraftforge/gradle/internal/MinecraftDependencyImpl.java
@@ -198,7 +198,7 @@ abstract class MinecraftDependencyImpl implements MinecraftDependencyInternal {
                 .matching(MinecraftDependencyInternal::is)
                 .size() == 1;
             runs.forEach(options -> {
-                var task = SlimeLauncherExec.register(getProject(), sourceSet, (SlimeLauncherOptionsImpl) options, module.get(), version.get(), asPath.get(), asString, single, minecraft.getEclipseOutputDir());
+                var task = SlimeLauncherExec.register(getProject(), sourceSet, (SlimeLauncherOptionsImpl) options, module.get(), version.get(), asPath.get(), asString, single);
             });
         });
 
--- a/src/main/java/net/minecraftforge/gradle/internal/MinecraftExtensionImpl.java
+++ b/src/main/java/net/minecraftforge/gradle/internal/MinecraftExtensionImpl.java
@@ -329,7 +329,11 @@ abstract class MinecraftExtensionImpl implements MinecraftExtensionInternal {
 
             project.getPluginManager().withPlugin("eclipse", eclipsePlugin -> {
                 if (mergeSourceSets)
-                    project.getExtensions().configure(EclipseModel.class, eclipse -> eclipse.getClasspath().setDefaultOutputDir(sourceSetsDir.getAsFile().get()));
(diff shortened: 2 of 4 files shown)


Jonathing:
Works. See 4bbe2b4 for the small change I made which sets both `defaultOutput` and `baseSourceOutput` to `build/sourceSets` if merged source sets is enabled. I am still not entirely sure how they work at anything more than a surface level, but it seems to have fixed the underlying issue, and I'm able to debug both my own code and external libraries.

Jonathing (approved):

PR #1046, opened 2026-02-17
[7.0] Change local Mavenizer repo output directory

### LINKED ISSUE #1043: [7.0] Consider moving Mavenizer output from `build` to `.gradle`
Name entails. I think it would be a good idea to move the Mavenizer output to the project directory's `.gradle` folder instead of the `build` folder. The main reason for this is that it is still a common practice to run `gradle clean build` to clean the build artifacts first then immediately build them again within the same Gradle invocation.

Jonathing:
This PR changes  outputs in `.gradle` instead of `build`, to allow for conventional task invocations such as `gradlew clean build`

- Closes #1043.


LexManos (approved):

Jonathing:
Tested locally with no caches and 7.0.5 caches.

PR #1051, opened 2026-02-27
[7.0] Add support for run tokens

### LINKED ISSUE #1048: [7.0] Minecraft 1.20.2 Forge 48.1 fails to runClient
I'm able to reproduce this using a clean, minimal FG7 MDK using 1.20.2 Forge 48.1.

Here's the MDK I used to test this: [fg7.0.7-mc1.20.2-mdk-test.zip](https://github.com/user-attachments/files/25461078/fg7.0.7-mc1.20.2-mdk-test.zip)  
And here's the runClient log: https://gist.github.com/Jonathing/1774fa157daed7dac1b9f6d7d1c378b8

LexManos:
Initial work on fixing #1048 (and older versions)
Mavenizer will need an update to support outputting mcp_version, mc_version, and modules
Slime Launcher will need to be updated as well, maybe make it responsible for mcp_version/mc_version from metadata.

Not sure what to do with extracting the natives, I think Slime should be responsible for that as well.

Did a little reorganizing, moved the register methods for Metadata/Eclipse to their own class.
Made MinecraftDependencyInternal expose a lot more data.
Made MinecraftDependecyInternal hold references to things that will be used by multiple runes, the metadata task, dependency configurations, etc..


LexManos:
Should be good to go, Bumped Mavenizer to expose the module list.
I have verified it works with 1.20.2 which uses {module} and {runtime_classpath_file}
The only pending things left that I have not implemented are {mcp_to_srg} and {source_roots}
I don't know what versions use {mcp_to_srg}, nothing shows up in search, and haven't written a scraper to figure that out.


It should be good to go, just want someone besides me to verify as its a fairly big change.

I've also included my Eclipse Run config changes from ForgeDev.
As well as cleaning up the MavenizerSync check flow stuff, as its not needed anymore with the new ValueSource approach.

Swackyy:
Took a bit of fiddling with gradle shenanigans but it works on my (multiloader) end well, aside from it not finding my mod during mod discovery it seems. Probably an issue on my part although I am mentioning it due to the possibility of it not being me
image

LexManos:
Nice verification that it works on someone elses machine is always good. As for not being able to find your mod. That might be the source_roots token.
https://github.com/MinecraftForge/ForgeGradle/blob/2fec3fd53defe2c5716f64dd31e3d73bdfd55007/src/common/java/net/minecraftforge/gradle/common/util/runs/RunConfigGenerator.java#L144
Is that meing used anywhere in your launch args?

Swackyy:
Doesn't seem to be, searched my console and scanned the `ModLauncher running: args []` array printed but nothing

LexManos:
Interesting. If you wanna donate a zip of your setup (or github link) i can take a look when im back home.

Swackyy:
[Sure here](https://github.com/Swackyy/Ohmega/tree/1.20.2), this will grab the plugin from maven though, so if you need to change the version or something to test, the forge `build.gradle` is located here: `loader/forge/build.gradle`

Swackyy:
[Here is a minimal 1.20.2 mdk](https://github.com/user-attachments/files/25667033/test.zip) running a mavenlocal published version of forgegradle (up to date with the commits attached in this thread), replicating the source_root mod not being found issue

LexManos:
Alright, so the issue is that the MOD_CLASSES environment variable needs to be filled in using the `{source_roots}` token. I've added support for specifying it using the FG6 style `mods` block in the run configs.

However I have NOT added the magic 'just add this to everything' that FG6 did. So for versions where this matters you may have to specify the environment variable yourself. I could probably add the special case but it feels dirty to do.


So for your first example:
```groovy
runs {
    configureEach {
        environment 'MOD_CLASSES', '{source_roots}'
        workingDir.convention layout.projectDirectory.dir("run")
        systemProperty "eventbus.api.strictRuntimeChecks", "true"
        systemProperty "forge.enabledGameTestNamespaces", modId
        args "-mixin.config", "${modId}.mixins.json"
        args "-mixin.config", "${modId}_common.mixins.json"
    }

    register("client") {
        with(sourceSets.test) {
            mods {
                "boobies" {
                    source sourceSets.boobies
                }
            }
        }
    }

    register("server")
}
```
Results in:
client: `MOD_CLASSES: DIR\build\sourceSets\main;DIR\build\sourceSets\main`
testClient: `MOD_CLASSES: boobies%%DIR\build\sourceSets\boobies;boobies%%DIR\build\sourceSets\boobies`
server: `MOD_CLASSES: DIR\build\sourceSets\main;DIR\build\sourceSets\main`
testServer: `MOD_CLASSES: DIR\build\sourceSets\main;DIR\build\sourceSets\test`
Note: By default a run config will have the main sourceset, and the test sourceset in the test variant.
Adding any mod entry will overwrite the default not add to it.

Your mod still doesn't work when launching in eclipse because your mods.toml has replacements in it. But thats an issue on your end (and gradle/eclipse being stupid) not ours. For this I would recommend that you move anything that is modified at build time to a template folder. And add processResources or something to your IDE sync tasks.

As it sits I need @Jonathing to give this a once over, and fix up some modder facing stuff/make gradle happy. Namely [this](https://github.com/MinecraftForge/ForgeGradle/pull/1051/changes#diff-e10cab4d9347e9004420d4efa4f36cd286058f0acd9732704dcfc419a99a8f68R218)

Once he is happy he'll pull it.

Swackyy:
I use IntelliJ and it works fine, really is just being awkward. For completeness I'll fix it, cheers for telling me that

src/main/java/net/minecraftforge/gradle/internal/MinecraftDependencyImpl.java
```diff
@@ -96,6 +94,49 @@ public void mappings(String channel, String version) {
...
+    public FileCollection getPatcherModules() {
+        if (this.patcherModulesConfig == null) {
+            var cfg = this.getProject().getConfigurations().detachedConfiguration();
+            cfg.setCanBeResolved(true);
+            cfg.setTransitive(false);
+            cfg.getDependencies().addAllLater(
+                this.getMavenizerInstance().getPatcherModules().map(lst -> {
+                    var ret = new ArrayList<Dependency>(lst.size());
+                    for (var gav : lst)
+                        ret.add(getProject().getDependencyFactory().create(gav));
```
Jonathing, line 124: If there are no patcher modules, the list is a single entry of empty String (due to `List.of(value.split(","))`), which nukes the build as an empty string is not a valid module. I'll see if I can take care of this real quick.

Jonathing:
Finally merging this, thanks for your patience. I had Lex do a sanity check with his tests, but I wouldn't be surprised if there may be a few extra things that creep up down the line. For sake of brevity, let's call ForgeGradle 7.1 the "stable release" for Minecraft 1.12.2 - 1.20.4 (UserDev 3 with obfuscation).

PR #1071, opened 2026-04-06
[7.0] Display a warning if Renamer is not present using legacy Forge versions

### LINKED ISSUE #1047: [7.0] Show a warning when using 1.20.4- without the renamer plugin
When using FG7 on 1.20.4 and older without the renamer plugin, everything appears to work at first - it runs, stuff compiles, a jar gets built successfully - but the jar won't work in production.

It would be good if FG7 could show some kind of warning or problems report when the Minecraft.dependency or mappings version is one that needs remapping and the renamer plugin is missing, to avoid silently broken builds.

Jonathing:
Name entails. Specifically, display warning if snapshot version (detected by looking for `w`) is less than 24w13(a) or main version is less than 1.20.5 (ghost .0 is added to versions such as 1.20).

- Closes #1047

### DIFF
--- a/src/main/java/net/minecraftforge/gradle/internal/ForgeGradleProblems.java
+++ b/src/main/java/net/minecraftforge/gradle/internal/ForgeGradleProblems.java
@@ -97,25 +97,40 @@ abstract class ForgeGradleProblems extends EnhancedProblems {
     }
 
     void reportMissingRenamerPluginForOldVersion(Dependency dependency) {
-        if (this.testFalse("net.minecraftforge.gradle.warnings.minecraft.legacy.renamer.missing")) return;
+        // This should never happen, so assert this check for testing purposes.
+        assert !this.testFalse("net.minecraftforge.gradle.warnings.minecraft.legacy.renamer.missing");
 
         LOGGER.warn("WARNING: Renamer Gradle not present with legacy Forge version. See Problems report for details.");
         this.report("legacy-missing-renamer", "Missing Renamer Gradle for legacy Minecraft dependency", spec -> spec
             .details("""
                 A legacy Forge dependency was declared, but Renamer Gradle has not been applied to the project!
-                While the workspace will continue to function, this may cause problems with resultant artifacts not being renamed to obfuscated mappings.
+                While your mod will work in development, the built jar won't work in production without Renamer Gradle's reobf functionality.
                 Legacy Forge versions use obfuscated mappings at runtime, so this is a requirement if you are publishing this project as a mod that uses Minecraft names.
                 Dependency: '%s'"""
                 .formatted(Util.toString(dependency)))
             .severity(Severity.WARNING)
             .solution("Apply the 'net.minecraftforge.renamer' plugin.")
+            .solution("Review MDKExamples to cross-reference your setup with a working example using Renamer Gradle.")
             .solution("Disable this warning in 'gradle.properties' if you are an advanced user: `net.minecraftforge.gradle.warnings.minecraft.legacy.renamer.missing=false`")
             .solution("Consider building your project for a newer version of Forge targeting Minecraft 1.20.5 or newer.")
             .solution(HELP_MESSAGE));
     }
 
     void reportMissingRenamerCheckFailed(Dependency dependency, Throwable e) {
-        LOGG
+        assert !this.testFalse("net.minecraftforge.gradle.warnings.minecraft.legacy.renamer.missing");
+
+        LOGGER.warn("WARNING: Failed to check if Renamer Gradle is required. See Problems report for details.");
+        this.report("failed-missing-renamer-check", "Failed to check if Renamer is required", spec -> spec
+            .details("""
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/gradle/internal/ForgeGradleProblems.java
```diff
@@ -96,6 +96,28 @@ void reportMissingMinecraftDependency() {
...
+                .formatted(Util.toString(dependency)))
+            .severity(Severity.WARNING)
+            .solution("Apply the 'net.minecraftforge.renamer' plugin.")
+            .solution("Disable this warning in 'gradle.properties' if you are an advanced user: `net.minecraftforge.gradle.warnings.minecraft.legacy.renamer.missing=false`")
+            .solution("Consider building your project for a newer version of Forge targeting Minecraft 1.20.5 or newer.")
+            .solution(HELP_MESSAGE));
+    }
+
+    void reportMissingRenamerCheckFailed(Dependency dependency, Throwable e) {
+        LOGG
```
PaintNinja: Possible typo?

src/main/java/net/minecraftforge/gradle/internal/ForgeGradleProblems.java
```diff
@@ -96,6 +96,28 @@ void reportMissingMinecraftDependency() {
...
+        LOGGER.warn("WARNING: Renamer Gradle not present with legacy Forge version. See Problems report for details.");
+        this.report("legacy-missing-renamer", "Missing Renamer Gradle for legacy Minecraft dependency", spec -> spec
+            .details("""
+                A legacy Forge dependency was declared, but Renamer Gradle has not been applied to the project!
+                While the workspace will continue to function, this may cause problems with resultant artifacts not being renamed to obfuscated mappings.
+                Legacy Forge versions use obfuscated mappings at runtime, so this is a requirement if you are publishing this project as a mod that uses Minecraft names.
+                Dependency: '%s'"""
+                .formatted(Util.toString(dependency)))
+            .severity(Severity.WARNING)
+            .solution("Apply the 'net.minecraftforge.renamer' plugin.")
```
PaintNinja, line 112: Mention looking at the MDK Examples repo as applying the plugin alone doesn't fix built jars iirc

src/main/java/net/minecraftforge/gradle/internal/ForgeGradleProblems.java
```diff
@@ -96,6 +96,28 @@ void reportMissingMinecraftDependency() {
...
     }
 
+    void reportMissingRenamerPluginForOldVersion(Dependency dependency) {
+        if (this.testFalse("net.minecraftforge.gradle.warnings.minecraft.legacy.renamer.missing")) return;
+
+        LOGGER.warn("WARNING: Renamer Gradle not present with legacy Forge version. See Problems report for details.");
+        this.report("legacy-missing-renamer", "Missing Renamer Gradle for legacy Minecraft dependency", spec -> spec
+            .details("""
+                A legacy Forge dependency was declared, but Renamer Gradle has not been applied to the project!
+                While the workspace will continue to function, this may cause problems with resultant artifacts not being renamed to obfuscated mappings.
```
PaintNinja: Something something "while your mod will work in dev, the built jar won't work in production without the renamer plugin applied and reobf used on it"

PaintNinja (changes requested):

src/main/java/net/minecraftforge/gradle/internal/ForgeGradleProblems.java
Jonathing: no way, i forgot to finish that method. awesome.

LexManos:
Any chance either of you could finish this up?

LexManos (approved):

src/main/java/net/minecraftforge/gradle/internal/ForgeGradleProblems.java
```diff
@@ -96,6 +96,43 @@ void reportMissingMinecraftDependency() {
...
 
+    void reportMissingRenamerPluginForOldVersion(Dependency dependency) {
+        // This should never happen, so assert this check for testing purposes.
+        assert !this.testFalse("net.minecraftforge.gradle.warnings.minecraft.legacy.renamer.missing");
+
+        LOGGER.warn("WARNING: Renamer Gradle not present with legacy Forge version. See Problems report for details.");
+        this.report("legacy-missing-renamer", "Missing Renamer Gradle for legacy Minecraft dependency", spec -> spec
+            .details("""
+                A legacy Forge dependency was declared, but Renamer Gradle has not been applied to the project!
+                While your mod will work in development, the built jar won't work in production without Renamer Gradle's reobf functionality.
```
PaintNinja:
```suggestion
                While your mod will work in development, the built jar won't work in production without Renamer Gradle's reobf functionality explicitly setup.
```

PaintNinja (approved):

Issue #1080, opened 2026-07-24
[7.0] Add support for transitive injected interfaces with Facade


PaintNinja:
AccessTransformers deliberately don't apply transitively because there's no easy way of knowing if MC had something public or if one of your dependencies changed it, which can result in unintended reliance on a mod's build time optional dependencies.

Interface injection on the other hand doesn't change existing code - it adds new methods to an existing class, and you can see where the methods are coming from because they're on the interface, not the class the interface is applied to. The risk of accidentally relying on another mod is reduced as they can see in their IDE that a new method they're referencing is coming from a non-MC package.

Some mods like GeckoLib heavily rely on interface injection to extend vanilla MC's APIs, as alternatives like utility methods and wrapping aren't as nice for API users to work with, are more effort to maintain and less performant. Currently mod devs need to setup Facade in their own project and manually check that their Facade config matches each of their deps in order to use them, it would be nice if injected interfaces applied transitively so that they'd work out of the box. It could also be handy for Forge, as some patches that inject interfaces could be replaced with a Facade config instead.

LexManos:
At the time that facades are injected into mavenizer from FG there is no information about transitive dependencies.

I am also cynical/skeptical that the surfacing of where a method comes from is sufficient for modders to not fall into the accidental dependency trap. Im not a fan of any automated magic. I would have to see concrete examples of what you see as the discoverability. 

However looking at the existing code, the whole chain should be able to support multiple facade config files as is. The problem becomes discovering/feeding them into fg.

Currently as they are purely a compile time tool there is no standardized location for the facade configs. (And no expectation that they are shipped in dev jars)

Perhaps we can add some form of generic `fg.depFile('foo:bar:1
0', '/facade.cfg')`

So that modders can opt-in to depending on the ijectected interfaces of dependencies by doing something like:

```groovy
dependencies {
  implementation fg.dependency(forge) {
  facade = files(
    '/src/main/resources/facade.cfg',
     fg.depFile('foo:bar:1.0', '/facade.cfg')
    )
  }
}
```

Im pretty sure you can do that already using zipTree but a helper wouldn't be such a bad idea.

If we standardized the location of the file in the jar (or added a manifest entry to locate it) it could be even smaller. `fg.depFacade('foo:bar:1.0')` or `fg.depFacades(configurations.hard_mod_deps)`

[...]

PR #1081, opened 2026-08-07
Add helper methods to access files inside zip files

LexManos:
This is a first draft of what I had in mind for addressing #1080
This just adds helper methods for converting a FileCollection of archives, to a FileCollection of files from those archives.

Example usages (untested, but should work):

```groovy
configurations {
  mods
  implementation.extendsFrom mods
}

dependencies {
  mods renamer.dependency(libs.modA)
  mods renamer.dependency(libs.modB)

  implementation minecraft.dependency(forge) {
    facades.setFrom(files(
        'src/main/resources/META-INF/facades.cfg',
        fg.findFacades(configurations.mods)
    ))
    accessTransformers.setFrom(files(
        'src/main/resources/META-INF/at.cfg',
        fg.findAccessTransformers(configurations.mods)
    ))
  }
}
```

Its fairly generic, so you can do something like this:
```groovy
configurations {
    sharedLibs
}

dependencies {
    sharedLibs 'org.json:json:20260719'
}
fg.findFiles(configurations.sharedLibs, 'META-INF/MANIFEST.MF').forEach { 
  file -> logger.lifecycle('File: ' + file.absolutePath) 
}
```
Outputs a nice:
`File: Z:\test\_test\build\minecraftforge\forgegradle\zip_data\82390308e0b67205bf0d0d507dc48e0d71410bf7\json-20260719\META-INF\MANIFEST.MF`

In addition this defacto standardizes the Facade files in deps to:
A space separated list in the `FORGE_FACADE` manifest attribute OR `META-INF/facades.cfg`

Javadocs need to be written.
Do you think I should promote the `BiFunction<File, JarFile, List<ZipEntry>>` to a proper `EntryFinder` interface and expose a `mapZip(FileCollection, EntryFinder)`?

Adding support for the access transformer helper requires parsing the mods.toml, as such I have to add a TOML dependency, so I add [Jezza Toml](https://github.com/Jezza/toml) which adds ~30kb to the jar. Same library I use for Renamer.

The functions that take Object instead of FileCollection just pass it directly to `project.files` as a small helper to make cleaner build files.

`findFiles(files, name)` just grabs the explicit file from any archive that has it
`findAccessTransformers(files)` tries to find ATs from the manifest, toml, and default location
`findFacace(files)` tries to find the facade config from the manifest, or default location

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,6 +64,7 @@ dependencies {
     implementation libs.maven.artifact
     implementation libs.bundles.utils
     implementation libs.srgutils
+    implementation libs.toml
 }
 
 license {
--- a/settings.gradle
+++ b/settings.gradle
@@ -56,5 +56,8 @@ dependencyResolutionManagement.versionCatalogs.register('libs') {
 
     // Used to compare Minecraft Versions
     library 'srgutils', 'net.minecraftforge', 'srgutils' version '0.6.5'
+
+    // Used to parse mods.toml to locate access transformer locations
+    library 'toml', 'com.github.jezza', 'toml' version '1.2'
 }
 //@formatter:on
--- a/src/main/java/net/minecraftforge/gradle/ForgeGradleExtensionForProject.java
+++ b/src/main/java/net/minecraftforge/gradle/ForgeGradleExtensionForProject.java
@@ -11,10 +11,47 @@ import org.gradle.api.file.FileCollection;
 ///
 /// @see ForgeGradleExtension
 public interface ForgeGradleExtensionForProject extends ForgeGradleExtension {
-    FileCollection mapZip(Object files, String name);
-    FileCollection mapZip(FileCollection files, String name);
-    FileCollection mapAccessTransformer(Object files);
-    FileCollection mapAccessTransformer(FileCollection files);
-    FileCollection mapFacades(Object files);
-    FileCollection mapFacades(FileCollection files);
+    /// Attempts to locate and extract a file inside archives.
+    /// @param files The files to search, accepts anything that can be passed to {@link org.gradle.api.Project#files(Object...)}
+    /// @param name The exact name to search for in archives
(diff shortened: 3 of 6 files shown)


XueGao5155:
Very good！

XueGao5155:
configurations {
    mods {
        canBeResolved = true
        canBeConsumed = false
        transitive = true
    }
    implementation.extendsFrom mods
}

dependencies {
    mods libs.modA, libs.modB
}

def collectModConfigs(String type) {
    def files = files("src/main/resources/META-INF/${type}.cfg")
    configurations.mods.resolvedConfiguration.resolvedArtifacts
        .findAll { it.type in ['jar', 'module'] }
        .collectMany { zipTree(it.file).matching { include "META-INF/${type}.cfg" }.files }
        .each { files += it }
    return files
}

minecraft.dependency(forge) {
    facades.setFrom(provider { collectModConfigs('facades') })
    accessTransformers.setFrom(provider { collectModConfigs('at') })
}

I worked on the changes together with a friend who knows coding; if something isn’t done well, please don’t scold me

XueGao5155:
I’m not very good at using GitHub; it’s my first time using it

Jonathing:
I would prefer something like `findFiles` be implemented primarily in gradleutils-shared, since it's so generic. We can create a bouncer to it in FG, though.

LexManos:
@XueGao5155 You dont need to write your own collectModConfigs that what this PR added. See the examples in the main post.

Ya, I thought about gradle-utils. As it is fairly generic/useful. But I shoved it in here for the same reason I didn't put the Facade/AT finders in their respective gradle plugins: lazieness.

Feel free to snag it for gradle-utils, its basically just a helper snippit anyways.

XueGao5155:
@LexManos Okay, I understand now; this post was posted by one of my ancestors who came out of his grave and accidentally sent it without intending to. I’ll delete it right away！

